PR #69, opened 2019-12-02
fix error of type hints

RenShuhuai-Andy:


nreimers:
Thank you so much for your PR :)

PR #76, opened 2019-12-11
fix comment for one of the examples

vitekzach:
Hey, I found a small mistake in the comment for one of the examples.


nreimers:
Thanks for the fix :)

PR #88, opened 2020-01-05
Replaced AgglomerativeClustering with KMeans clustering

goodrahstar:


nreimers:
Hi,
thank you for your contribution.

Could you explain why you would like to change AgglomerativeClustering with KMeans clustering? 

Kmeans is usually faster than agglomorative clustering, but on word / sentence embeddings, I usually observe better results with agglomorative (i.e., better clusters) than with KMeans. As it is only a toy example, speed is I think not that critical.

Best
Nils Reimers

laugustyniak:
I have the same experiences as @nreimers with many clusterings of documents [paragraphs, sentences, calls in the call centre with different granularity sucha as turns, sentences, windows]

@goodrahstar if your concerns are related to the performance, I would rather use PCA/UMAP and then Agglomerative rather than changing to kmeans.

nreimers:
Updated this in v0.2.5 => The example mentioned to use kmeans, but used allgomerative clustering instead.

PR #113, opened 2020-01-31
add device selector to Dense

RadionBik:
I ran into an issue when tried to load the distiluse-base-multilingual-cased model using CPU only. This fix adds the device selector. This PR is safe to merge. Looking forward to use the updated package.

### DIFF
--- a/sentence_transformers/models/Dense.py
+++ b/sentence_transformers/models/Dense.py
@@ -41,5 +41,6 @@ class Dense(nn.Module):
 
         config['activation_function'] = import_from_string(config['activation_function'])()
         model = Dense(**config)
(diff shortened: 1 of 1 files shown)


nreimers:
Thanks for pointing this out.

I fixed it with commit 
https://github.com/UKPLab/sentence-transformers/commit/a96ccd3f2a6c031a33a929bbc5f110831f3a9f15

When the model is loaded, it is now always mapped to the CPU. This is the identical behavior as for the BERT models from HuggingFace.

If the model runs on the GPU, it is then moved to the GPU by SentenceTransformer class.

By having the fixed mapping to the CPU, it is possible to run it on a CPU even if cuda is available. The device parameter in SentenceTransformer  defines on which device it should run.

PR #214, opened 2020-05-03
Integration of fast Rust-based tokenizers for WordPiece models

Mirofil:
Hello, I have rewritten the tokenization for WordPiece-based models to make them use the Fast versions of Huggingface tokenizers (the Fast versions are implemented in Rust rather than pure Python) - this yields ~10x improvement in speed for longer texts.

The Fast tokenizers also allow for multi-threaded tokenization on batches - this required changes in the SentencesDataset loop to make it batched. I observed that using 15 CPU cores, it runs around 60x faster in my use-case than the current master branch.

There are some interesting considerations:

- I was able to replicate the results you have in Github Readme. In fact, the results are now even better. Running `/examples/evaluation/evaluation_stsbenchmark.py` for Bert/Distilbert replicates the table in "Trained on NLI Data" section, but Roberta has around 0.5 higher score. Likewise, I tried running `/examples/training_transformers/training_stsbenchmark.py` with both bert-base-uncased and albert-base-v1 (which does not even have a Fast tokenizer available) and in both cases, my branch has 0.5-1 point higher score on STS eval than the current master branch.

- I also ran the Clustering/Semantic search examples - the semantic search results seem significantly improved in the sense that the difference between relevant/non-relevant sentences is much bigger

- I tried running the tests but it did not work for me even on master

- _Technical details_ - The Fast and Slow versions have non-uniform interfaces. In order to support both, I had to make changes to the tokenization methods, but it seems to work well as it is now. For example, `BertTokenizerFast.tokenize` already inserts special tokens while the slow versions do not. In my branch, the `SentenceTransformer.tokenize` outputs the special tokens as well (which is not the case for master) but they are not added in `SentenceTransformer.get_sentence_features` (again opposite of master). It thus cancels out. 

I thought it would be possible to add `add_special_tokens=False `to the Fast tokenizer initialization and to calls to `.encode` which would maintain the current behavior, but then DistilbertTokenizerFast was failing for some reason (BertTokenizerFast was fine however; there is some internal difference between them in how they handle the parameters I think)

It seems to me that the tokenizers overall have some slight inconsistencies/bugs, and the current tokenization loop works better now for unknown reasons.

In any case, I hope you will find this useful.

Best regards

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -176,7 +176,10 @@ class SentenceTransformer(nn.Sequential):
         return self._first_module().tokenize(text)
 
     def batch_tokenize(self, batch):
-        return self._first_module().tokenizer.batch_encode_plus(batch, return_attention_mask=False)["input_ids"]
+        if hasattr(self._first_module().tokenizer, "batch_encode_plus"):
+            return self._first_module().tokenizer.batch_encode_plus(batch, return_attention_mask=False)["input_ids"]
+        else:
+            return [self._first_module().tokenizer.tokenize(text) for text in batch]
 
     def get_sentence_features(self, *features):
         return self._first_module().get_sentence_features(*features)
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -175,6 +175,9 @@ class SentenceTransformer(nn.Sequential):
     def tokenize(self, text):
         return self._first_module().tokenize(text)
 
+    def batch_tokenize(self, batch):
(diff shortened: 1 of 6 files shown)


nreimers:
Thanks for this great PR.

I will check it as soon as possible and then merge it into the repository.

Quite strange, that you get performance improvements. I would have expected that the tokenization remains the same independent if you use the slow Python tokenizer or the faster Rust tokenizer from HuggingFace.

nreimers:
With the new tokenizer API 3.x I added an option that allows to change the tokenizer. It can be used like this:
```
model = sentence_transformers.SentenceTransformer('distilbert-base-nli-mean-tokens')
model.tokenizer = DistilBertTokenizerFast.from_pretrained('distilbert-base-uncased')
```

Then, the fast tokenizer based on Rust will be used.

PR #232, opened 2020-05-16
fix unnecessary "is" to "==" for avoiding SyntaxWarning in python 3.8

uidilr:
the current code results following SyntaxWarning.
I replaced "is" with "==" to avoid this warning.

~/.pyenv/versions/3.8.1/lib/python3.8/site-packages/sentence_transformers-0.2.5-py3.8.egg/sentence_transformers/SentenceTransformer.py:52: SyntaxWarning: "is" with a literal. Did you mean "=="?


nreimers:
Thanks for the fix

PR #250, opened 2020-05-28
fixed file name in training_multilingual example

SidJain1412:
File name was generalized so error wasn't correct


nreimers:
Thanks for the PR :)

PR #277, opened 2020-06-17
Added method sentence_transformers.SentenceTransformer.embed(List[int]) that takes pre-encoded text from ex. transformers.BertTokenizer

rjurney:
This PR is not ready! I have to factor out redundant code. I want you to see it to find out if you'll accept it and what the `embed(List[int])` method name should be :)

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -220,8 +220,8 @@ class SentenceTransformer(nn.Sequential):
                     features[feature_name].append(sentence_features[feature_name])
 
             for feature_name in features:
-                #features[feature_name] = torch.tensor(np.asarray(features[feature_name])).to(self.device)
-                features[feature_name] = torch.cat(features[feature_name]).to(self.device)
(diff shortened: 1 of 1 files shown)


nreimers:
The PR sounds interesting.

I am currently on vacation until end of June. I will then have a look and integrate it.

rjurney:
@nreimers thanks, I'll factor out redundant code then.

nreimers:
Thanks for the PR.

I included the function in the encode function:
```
tokenized_sentences = [model.tokenize("my first sentence"), model.tokenize("my second sentence")]
model.encode(tokenized_sentences, is_pretokenized=True)
```

rjurney:
@nreimers I’m assuming you factored out the duplicated code? It came down to a 3-4 line change the rest could be removed.

nreimers:
Hi @rjurney 
Yes, I slightly changed the encode function so that it first tokenizes all text before it generates the batches. With this change, you just need a flag if the input is already tokenized or not.

PR #284, opened 2020-06-23
Add autonotebook to fix training in colab

manycoding:
tqdm's default console progress produces millions of lines, which slows down a Colab page specifically.

So I added auto check.


nreimers:
Thanks for the PR. Did know that this autonotebook exists in tqdm

PR #319, opened 2020-07-22
SentenceDataset.convert_input_examples docstring fix

lowzhao:
SentenceDataset.convert_input_examples doesn't return a SmartBatchingDataset. Remove return statement from `convert_input_examples`'s Docstring.


nreimers:
Thanks for  the PR

PR #331, opened 2020-07-29
fix test_roberta_wkpooling by replacing model.Auto with model.Transformer

vkkb:
Hi, the unit test `test_roberta_wkpooling` was resulting in `AttributeError: module 'sentence_transformers.models' has no attribute 'Auto'`. 
I replaced `model.Auto`, which was apparently renamed, with `model.Transformer`. This unit test should run without errors now :)


nreimers:
Thanks for the PR :)

PR #343, opened 2020-08-06
Clarifying where the 'models' class is coming from

azdaly:
It took me embarrassingly long to figure it out. I figured this may help others!


nreimers:
thanks for the PR

PR #396, opened 2020-08-31
solved error during downloading dataset 

Adarsh1999:
Added wget module into project and in the examples like semantic search with faiss and elasticsearch as they are the only one I checked but the error may present in other files too.

Note:- This error wasn't there in before commit 486756803ec60f390f820d5e69898c6240b55269 on Aug 24th I guess so if you don't want to add new dependency you can check the problem before that.

My approach was to remove the problem permanently means in future this problem will never happen my  using `wget` lib which is been working since 2015 so if any changes we do in `https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/util.py` 
here will not effect the following in the future.

To recreate this error for mods just simply delete the .tsv model from the path and it will show the error.

If any better recommendations please reply to the PR...

### DIFF
--- a/examples/applications/semantic_search_quora_elasticsearch.py
+++ b/examples/applications/semantic_search_quora_elasticsearch.py
@@ -21,6 +21,7 @@ import csv
 import time
 import tqdm.autonotebook
 from elasticsearch import helpers
+import wget
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


nreimers:
This is fixed in the latest release. Thanks for pointing this out.

PR #401, opened 2020-09-02
Fix a typo

joker-xii:


nreimers:
Thanks for the PR :)

Best
Nils Reimers

PR #417, opened 2020-09-09
Changed label logic to better identify integers

97harsh:
Added this change to better identify integers, the earlier method cannot did not consider numpy integers (np.int16/32/64, signed, longlong, ...) as valid integers, creating errors in subsequent pipeline
classification problems where expected dtype is integer but is processed as long, where label type was np.int16/32/ ...

### DIFF
--- a/sentence_transformers/datasets/SentencesDataset.py
+++ b/sentence_transformers/datasets/SentencesDataset.py
@@ -1,3 +1,4 @@
+import numpy as np
 from torch.utils.data import Dataset
 from typing import List
 import torch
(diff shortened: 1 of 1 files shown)


97harsh:
#418

nreimers:
Hi,
this PR has the issue,  that no longer Python integers will be working. They will would get the torch.float type.

I will add a fix later to the code base that works for both cases (numpy and regular ints)

97harsh:
Created the PR for this issue(418)

PR #424, opened 2020-09-12
Added a seperator in LabelSentenceReader

zeno17:
LBS only works with .tsv. This just allows anything else with a specific seperator to also be entered


nreimers:
Thanks for the idea. Fixed some bugs and merged it.

PR #438, opened 2020-09-18
Remove .to(self.device). It raises exception 'no attribute named device'

monatis:
This small change fixes the error.


nreimers:
Thanks for the PR

PR #439, opened 2020-09-19
Always store best_score

PhilipMay:
This is just a small fix.

- the best_score during training had been stored only when save_best_model was true
- now best_score is always stored


PhilipMay:
Just wanted to bump this. @nreimers

nreimers:
Sorry for the delay. Thanks for the PR, it is merged :)

PhilipMay:
Yay! ;-)

ReySadeghi:
Hi,
save best model by default is based on cosine_average_precision?

nreimers:
@ReySadeghi 
It is based on what the evaluator, passed to the fit method, returns.

PR #449, opened 2020-09-24
Changed equal to not equal

zeno17:
This look currently continues if positive has the same label.

If it has the same label, it continues looping.
if it doesnt, then it terminates and goes for the negative.

But given that this is for the positive one, arent the labels supposed to be equal?


nreimers:
Here, the guid is compared. The guid is unique for each example. 

That code ensures that the positive is not equal to the anchor.

Issue #448, opened 2020-09-24
Cant weights to change in BatchSemiHardTripletLoss/Training does not alter weights


zeno17:
Hi,

Im trying to alter the embeddings of SBert based on some data. Ideally, i would have 2 sentences and a similarity score to finetune on but i dont and now im trying to use a proxy which is a sentence and a label (0 to 10). I dont actually care about it getting the labels correct, but given that it is just a proxy for some other process that i am trying to simulate (and of which data is hard to get).

This is an example of the data that im trying to read in (shown in a dataframe)
```
  | Sentence | Score
-- | -- | --
Some test sentence | 0.0
Just another sentence to see if it works | 1.0
Well, maybe this one will actually work | 6.0
```
Then i have the following setting
base_model = SentenceTransformer('distiluse-base-multilingual-cased')
model_mli = SentenceTransformer('distiluse-base-multilingual-cased')
qs_reader = LabelSentenceReader(finetunedir, 1, 0)
train_data = SentenceLabelDataset(qs_reader.get_examples('data.tsv'), model_mli)
train_dataloader = DataLoader(train_data, shuffle=True)
train_loss = BatchSemiHardTripletLoss(model_mli, margin=0.01)

Then i try to train it using the following settings (ive used more hyperparameter settings, but ive been trying absolute extreme ones that in practice should wreck any prior weights just to see if it does anything) :

```
model_mli.fit(train_objectives=[(train_dataloader, train_loss)],
                                  epochs=1,
                                  scheduler= 'constantlr',
                                  evaluation_steps=1000,
                                  evaluator = None, 
                                  optimizer_params = {"lr": 1},
                                  output_path=finetunedir + 'test')
```
After training i try to obtain an encoding from both the old model and the one that should be finetuned:
`
enc1 = base_model.encode(['Dit is een testzin die helemaal nergens over gaat. Hoe cool is dat?'])
enc2 = model_mli.encode(['Dit is een testzin die helemaal nergens over gaat. Hoe cool is dat?'])
`

followed by a sanity check 

`np.sum(enc1 == enc2)`

If this sanity check is 512 it means that weights did not move at all.

Unfortunately, trying many hyperparameters and increasing the dataset upto ~6000 sentence-label pairs does not seem to be able to change anything. So im wondering if theres anything wrong with my setup

PR #455, opened 2020-09-28
Fix tripletLoss

### LINKED ISSUE #391: Model didn't converge when fine-tuning with my own dataset
I use the summarization dataset and keywords of the content to fine tune the model for search.
But by checking out the evaluation result, the model seems it didn't change before and after.
I've tried different learning rate and loss function, still the same situation. Even use super high learning rate, the evaluation result still keep same.
[https://colab.research.google.com/gist/WangHexie/42089737a5f17f6c4ee173e92bdebea4/sentencebertforfinetuningdebug.ipynb](https://colab.research.google.com/gist/WangHexie/42089737a5f17f6c4ee173e92bdebea4/sentencebertforfinetuningdebug.ipynb)

(And another question: Other than trying different parameters, do you have any trick, process or tool to debug neural network when construct such a project?)

WangHexie:
fix #391 

[this line of the code](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/losses/BatchHardTripletLoss.py#L93) break the loss. You cannot only take the first one of the return. If you put all instances of one class in a InputExample, this line made the code only use the first example to train. And in the following procedure, the in-class distance will be taken as 0, which make it much smaller than the distance between different classes. Gradient will be 0, and network won't converge.

`        return self.batch_hard_triplet_loss(labels, reps[0])
`
My fix ignore the situation which use label as `the true distance between instances`,  like the implement in #448

I have not tested the other class which  inherit `BatchHardTripletLoss`.

### DIFF
--- a/sentence_transformers/losses/BatchHardTripletLoss.py
+++ b/sentence_transformers/losses/BatchHardTripletLoss.py
@@ -90,7 +90,9 @@ class BatchHardTripletLoss(nn.Module):
 
     def forward(self, sentence_features: Iterable[Dict[str, Tensor]], labels: Tensor):
         reps = [self.sentence_embedder(sentence_feature)['sentence_embedding'] for sentence_feature in sentence_features]
-        return self.batch_hard_triplet_loss(labels, reps[0])
+        repss = torch.cat(reps)
+        labels = labels.repeat(len(reps))
+        return self.batch_hard_triplet_loss(labels, repss)
 
 
     # Hard Triplet Loss


nreimers:
Hi @WangHexie 
Thanks for putting this PR.

But I think you have an issue how the data is represented in your dataset.

The BatchTripletLoss-Functions expects that your dataset looks like this
```
[
InputExample(texts=['First Sentence'], label=0),
InputExample(texts=['Second Sentence'], label=0),
InputExample(texts=['Another Sentence'], label=1),
InputExample(texts=['Another Sentence again'], label=1)
]
```

So each input example should only consists of one texts and the respective label.

Adding multiple texts under InputExample.texts will not work with the other parts of this framework.

If the data looks as intended (i.e. each InputExample is a single example), than BatchTripletLoss, works as intended (correct me if I am wrong).

WangHexie:
After enabling shuffle in DataLoader, I doubt the how large the chance that more than two instances of same class are in a batch could be, since in my use case there're only two or three instances in a class while there're more than 200,000 classes and batch size is relatively small. I use it in information retrieval, each class formed of a query sentence and only a few related documents.
However,  each choice has its pros and cons. My fix is not fit for the dataset that has many instances in a class, the dataset will create much larger batch than the parameter that user set. The simple fix might need more codes afterwards to enable sampling across the instances of same class to adapt the large dataset.

nreimers:
Hi @WangHexie 
Yes, this is an issue.

However, the right approach is to use a data sampler:
https://pytorch.org/docs/stable/data.html#data-loading-order-and-sampler

To construct batches that contain samples with the same label.

Such a sampler is part of the repository:
https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/datasets/sampler/LabelSampler.py

You can add it to your data loader to ensure that there at least 2 examples with the same label in a batch.

By the way, for your case I can highly recommend the MultipleNegativesRankingLoss
https://www.sbert.net/docs/package_reference/losses.html#multiplenegativesrankingloss

Your input examples looks like this:
```
[InputExample(texts=['Anchor 1', 'Positive 1']),
    InputExample(texts=['Anchor 2', 'Positive 2'])]
```

In my experiences, it works much better than the BatchHard strategy if you only have 2 (or a few) examples per class.

WangHexie:
Thanks a lot for your suggestion.

Issue #522, opened 2020-10-26
GPU running OOM when doing predictions


PhilipMay:
Hi,
we are doing predictions on about 4,000,000 texts. We observe that the GPU memory usage goes up and up and then runs OOM.

We checked the code and have the impression that the results remain on GPU until the full prediction on all batched is done.
Then the results are moved to CPU. Could you please check that?

Here are the critical lines:

Prediction loop on batches:
https://github.com/UKPLab/sentence-transformers/blob/3f6655c0ee7cbbc3153a9628ff32bd5f44429846/sentence_transformers/SentenceTransformer.py#L175

Move results to CPU:
https://github.com/UKPLab/sentence-transformers/blob/3f6655c0ee7cbbc3153a9628ff32bd5f44429846/sentence_transformers/SentenceTransformer.py#L197

IMO a fix would be to move the results back to CPU after each batch of prediction. As an optimization you could do that each configurable number of iteration.

PS: If you want to I can work on a PR tonight...

nreimers:
Yes, this is a know thing:
https://github.com/UKPLab/sentence-transformers/issues/487

The issue is that there are different use cases. For some downstream use cases, it make sense to keep the tensors on GPU, for example, if you continue using them (e.g. to compute cosine similarities. There it is better to compute them on the GPU).

What could be done is to check if convert_to_tensors is True and then moving them within the loop to CPU instead of moving them at the end.

PhilipMay:
> What could be done is to check if convert_to_tensors is True and then moving them within the loop to CPU instead of moving them at the end.

I think you mean: _"check if **convert_to_numpy** is True and then moving them within the loop to CPU instead of moving them at the end."_ - right?

nreimers:
Yes, right

PhilipMay:
I started a PR: https://github.com/UKPLab/sentence-transformers/pull/525

PhilipMay:
Fixed - so closing it.

Quetzalcohuatl:
@PhilipMay Your PR only works if convert_to_numpy == True. I just made a PR that allows it when convert_to_numpy == False. Sometimes we want to keep the embeddings still as a tensor. (Or sometimes we want to offload to cpu when we also want to save token embeddings). Check it out

Issue #487, opened 2020-10-09
VRAM use grows when using SentenceTransformer.encode + potential fix. 


martinritchie:
Hey, 

I have been using this repository to obtain sentence embeddings for a data set I am currently working on. When using SentenceTransformer.encode I noticed that my VRAM usage grows with time until a CUDA out of memory error is raised. Through my own experiments I have found the following: 

1. detaching the  embeddings before they are extended to [`all_embeddings`](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L182), using: `embeddings = embeddings.to("cpu")` greatly reduces this growth. 
2. Even with the above line added the VRAM use grew, albiet slowy, by adding the line `torch.cuda.empty_cache()` after the above VRAM usage appears  to stop growing over time. The first point makes sense as a fix but I am unsure why this line is necessary? 

I am using: 
pytorch 1.6.0,
transformers 3.3.1,
sentence_transformers 0.3.7. 

 Have I missed something in the docs or am doing something daft?  I am happy to submit a pull request if needs be? 

Thanks,

Martin

nreimers:
How do you call the encode method? I.e., which parameters do you pass? Especially how to you set convert_to_numpy or convert_to_tensor

martinritchie:
Thank you for a quick response! 

Here is an example: 
```
   model = SentenceTransformer("distilbert-base-nli-stsb-mean-tokens") 
   y = model.encode( text_list, 
                                     batch_size=256,
                                     show_progress_bar=True,
                                     is_pretokenized=False )
```
So I see that  I am using the default values for: `convert_to_tensor=False` and `convert_to_numpy=True`, where `text_list` contain millions of sentences. Shouldn't the embeddings be detached during the for loop define on [line 168](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L168)?

nreimers:
A quick solution would be to break down text_list to smaller chunks (e.g. only 100k sentences) and to append the embeddings afterwards instead of passing Millions of sentences at once.

I have to think if you can detach at line 168 and what implications this will have if you e.g. want the tensors for some downstream application (e.g. as input to some other pytorch model).

martinritchie:
[...]

PR #525, opened 2020-10-26
Fix for OOM problems on GPU at large datasets.

PhilipMay:
see #522 and #487

## TODO
- [x] do testing
- [ ] get review & feedback


PhilipMay:
Internal tests with large data sets show that the code fixed the oom problem.
I made the following changes:

1. I detatch the whole `embeddings` once instead of each single `emb` - this should be more efficient
2. If `convert_to_numpy` I move the whole `embeddings` to cpu.

IMO this PR is ready for review and merging.

nreimers:
Thanks for the work. I will have a look today afternoon and merge it

PR #576, opened 2020-11-19
Fixed link in semantic-search

RafaelWO:
Fixed path to `semantic_search.py` in README of semantic-search


nreimers:
Thanks for the fix :)

PR #578, opened 2020-11-19
Fixed (another) link in semantic-search

RafaelWO:
Fixed path to `semantic_search_quora_pytorch.py` in README of semantic-search. Apparently I missed this one before :see_no_evil:


nreimers:
Thanks for finding these and fixing them 👍

PR #582, opened 2020-11-23
Update CNN.py

omerarshad:
added , map_location=torch.device('cpu')


nreimers:
Great, thanks for this fix

PR #593, opened 2020-11-28
Update bitext_mining_utils.py

yucongo:
I added a `return` in the line `open(filepath, 'r', encoding='utf8')`   in `file_open` for non `gz`/`xz` files.


nreimers:
Great, thanks for the fix

Issue #514, opened 2020-10-22
Change to use module logger instead of the root logger


josemarcosrf:
In `SentenceTransformer.py` mesages are being logged using the root logger as `logging.info` is used.

If a logger specific to the library were to be used would be easier to change the level or formatting of the logger only of SentenceTransformer without having to affect the root logger.

Right now the logging looks like (ignore the formatting as this is specific to my application - just to illustrate it's using the root logger):
```
INFO     root                                         :  38 - Load pretrained SentenceTransformer: distilbert-base-nli-mean-tokens
INFO     root                                         :  42 - Did not find folder distilbert-base-nli-mean-tokens. Assume to download model from server.
INFO     root                                         :  92 - Load SentenceTransformer from folder: /home/jose/.cache/torch/sentence_transformers/sbert.net_models_distilbert-base-nli-mean-tokens
INFO     root                                         : 116 - Use pytorch device: cpu 
```

Instead of `logging.info` doing:

```python
logger = logging.getLogger(__name__)
...
logger.info(....)
```

Should solve the issue.

Happy to submit a PR if this is of interest.

nreimers:
Yes, a PR would be nice.

jmrf:
Oh I missed your reply sorry. Sure, I will submit a PR then!

jmrf:
Quick update on my side. I am still planing to do this, just haven't had the chance to do it yet but will be coming shortly.

nreimers:
Great @jmrf 
thanks for the update

jmrf:
Sorry for the delay! Finally there's a PR for it: https://github.com/UKPLab/sentence-transformers/pull/623

Martins6:
I was trying to debug an issue with the logger and I found this issue. But it seems that the PR was merge and the issue stayed open. Just a friendly reminder that the admin can close it if ti wants to! @nreimers

nreimers:
Thanks @Martins6

PR #623, opened 2020-12-14
Per Module Logger

jmrf:
This PR addresses [issue#514](https://github.com/UKPLab/sentence-transformers/issues/514)

**Main changes**:

- Added `install_logger` function in `LoggingHandler` to simplify the configuration of the library logger
- Added call to `install_logger` function in `__init__.py` to set the library logger properties (level, format, etc) and optionally from a env. variable `ST_LOG_LEVEL` to control the library logging level
- Removed the use of `logging.info` and similar calls for `logger.info`. Each file uses its logger inheriting from the library root logger. 

**Dependency additions**:

- Dependency addition:  `coloredlogs` to enable colored logging based on the logging level. 

This is actually optional, I consider it helpful when reading logs but totally optional and can be removed if prefered.

-----

> **NOTE**: I've run the tests as `pytest -s tests/` both in the `master` branch and the one for this PR. With some test failing. I am unsure if this tests should be run differently.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea
+.vscode
 *.pyc
 *.gz
 *.tsv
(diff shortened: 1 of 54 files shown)


nreimers:
Thanks again for the PR. It is merged into 0.4.1. branch.

I removed the coloredlogs dependency: I try to keep the dependencies for the package as small as possible to ensure that it can run smoothly on most systems.

PR #640, opened 2020-12-21
cos similarirty between embeddings instead of direct MSE

kenkyusha:
Since our embeddings might never achieve very good similarity in terms of MSE (hyperdimensional space), we would much rather prefer the cosine similarity as this is the angular similarity measure.
1. Cosine similarity between the our embeddings and then MSE between the returned similarity measures and label of 1.0

### DIFF
--- a/sentence_transformers/losses/COSLoss.py
+++ b/sentence_transformers/losses/COSLoss.py
@@ -1,0 +1,24 @@
+import torch
+from torch import nn, Tensor
+from typing import Union, Tuple, List, Iterable, Dict
+
+
+class COSLoss(nn.Module):
+    """
+    Computes the cosine similarity between the computed sentence embedding and a target sentence embedding. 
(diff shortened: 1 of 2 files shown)


nreimers:
Thanks for creating this PR.

This PR substantially changes the default behavior of MSELoss, which is not a good thing as other might build  and rely upon this loss.

Instead of having 
```self.loss_fct = nn.MSELoss()```

one could add this parameter as a parameter to the constructor: __init__ 

In that case, people could pass their own loss function as a parameter to MSELoss. This would people give the choice to either use MSELoss or to use a loss based on cosine similarity.

Further, MSE loss and cos-sim-loss should be compared on the knowledge distillation tasks to see if changing the loss improves or decreases the performance.

kenkyusha:
Yeah I guess separate class would be the best...
Would you bale to compare it with MSE metrics?
I also thought L1/MAE might be more interesting than MSE in this particular use-case.

PR #650, opened 2020-12-23
Rename training_OnlineConstrativeLoss.py to training_OnlineContrastiv…

Simaex:
…eLoss.py

Typing error in file name prevent proper hyperlinking from https://www.sbert.net/examples/training/quora_duplicate_questions/README.html#pretrained-models


nreimers:
thanks for the fix :)

PR #652, opened 2020-12-27
Fixing SyntaxError

mhmohona:


nreimers:
Thanks for the fix :)

Issue #631, opened 2020-12-17
model file convert to onnx or pt


codingliuyg:
hello, i have a fine tuned model file,and reduce the dimension by pca. Now，i wanted to convert it to the format of  onnx or pt.
As shown in the figure below。

![image](https://user-images.githubusercontent.com/44610688/102454289-bf059780-4078-11eb-80fb-354d6ded1715.png)

Is there any ways  to get the input and output tensor name of the sentence bert?

@nreimers Looking forward to your reply,thank you in advance!

nreimers:
@codingliuyg 
I sadly never worked with ONNX, so I sadly cannot help here as I don't know what happens

oborchers:
@nreimers : Thanks a bunch for this library. Do you think it is possible to modify the library, so that the forward argument is not a dictionary? Exactly as it is done in the transformers library. Because that would fix the ONNX incompatibility. ONNX can speed up the models by a factor of 2 to 8 on a V100.

While the ONNX export of the model works, specifying dynamic axes (which are required for inputs) does not work with a dict input. Thus, one can only export for a fixed size input, which is useless.

To give a working example for distilroberta-base:

1. Exporting the model:

```
import torch
import transformers

from transformers import convert_graph_to_onnx

from sentence_transformers import SentenceTransformer, util
sent_roberta = SentenceTransformer('msmarco-distilroberta-base-v2', device="cuda")

base_roberta = convert_graph_to_onnx.load_graph_from_args("feature-extraction", "pt", "distilroberta-base", None)
with torch.no_grad():
    input_names, output_names, dynamic_axes, tokens = convert_graph_to_onnx.infer_shapes(base_roberta, "pt")
    ordered_input_names, model_args = convert_graph_to_onnx.ensure_valid_input(base_roberta.model, tokens, input_names)

# Result of the upper assignments:
input_names = ['input_ids', 'attention_mask']
output_names = ['output_0', 'output_1']
dynamic_axes = {
    'input_ids': {0: 'batch', 1: 'sequence'},
    'attention_mask': {0: 'batch', 1: 'sequence'},
    'output_0': {0: 'batch', 1: 'sequence'},
    'output_1': {0: 'batch'}
}
tokens = {
    'input_ids': torch.tensor([[   0,  713,   16,   10, 7728, 4195,    2]]).long(), 
    'attention_mask': torch.tensor([[1, 1, 1, 1, 1, 1, 1]]).long()
}

[...]

PR #386, opened 2020-08-27
Adding example using ONNX model to get sentence embeddings

SidJain1412:

Issue #247, opened 2020-05-26
Doubt about distiluse-base-multilingual-cased


Guillermogsjc:
HI @nreimers,

first congratulations for the library and the papers about sentence embeddings.

I have a doubt about `distiluse-base-multilingual-cased`. Why the output tensor dimension for one sentence  is  **512** instead typical 768? Is not it trained with DistiBERT?

Regards.

nreimers:
Hi @Guillermogsjc 
It is based on DistilBERT, however, with a dense layer, the 768 dimension output were down projected to 512 dimensions.

The model is a distillation of multilingual Universal Sentence Encoder which uses 512 dimension, hence, this down projection was necesary.

Best
Nils Reimers

Guillermogsjc:
Thank you very much for the quick response.

Kind regards.

Guillermogsjc:
Hi @nreimers ,

another question is, it is usable this model as it were a transformer from hugginface collection? In that case do you plan to add it to the pool?

The use case is to use it for *cold embeddings* work like dataset creation etc, then transfer with a softmax or multisigmoidal layer for a text classification problem for example.

Regards.

nreimers:
Hi @Guillermogsjc 
I am working on it to add the models to the huggingface models collection. I should be able to add them within the next 2 weeks.

Best
Nils

bernardocaldas:
Hi @nreimers , thanks for your work! Are you still planning to add the models to the huggingface collection?

Best,
Bernardo

nreimers:
Hi @bernardocaldas 
Sorry to be behind schedule. Yes, models will be added to the models collections. Hope I can do that soon.

Best
Nils

Ierezell:
Hi, @nreimers 

I have a pipeline for a company to save huggingface model in onnx format and run them in NodeJs. Mostly thanks to hugging face models, scripts for converting and the rust tokenizer repo with bindings in Node.  

However, for semantic search, the best results on french text were with the distilUSE sentence bert model you gave. I tried to add a custom hugging face french model (camembert) but I lost perfomance.

I would like to know if adding the models to hugging face can come soon so I will export it from HF because it's really easy and run it in Node. Else if you have another way to pass your SBERT model to onnx i'm good with it. 

If you need any help to port SBERT to HF I would be glad to help on a PR.

nreimers:
Hi @Ierezell 
[...]

Issue #46, opened 2019-11-04
sentence bert model in onnx format


rachel2011:
I would like to convert sentence bert model from pytorch to tensorflow use onnx, and tried to follow the standard onnx procedure for converting a pytorch model. But I'm having difficulty determining the onnx input arguments for sentence bert model, I encounter `TypeError: forward() takes 2 positional arguments but 4 were given.` Suggestions appreciated!
`model = SentenceTransformer('output/continue_training_model')`
 `device = torch.device("cuda" if torch.cuda.is_available() else "cpu")`
 `model.to(device)`
 `dummy_input0 = torch.LongTensor(batch_size, max_seq_length).to(device)`
 `dummy_input1 = torch.LongTensor(batch_size, max_seq_length).to(device)`
 `dummy_input2 = torch.LongTensor(batch_size, max_seq_length).to(device)`
 `torch.onnx.export(model,(dummy_input0, dummy_input1,dummy_input2), onnx_file_name, verbose=True)`

nreimers:
Sadly I never worked with onnx.

In SentenceTransformer, the forward function takes in one argument: features (and the second in python is self).

Features is a dictionary, that contains the different features, for example, token ids, word weights, attention values, token_type_ids.

For the BERT model, I think your input must look like this:
input_features = {'input_ids': dummy_input0, 'token_type_ids': dummy_input1, 'input_mask': dummy_input2}

And then:
```
torch.onnx.export(model,input_features, onnx_file_name, verbose=True)
```

tuner007:
@rachel2011 were you able to convert pytorch to onnx ?

ycgui:
@rachel2011 Did you get any solutions to successfully convert sentence bert model to onnx format?

I'm also wondering how to feed text inputs into a converted onnx model. Could we do something similar as 
```
sentences = ['This framework generates embeddings for each input sentence',
    'Sentences are passed as a list of string.', 
    'The quick brown fox jumps over the lazy dog.']
sentence_embeddings = model.encode(sentences)
```
by replacing the model to the converted onnx model? Any ideas?

nreimers:
Hi @ycgui 
I started to add the models to HuggingFace Models Hub:
https://huggingface.co/sentence-transformers

Huggingface also provides methods / scripts to convert models to ONNX.

I hope this helps.

ycgui:
Thanks @nreimers. This is awesome!

SidJain1412:
[...]

PR #668, opened 2021-01-06
Added Notebook for ONNX Inference Model Creation

oborchers:
This PR addresses the requests ( #631 #46 #247 #46 ) to provide support for exporting ONNX models from the SentenceTransformers library.

This notebook shows how to create an ONNX model from the models in transformers model hub and adds a simple pooling layer on top of it to drop all external torch dependencies.

This notebook addresses shortcomings of PR #386, in which the pooling layer was not integrated into the ONNX model.


oborchers:
I've added another benchmark comparing batched vs online encoding speed, because I was wondering how you did achieve speeds of more than a thousand samples per second. In the batch conversion case the speed increase is only by factor of 1.3 to 2x for bert-base-nli-stsb-mean-tokens (2000 to 4000 sentences / s). But this is only applicable for shorter spans and diminishes with longer spans to 1.3x.

tomaarsen:
I've added native ONNX support in Sentence Transformers, so users can now look at the [Speeding up Inference](https://sbert.net/docs/sentence_transformer/usage/efficiency.html#onnx) documentation, under the ONNX section.

Thanks for making this PR - it helped show me the importance of different backends.

- Tom Aarsen

PR #692, opened 2021-01-15
Update semantic_search.py to handle corpus <5

George-T-Henry:
I was adjusting the `top_k` value frequently when doing quick tests with a corpus smaller than 5 using this script. With this change, `top-k` value will revert to the length of the corpus if it is under 5 automatically. The print message reflects the same.
(thank you for all of this 八)


nreimers:
Thanks for pointing this out.

I added a check to the example (but in a slight different way by using min() instead of if-else).

PR #713, opened 2021-01-24
fix typos in readme 

sid-sundrani:
There are two typos in the [readme file](https://github.com/UKPLab/sentence-transformers/blob/master/examples/training/sts/README.md) of the example of training Semantic Textual Similarity model 

I've accordingly suggested corrections in this PR


nreimers:
Thanks for fixing these :)

Issue #716, opened 2021-01-26
unsupported argument output_path_ignore_not_empty not updated in examples 


sid-sundrani:
In the examples of sentence transformers, the fit methods are still taking in `output_path_ignore_not_empty` as input. This argument has been deprecated and removed previously- [commit](https://github.com/UKPLab/sentence-transformers/commit/0babd8bf6872daf02b4c02d4e0690db6fe2f8472) 

It would be great if the fit method in many of the [examples](https://github.com/UKPLab/sentence-transformers/tree/de558ab054f38f7dac30845f04683cf18b447750/examples/training) ([1](https://github.com/UKPLab/sentence-transformers/blob/de558ab054f38f7dac30845f04683cf18b447750/examples/training/data_augmentation/train_sts_seed_optimization.py), [2](https://github.com/UKPLab/sentence-transformers/blob/de558ab054f38f7dac30845f04683cf18b447750/examples/training/quora_duplicate_questions/training_OnlineContrastiveLoss.py) are some) would remove this argument to avoid errors

nreimers:
Thanks for fixing it :)

PR #717, opened 2021-01-26
remove unsupported arg output_path_ignore_not_empty from examples 

sid-sundrani:
The `fit` methods in these examples are still taking in `output_path_ignore_not_empty` as input. This argument of `fit` has been deprecated and removed previously. Hence, it's leading to errors when installed from source 

I removed this as an argument in the fit methods of the examples that were using them 

Issue #716


nreimers:
Great, thanks for spotting this and fixing it.

PR #748, opened 2021-02-08
Update index.rst

lispsil:
fix broken url


nreimers:
great, thanks for the fix

PR #765, opened 2021-02-18
add train_mode to encode

fulaphex:
I'm using the `SentenceTransformer` as a layer in a model I'm working on. I want to finetune the transformer layer in the training, so I need this modification to the `encode` function so that gradients are being computed.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import shutil
+from contextlib import nullcontext
 from collections import OrderedDict
 from typing import List, Dict, Tuple, Iterable, Type, Union, Callable
 from zipfile import ZipFile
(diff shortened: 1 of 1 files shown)


nreimers:
Hi @fulaphex 
Thanks for proposing this PR.

Not sure if a 'train_mode' flag in the encode function would be the best way. 

The SentenceTransformer object inherits from nn.Sequential, i.e., it has a regular forward function which can by used within pytorch.

You can get a trainable sentence embedding like this:
```python
model = SentenceTransformer('stsb-roberta-base')


batch = ["Hello World", "This is a test"]

#Get the features for the batch, i.e. tokenize them
features = model.tokenize(batch)

#Move features to the device
features = util.batch_to_device(features, model.device)

#Call the forward function of the model
output = model(features)

#The model returns various outputs, where 'sentence_embedding' is your sentence embedding
embeddings = output['sentence_embedding']
print(embeddings)
```

PR #776, opened 2021-02-24
Fix typo in msmarco-v2.md

Omarfos:


nreimers:
great, thanks for the fix

PR #779, opened 2021-02-25
Fix spelling error

Omarfos:


nreimers:
thanks for the fix :)

PR #806, opened 2021-03-11
Fix broken link in README

nate-bush:
Relative link in README was broken. I fixed it.


nreimers:
Great, thanks for fixing it.

PR #816, opened 2021-03-16
Allow `batch` to be a tuple in `batch_to_device`

FremyCompany:
While evaluating using the `LabelAccuracyEvaluator`, the batch are not dictionaries, but rather a tuple containing a dictionary as first value, and a tensor as second value. This causes `batch_to_device(...)` to fail with error `TypeError: tuple indices must be integers or slices, not list`.

I fixed `batch_to_device` to be more tolerating in what types it accepts. Possibly `LabelAccuracyEvaluator` could use a refactor to avoid this scheme, but I think the solution I propose has the advantage of always outputting a clear error message.

### DIFF
--- a/sentence_transformers/util.py
+++ b/sentence_transformers/util.py
@@ -241,10 +241,19 @@ def batch_to_device(batch, target_device: device):
     """
     send a pytorch batch to a device (CPU/GPU)
     """
-    for key in batch:
-        if isinstance(batch[key], Tensor):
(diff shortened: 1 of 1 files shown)


nreimers:
Thanks for the PR.

The LabelAccuracyEvaluator works in the current master branch and a working version will be part of the next release.

As the batch_to_device is quite a critical function, I prefer not to change it. Also it enforce other methods to create batches with the right format. This avoids errors down the line.

PR #842, opened 2021-03-29
Add path for tokenizer if its name is different than model name

adbelniak:
I have tried to use SBERT with herBERT. It's required to load tokenizer which has different name than model. I added an option for that


nreimers:
Thanks for the commit and the idea.

It is sadly not working completely if you use different type of tokenizer, e.g. you have a BERT model that requires a RoBERTa tokenizer. It will work when you create the model, but saving & loading from disc will sadly not work as Huggingface Transformers will try to load a BERT tokenizer the next time.

Will fix it and integrate it.

PR #855, opened 2021-04-04
fix: corpus size in example does not align with comment

mathislucka:
Just a minor one:

I realized that the comment says max dev corpus size should be 10k. The actual code put it at a 100k.


nreimers:
Thanks for fixing it :)

PR #864, opened 2021-04-12
Imports fix in tokenizer __init__.py

vladimir-rigvava:


nreimers:
Sorry for my late reply.

Thanks for this fix. It will be part of the next release.

PR #908, opened 2021-05-04
[Draft] Hugging Face model hub integration

osanseviero:
This integration allows easily using models from the Hugging Face hub.

**Current State**
Today, users can either use local models or use models from the server. If both fail, a model is built using `Transformer`. 

**Proposal**
The models will now be hosted in the Hugging Face hub, which comes with many benefits such as all HugginFace users being able to use the models, free storage, versioning, etc. 

There are two directory structures supported thanks 

- New: Base model files are in the main directory. Pooling and additional layers will be in sub-directories. This allows `Transformers` users to use `from_pretrained` easily and add their custom pooling afterwards, which is the current custom use.
- Old: Base model files are in `0_Transformer` directory. This ensures everything still work if users have the old directories.

The new structure requires changing the path of the base in `modules.json` to `"path": ""`,

**Tested**
(Note, osanseviero should be changed to `sentence-transformers` once the models are ready in the organization)
```
# Sentence Transformer Model from organization with modules.json
model = SentenceTransformer('osanseviero/full-sentence-distillroberta2')

# Sentence Transformer Model adding pooling layer to global model
model = SentenceTransformer('bert-base-cased')

# Transformer Model which can also use the uploaded model
model = AutoModel.from_pretrained('osanseviero/full-sentence-distillroberta2')
```

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -330,7 +330,7 @@ class SentenceTransformer(nn.Sequential):
 
         for idx, name in enumerate(self._modules):
             module = self._modules[name]
-            model_path = os.path.join(path, str(idx)+"_"+type(module).__name__)
+            model_path = path if idx == 0 else os.path.join(path, str(idx)+"_"+type(module).__name__)
             os.makedirs(model_path, exist_ok=True)
             module.save(model_path)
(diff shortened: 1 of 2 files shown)


osanseviero:
This is not complete, this still requires changing the save method to keep same structure.

osanseviero:
@nreimers Feel free to take a look :)

nreimers:
This PR is integrated in the v2 branch:
https://github.com/UKPLab/sentence-transformers/tree/v2_dev

PR #925, opened 2021-05-12
Allow pushing models to the HuggingFace hub

osanseviero:
Here is an example of a model uploaded with this method 
https://huggingface.co/osanseviero/upload-to-hub/tree/main

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -3,13 +3,13 @@ import logging
 import os
 import shutil
 from collections import OrderedDict
-from typing import List, Dict, Tuple, Iterable, Type, Union, Callable
+from typing import List, Dict, Tuple, Iterable, Type, Union, Callable, Optional
 from zipfile import ZipFile
 import requests
 import numpy as np
 from numpy import ndarray
 import transformers
-from huggingface_hub import snapshot_download, hf_hub_url, cached_download
+from huggingface_hub import HfApi, HfFolder, Repository, snapshot_download, hf_hub_url, cached_download
 import torch
 from torch import nn, Tensor, device
 from torch.optim import Optimizer
@@ -18,6 +18,8 @@ import torch.multiprocessing as mp
 from tqdm.autonotebook import trange
 import math
 import queue
+import tempfile
+from distutils.dir_util import copy_tree
 
 from . import __MODEL_HUB_ORGANIZATION__
 from .evaluation import SentenceEvaluator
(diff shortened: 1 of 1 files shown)


Kvit:
Is it possible to produce sentence embeddings or, even better, a similarity score between two sets of sentences using Huggingface inference API after the model is uploaded to the Huggingface hub?

nreimers:
Hi @Kvit 
We are currently working on this and hope to integrate it soon into the inference API.

Currently we look for user feedback for the best integration. Would be great to hear from you:
- What you would like to pass as input (one set of sentences, two set of sentences, some other data structure)?
- What you would like to get as output (the raw embeddings, some cosine similarity scores?)

Kvit:
Thanks for asking :). Input would be an array of sentences `input:{params:{...},sentences:[sent1, sent2, sent3] }` the results could be an array of similarity scores `[{ind1, ind2, score}, {ind1,ind3,score}]`. For a large sentence array, the source may be a blob object in a storage bucket.   I will be happy to beta test.

osanseviero:
Hey @nreimers. I saw you merged the other PR. I think this one had two super-small nit commits at the end related to incomplete docstrings that we should merge as well. Feel free to make the change directly, just wanted to let you know :)

Sorry for missing this in the other PR!

osanseviero:
Hey @Kvit. I got some news for you!

Here is a full-working example of `sentence similarity` working! With the v2 release, we still need to update the repos from the `sentence-transformers` organization repos with the latest tags and adding info on additional layers, but this will already work for you if your use case is mean pooling after base transformer! Let me know your thoughts :) 

Currently, this is using cosine similarity. We'll add params to `sentence-similarity` so users can easily specify other distance metrics. We just open-sourced the inference [repo](https://github.com/huggingface/huggingface_hub/blob/main/api-inference-community/docker_images/sentence_transformers/app/pipelines/sentence_similarity.py) in case you're interested: 

```
import json
import requests

API_URL = "https://api-inference.huggingface.co/models/osanseviero/full-sentence-distillroberta2"
headers = {"Authorization": "Bearer your_token"}

def query(payload):
	data = json.dumps(payload)
	response = requests.request("POST", API_URL, headers=headers, data=data)
	return json.loads(response.content.decode("utf-8"))

data = query(
    {
        "inputs": {
            "source_sentence": "That is a happy person",
            "sentences": ["That is a happy dog",
                          "That is a very happy person",
                          "Today is a sunny day"]
        }
    }
)

print(data)
# [0.6984518766403198, 0.9128190279006958, 0.2244246006011963]
```

Kvit:
Awesome, we will give it a try right away,  Can I use my own Roberta model
that was trained on MLM ? I have it hosted in a private repo on
transformers hub.
I'm still a little hazy on the magic that  transformers Inference is using
to make task-specific pipeline out of the model although I got a lot of
secret knowledge from the Hugginface team lately. I know I can either use
task-specific API endpoint eg /task-pipeline/{repo}/{model} or I can use
pipleline_tag: in the model card to make the magic. Any that will work with
sentence transformers?

Vitali

osanseviero:
Yes, this should work with your own models and even if hosted privately :). The only thing you need to do is specify the task you want to get solved in the model card (see [here](https://huggingface.co/osanseviero/full-sentence-distillroberta2/blob/main/README.md)). The task inferred in the inference API is from there and that's why in the example above I could do `/models/osanseviero/full-sentence-distillroberta` instead of `/pipeline/feature-extraction/osanseviero/full-sentence-distillroberta`. 

Let me know if you have any problems and excited for this!
Omar

Kvit:
Thanks a lot, this is great, it should solve a major problem: the need for
GPU on client node to use model fast.
I had a problem using inference with custom pipelines with pinned models
lately (see here
<https://github.com/huggingface/transformers/issues/11913#issue-904492548>),
we will test the method you described I will let you know of any issues.
Many thanks!

Vitali

Kvit:
Is there a way to define a pipeline that makes just sentence embeddings? In
my case, I need to compare sentences from medical text against the same
diagnostic key phrases (about 100 phrases), and it does not make sense to
send these 100 key phrases via API every time. It is better to embed key
phrases once and do cosine similarly of text embeddings locally.

Kvit:
I did the test using my Roberta model `ligolab/DxRoberta` and got the error:
```
text input must of type `str` (single example), `List[str]` (batch or single pretokenized example) or `List[List[str]]` (batch of pretokenized examples)
```

Kvit:
I was able to overcome the error by forcing a specific pipeline in API endpoint. `https://api-inference.huggingface.co/pipeline/sentence-similarity/ligolab/DxRoberta`

osanseviero:
Hi @Kvit. Sorry for the confusion. There is some custom logic that sometimes infers the task based on the model (your model was Roberta so it was inferring `fill-mask`). 

I went ahead and added `pipeline_tag: sentence-similarity` to the model card as recommended in the issue you linked. I think this should work now without specifying the task in the url.

If you want to only get the embedding vectors, use `feature-extraction` instead. Please let me know if this works! :)

Kvit:
if I use `pipeline_tag: feature-extraction` on model `ligolab/DxRoberta` i get the usual Roberta embedding vectors, one for each word in the sentence. feel free to try.

osanseviero:
Hi @Kvit. Could you try again with some sentences? The issue should have been fixed now.

To provide context, the model was pinned which loads the model. You then changed the tag, which do not update the model. This is an issue on our side.

Anyways, I just tried it out and I'm getting a sentence embedding as expected. Let me know if it works for you :)

Kvit:
It works thanks! my only wish is that the sentence embeddings pipeline could take an array of strings (not just a single string) and return the array of embeddings, this way we can pass the entire paragraph in one API call and reduce the overhead.

osanseviero:
@Kvit unrelated news: you can now use the `feature-extraction` and `sentence-similarity` tags in your repo and you will get a working widget with which you can make direct inference in the browser! :)

Kvit:
@osanseviero, nice! will we be able to use an array of sentences for an API request or it will remain one sentence per API call?

osanseviero:
Hi Kiev. At the moment we're not supporting batched input in the community API, but please feel free to open a issue at the corresponding repo to continue the discussion or ask for this: https://github.com/huggingface/huggingface_hub

PR #933, opened 2021-05-16
Fix LSTM layer to work around https://github.com/pytorch/pytorch/issues/43227

fros1y:
Recent Torch versions require lengths input for `nn.utils.rnn.pack_padded_sequence` to be on the CPU.  Change just adds a `.cpu()` to ensure right location.  See https://github.com/pytorch/pytorch/issues/43227


lishuaiCN:
still error as you guide, can you show further details @fros1y

tomaarsen:
Hello!

This has been fixed in #1420, which proposed the same fix as this PR. I'll include this in the next release.

- Tom Aarsen

PR #934, opened 2021-05-16
Add optional dropout layer to Dense

fros1y:


nreimers:
Thanks for the PR.

Did you do some tests with it?

I was thinking if if might make more sense to have the dropout layer before the linear layer instead of after the linear layer? Also maybe if make sense to make the dropout an independent layer, so that you can position it anywhere where you want?

fros1y:
Agreed, will submit a new pull request with a Dropout that is an independent layer.

PR #938, opened 2021-05-18
Add dropout layer

fros1y:
Added independent dropout layer.  Intention is for it to be used between layers of a DAN, but suitable for any use after the Pooling layer, once a "sentence_embedding" is formed from individual tokens.


nreimers:
Thanks for the PR. It will be part of version 2 of the framework.

PR #953, opened 2021-05-25
Fix-up clustering README

chrisji:


nreimers:
Great, thanks for finding this and fixing it :)

PR #961, opened 2021-05-27
Fixed off by one error in ContrastiveTensionDataLoader

seanMahon:
The ContrastiveTensionDataLoader throws an IndexError on the last batch if len(self.sentences) % batch_size = batch_size - 1. 

This comes from the while loop body incrementing sentence_idx from len(self.sentences) -1 to len(self.sentences) and attempting to access self.sentences[sentence_idx]. Changing the while loop condition to "while sentence_idx + 1 < len(self.sentences)"  prevents this from happening and ends the loop at the correct time.


nreimers:
Thanks for finding the issue.

I merged it and it will be part of v2.

PR #971, opened 2021-05-31
Fix typo in example for SoftmaxLoss

kddubey:
Remove extra `InputExample(`


nreimers:
Thanks for the fix, it will be fixed in the next version :)

PR #976, opened 2021-06-02
fix typo 'independet'

Quetzalcohuatl:


nreimers:
Thanks for the fix :)

PR #991, opened 2021-06-09
Fix typos

lukesalamone:


nreimers:
Great, thanks for finding these and fixing them :)

PR #1073, opened 2021-07-16
Gradio Web Demo Image Search

AK391:
Web Demo on Gradio Hub for Image Search

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ We provide an increasing number of **[state-of-the-art pretrained models](https:
 
 Further, this framework allows an easy  **[fine-tuning of custom embeddings models](https://www.sbert.net/docs/training/overview.html)**, to achieve maximal performance on your specific task.
 
+***New***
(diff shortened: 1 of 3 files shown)


nreimers:
Thanks for creating this cool demo.

The PR has some issues, e.g. you add 'sentence-transformers' as a requirement for 'sentence-transformers' => Not sure what will happen here.

Instead of merging it here, I think the better place would be to merge it into a new feature we have at huggingface: [Spaces](https://huggingface.co/docs/hub/spaces)

Then, the demo can run on the HF infrastructure  and the demo would be listed here:
http://hf.co/sentence-transformers

Would this be OK for you?

=============

On major re-work of the documentation is on my roadmap, to be more centered around demos. Your demo would then also be included.

AK391:
@nreimers that would be fine, I think the spaces is still in beta and invite only currently? I can remove the sentence transformer requirement

tomaarsen:
Hello!

It's nice to see an old PR for Gradio <3 . I take it that this is now wildly outdated? I'll close it, but let me know if you are still interested in this work!

- Tom Aarsen

PR #1085, opened 2021-07-29
Add batch prediction to CERerankingEvaluator

hugoabonizio:
This PR adds a parameter to `CERerankingEvaluator` to make batch predictions, keeping the current behavior by default of predicting one sample at a time. I made a comparison using my private dataset and the validation time jumps from ~41min using `batch_size=1` to ~3min with `batch_size=128`.

### DIFF
--- a/sentence_transformers/cross_encoder/evaluation/CERerankingEvaluator.py
+++ b/sentence_transformers/cross_encoder/evaluation/CERerankingEvaluator.py
@@ -2,8 +2,14 @@ import logging
 import numpy as np
 import os
 import csv
+from itertools import zip_longest
 
 logger = logging.getLogger(__name__)
+
+# https://docs.python.org/3/library/itertools.html#itertools-recipes
+def grouper(iterable, n):
+    args = [iter(iterable)] * n
+    return zip_longest(*args)
 
 class CERerankingEvaluator:
     """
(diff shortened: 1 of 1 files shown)


nreimers:
Thanks for the PR.

Are you sure that the evaluator still computes the same? A reduction of 41min to 3min looks too extreme. For this evaluator, you usually have several passages per query, e.g. 100 passages per query you want to re-rank. Then batch encoding is used on these 100 passages.

By increasing the batch size and having (query, passage) pairs encoded from different queries in the same call should not really lower the runtime. Maybe you win 10%, but the validation should not jump from 41min to 3min.

hugoabonizio:
Hi @nreimers! I forgot to mention that my dataset contains only one positive and one negative example per query, thus predicting one at a time makes the GPU usage low during validation, and that's my main motivation for this feature. I understand that other datasets might not benefit from this behavior, so I kept the default when the parameter is not used.

However, the code wasn't computing the same, I forgot to use the same `for` loop to compute the MRR. I fixed it and now it computes the same values as the original `CERerankingEvaluator`.

But I'm in doubt about the condition `if len(positive) == 0 or len(negative) == 0: continue`, is it really required to compute the MRR?

nreimers:
When there are no positives or no negatives, computing MRR does not really make that much sense.

If len(negative) == 0 => MRR is always 1, independent how you rank
If len(positive) == 0 => MRR is always 0, independent how you rank

hugoabonizio:
@nreimers I conducted an experiment with MS-MARCO and the original code, the new version with batch size 1 and batch size 128 are all resulting in a MRR@10 of `0.726609126984127`, and all took 45s. So MS-MARCO will not benefit much from this change, but datasets with few examples per query might have nice gains.

The experiment script (extracted from [train_cross-encoder-v2.py](https://github.com/UKPLab/sentence-transformers/blob/8e3768477504946d22ae0406def9c5763311e712/examples/training/ms_marco/train_cross-encoder-v2.py)): https://gist.github.com/hugoabonizio/e958c8bcc196bdf602d389b0bd6b6834

tomaarsen:
We've added `batch_size` support to the `predict` call; I'd like to stick with that approach for now, I think.

- Tom Aarsen

PR #1089, opened 2021-08-01
Adding support for using weights for training in softmax loss

peyajm:
Adding optional argument weight (a 1D Tensor assigning weight to each of the classes), particularly useful when there is an unbalanced training set.


nreimers:
Hi,
thanks this suggestion / PR.

I added a solution in a recent commit. Instead of adding a weight parameter, you can now pass your custom loss_fct. This allows to set / change all the parameters of nn.CrossEntropyLoss().

It can be used like this:
```
train_loss = losses.SoftmaxLoss(model=model, sentence_embedding_dimension=model.get_sentence_embedding_dimension(), num_labels=len(label2int), loss_fct=nn.CrossEntropyLoss(...))

```

peyajm:
@nreimers Thanks for the commit - it looks quite good and useful. I'm glad you found my suggestion helpful.

PR #1092, opened 2021-08-02
Add gradient accumulation to CrossEncoder

hugoabonizio:
This PR adds gradient accumulation to CrossEncoder training, which might be very useful when using larger models or training with larger batch sizes.


tomaarsen:
Apologies for the delay. This behaviour has been added in the Sentence Transformers v4.0 release, which introduced the CrossEncoderTrainer.

- Tom Aarsen

PR #1103, opened 2021-08-10
Pass model and tokenizer arguments when instantiating `SentenceTransformer`

### LINKED ISSUE #794: Experiencing https://github.com/huggingface/tokenizers/issues/537 issue when sentence-transformer is used for generating embeddings
Hello, 
I am using sentence-transformer to get the text embeddings using SentenceTransformer.encode(). 
This function is getting invoked from multi-threaded program. 
I see sometimes this encode method fails with 'Already Borrowed' exception. 
The issue is same or similar to https://github.com/huggingface/tokenizers/issues/537. 
My sentence-transformer version is  0.4.1.2.

Which version of sentence-transformer will not have this issue? Or what is the alternative to resolve this issue?

aphedges:
Fixes #794.

The `Transformer` class already has `model_args` and `tokenizer_args` arguments in its constructor to allow overriding when instantiating models and tokenizers from the `transformers` library. However, there is no way to set these from the `SentenceTransformers` class, which is the one that many users would be instantiating instead. I have added arguments to the `SentenceTransformers` constructor to allow these to be more easily used.

My specific motivation in this case is to make disabling the fast tokenizer easier, but it should work for any arguments that users might want to override.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -44,11 +44,18 @@ class SentenceTransformer(nn.Sequential):
                  modules: Optional[Iterable[nn.Module]] = None,
                  device: Optional[str] = None,
                  cache_folder: Optional[str] = None,
-                 use_auth_token: Union[bool, str, None] = None
+                 use_auth_token: Union[bool, str, None] = None,
+                 model_args: Optional[Dict] = None,
+                 tokenizer_args: Optional[Dict] = None,
                  ):
         self._model_card_vars = {}
         self._model_card_text = None
         self._model_config = {}
+
+        if model_args is None:
+            model_args = {}
+        if tokenizer_args is None:
+            tokenizer_args = {}
 
         if cache_folder is None:
             cache_folder = os.getenv('SENTENCE_TRANSFORMERS_HOME')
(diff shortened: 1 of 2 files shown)


nreimers:
Thanks for the PR. I will have a look after my vacation (after Aug. 23rd)

aphedges:
@nreimers, I'm wondering if you have time to look at this PR now.

nreimers:
@aphedges Sadly not yet, sorry for that. I'm currently work on a major release of new models. Once that is done, I will have a look  the PRs and integrate new functions into SBERT

aphedges:
@nreimers, understood! Thanks for the update!

aphedges:
@nreimers, sorry to keep bothering you about this, but I was hoping you could take a look at this PR. If there is anything I can do to make it easier to review (e.g. splitting it up into multiple PRs), please let me know.

nreimers:
Sorry that I did not yet  look into it.

Could you write a bit more about the use-case? The use case is when you run SentenceTransformers in a multi-process setup to disable the fast tokenizer? The linked issue in tokenizer appears that this is still an issue, or?

aphedges:
It's okay! I know you have other priorities when it comes to this library.

Your recounting of the use case is correct. My specific use case is running SentenceTransformers in a multi-threaded setup (e.g. Flask). As far as I'm aware, the underlying problem will always be present because of differences between the CPython and Rust runtimes. This allows me to disable the fast tokenizer by just passing in a parameter.

I specifically made this fix more general than needed to allow any model or tokenizer arguments to be passed when the model and tokenizer are loaded. There are probably other use cases for this PR that I am unaware of.

aphedges:
@nreimers, just pinging this again.

aphedges:
I have rebased this branch against `master` to resolve conflicts.

tomaarsen:
Hello!

Apologies for the delay. This has been implemented via #2578 and released in the v3.0 release last week. See the documentation for it [here](https://sbert.net/docs/package_reference/sentence_transformer/SentenceTransformer.html#sentence_transformers.SentenceTransformer).

- Tom Aarsen

PR #1113, opened 2021-08-13
Use float for mlm_prob

Ningchen-Wang:
mlm_probability param of DataCollator accepts 0-1 float not integer
https://huggingface.co/transformers/main_classes/data_collator.html?highlight=datacollator#datacollatorforlanguagemodeling


nreimers:
Thanks for the fix. Not sure why the value was 15 and not 0.15

PR #1114, opened 2021-08-14
Display correct pooling mode in model card comment

aphedges:
The model card creation would always include a comment saying `max pooling`, even if another pooling mode was used. This PR makes sure it is substituted along with the pooling function and pooling function name.


nreimers:
Great, thanks for fixing this :)

aphedges:
You're welcome! Thanks for the useful library!

PR #1116, opened 2021-08-16
Fix downloading models from Hub

osanseviero:
The current approach has some issues when downloading repos from the Hub. The current approach skips downloading if a local directory is already present. This means that local files can easily get outdated if there are changes in the repo in the Hub. Users only get the latest model if they delete the directory from the cache. Additionally, if one file is missing (such as `modules.json`, it doesn't get automatically downloaded).

With this new approach, we always call `snapshot_download` and directly save in the expected directory instead of renaming the directory afterwards. It has this behavior:
 * If the file already exists in the cache, it won't get re-downloaded. 
 * If a file is missing locally, it gets downloaded.
 * If a file is updated in the Hub, it also gets downloaded.

This ensures that the local files are updated but maintaining the benefits of client-side caching.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -76,15 +76,12 @@ class SentenceTransformer(nn.Sequential):
 
                 model_path = os.path.join(cache_folder, model_name_or_path.replace("/", "_"))
 
-                if not os.path.exists(model_path):
-                    # Download from hub
-                    model_path_tmp = snapshot_download(model_name_or_path,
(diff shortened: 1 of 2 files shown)


osanseviero:
Pinging @nreimers since this bug will likely hit again in some time.

nreimers:
Hi @osanseviero 
Sorry they I have not yet reacted. Was quite busy in the last days to finalize the release of the new models.

Don't worry, I did not forgot this PR. Will try to have a look next week.

nreimers:
Thanks, the fix works well :)

PR #1133, opened 2021-08-23
Fix json serialization bug in fit() function

### LINKED ISSUE #1034: [2.0] Bug with `.fit` model card - `json.dumps` fails when `callback` is not None
https://github.com/UKPLab/sentence-transformers/blob/2e7e20880f9a4a9969bc1684eec393263fb697bc/sentence_transformers/SentenceTransformer.py#L612

Fails because python functions by default are not JSON serializable. I thought of submitting a fix with `fullname` but that seems to be only for methods on a class.

lukesalamone:
Fixes #1034


lukesalamone:
@nreimers

nreimers:
Thanks for the fix :)

PR #1141, opened 2021-08-30
Typo in "pharaphrases"

aprikyan:


nreimers:
Thanks for fixing it :)

PR #1148, opened 2021-09-02
Fix typos in docs

yks72p:
Hello! I've found a couple of typos in documentation. 
I've also fixed them to be correct in my understanding of corresponding phrases meaning (please correct my edits if they are wrong).


nreimers:
Thanks for fixing them :)

PR #1176, opened 2021-09-20
minor typo fix

Ishan-Kumar2:


nreimers:
Thanks for the fix :)

Ishan-Kumar2:
@nreimers no problem!
Also I am looking to contribute more to sentence-transformers let me know if there is some issue I can help out with thanks :)

PR #1184, opened 2021-09-24
link fix

oztalha:
~s/cross-encoders/cross-encoder/


nreimers:
Thanks for the fix :)

PR #1191, opened 2021-10-01
docs: typo and simplify code

sbrugman:
- f-strings since we support py3.6+
- typo "similarities"
- basic padding to outline printed results
- dict comprehension


nreimers:
Thanks for this PR. I fixed the typos.

Regarding the examples, I try to keep them as simple as possible and I have the feeling not everyone yet knows about the f-string.

PR #1242, opened 2021-11-02
Update Author name in wikipedia-sections-models.md

elronbandel:
The last name of the author is Ein Dor


nreimers:
Thanks for fixing this :)

PR #1253, opened 2021-11-07
util.community_detection : init_max_size should not exceed the dataset size

dingusagar:
Added a fix for init_max_size argument when the size of the dataset is less than the default value 1000 for init_max_size.


nreimers:
Thanks for the PR :)

PR #1257, opened 2021-11-09
Fix `semantic_search()` docstring

aphedges:
The return value for `semantic_search()` did not match what was documented, so I fixed the docstring. My coworker (who noticed this discrepancy) @spigo900 helped with the new wording.

I also made some typo fixes in the file that I noticed while editing it.


nreimers:
Thanks for fixing this and making the docs better :)

PR #1352, opened 2021-12-31
Restrict version range of tokenizers

rom1504:
This is needed to avoid the error
transformers==4.15.0 requires tokenizers<0.11,>=0.10.1 but tokenizers 0.11.1 was resolved
when installing sentence-transformers

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ setup(
     packages=find_packages(),
     install_requires=[
         'transformers>=4.6.0,<5.0.0',
-        'tokenizers>=0.10.3',
+        'tokenizers>=0.10.3,<0.11',
         'tqdm',
         'torch>=1.6.0',
         'torchvision',


nreimers:
Hi,
thanks for the PR.

Sadly I cannot reproduce the error.

When I run (in a clean python 3.9 env):
```
pip install sentence-transformers
```

transformers==4.15.0 and tokenizers==0.10.3 is installed and everything works fine.

Independent of that, such a dependency restriction should be handled by `transformers`.

Which steps did you use to get this error?

rom1504:
Hi,

This can be reproduced by using any pip version before 20.3 (for example `pip install pip==20.1.1`) which introduced dependency resolution (see https://discuss.python.org/t/announcement-pip-20-3-release/5948)
sadly before that version, the dependency resolution mecanism was pretty bad
It will first install all the direct dependencies before the transitive dependencies, and since the dependency resolution was "first installed version wins", tokenizer 0.11 is installed before since you don't restrict the version here, then transformers tries to say "no I want an earlier version" but it's too late with the old dependency resolution mecanism
hence the 
> ERROR: transformers 4.15.0 has requirement tokenizers<0.11,>=0.10.1, but you'll have tokenizers 0.11.2 which is incompatible.

Starting from pip 20.3 this stops being an issue because they introduced proper dependency resolution which takes the intersection of the ranges.

I'm in an environment which prevents me from updating pip to a newer version (for now), and I bet some other people are in the same case.
So I think it could be a nice improvement to follow the same tokenizer restriction as transformers in this sentence transformers package.

nreimers:
Thanks for the explanation. I'm just afraid what happens e.g. with transformers 4.16, which will be compatible with tokenizers 0.11, always changing the setup.py does not appear to be the optimal solution.

Do you know if there is another way for older pips to check dependencies from underlying packages? Maybe removing tokenizers all-together from sentence-transformers dependencies could solve it?

rom1504:
> Maybe removing tokenizers all-together from sentence-transformers dependencies could solve it?

Yes that would work as well

nreimers:
Great, will give it a proper test with different pip versions and then update the dependencies.

Thanks for reporting this issue. pip sadly becomes quite complex with the different dependency resolutions in the different versions

nreimers:
Removed this requirement from the master branch. Thanks for reporting

rom1504:
Great !

PR #1385, opened 2022-01-18
Draft: Single layer classifier evaluator

### LINKED ISSUE #1380: Feature Request: SingleLayerClassificationEvaluator.
Would be great to have an evaluator which judges how easy it is to classify a sentence embedding.

This could be done by an evaluator that trains & tests a classification head for each evaluation.

I think having this evaluator would help test downstream task performance.

is there already a way to do this? I would be happy to contribute this evaluator to the repo.

Fraser-Greenlee:
Resolves: https://github.com/UKPLab/sentence-transformers/issues/1380

Uses an sklearn single layer perceptron trained on a set of sentence embeddings and class labels.

The classifier accuracy indicates how good the semantic encoding is.

This could be used to evaluate a semantic encoders performance on a range of classification task while it trains.

### DIFF
--- a/tests/test_evaluator.py
+++ b/tests/test_evaluator.py
@@ -2,9 +2,9 @@
 Tests the correct computation of evaluation scores from BinaryClassificationEvaluator
 """
 from sentence_transformers import SentenceTransformer, evaluation, util, losses, LoggingHandler
-import logging
 import unittest
 from sklearn.metrics import f1_score, accuracy_score
+import pandas as pd
 import numpy as np
 import gzip
 import csv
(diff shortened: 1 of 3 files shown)


tests/test_evaluator.py
```diff
@@ -59,6 +59,16 @@ def test_LabelAccuracyEvaluator(self):
         acc = evaluator(model)
         assert acc > 0.2
 
+    def test_SingleLayerClassificationEvaluator(self):
+        """Tests that the SingleLayerClassificationEvaluator can loads and runs with reasonable performance."""
+        model = SentenceTransformer('paraphrase-distilroberta-base-v1')
+        df = pd.read_csv('datasets/ag_news.csv')[:1_000]
```
Fraser-Greenlee, line 65: Taken from https://huggingface.co/datasets/ag_news

Jacobluke-:
Hi! I'm also very interested in creating more evaluator for testing downstream tasks. I tried creating a similar evaluator like yours by using an sklearn single layer perceptron for regression. The score of sklearn [MLPRegressor](https://scikit-learn.org/stable/modules/generated/sklearn.neural_network.MLPRegressor.html#sklearn.neural_network.MLPRegressor) is R2 score, which might not be useful because it sometimes gets a negative result. Do you have a better idea?

tomaarsen:
Closing as this is stale, and for most evaluation needs MTEB's classification tasks are now the canonical path.

- Tom Aarsen

PR #1406, opened 2022-02-02
Adding auth_token to SentenceTransformers and args to CrossEncoders to allow the use of private models

zoltan-fedor:
To use private models from the HuggingFace HUB we need to provide your HuggingFace auth token to download models.

This PR adds the option to provide your private token to the SentenceTransformer and CrossEncoder class and so allows it to be used with private models from the HUB.


nreimers:
Thanks so much for adding it :)

PR #1421, opened 2022-02-13
Adding the possibility to pass arguments to the AutoConfig() in the Cross-Encoder

zoltan-fedor:
This is to allow the specification of additional arguments for the `AutoConfig()`.

Apologies, I probably should have included this in https://github.com/UKPLab/sentence-transformers/pull/1406. 
As such, this could be considered as a bug-fix or a new feature. Depending on one's use-case.

If one wants to pull a private Cross-Encoder model with its config and tokenizer from the Hub, then this is bug-fix for that use-case, as that won't work without the `use_auth_token` being supplied to the `AutoConfig()` too.
If one wants to just add some random arguments to the `AutoConfig()`, then this is a new feature to allow that.

### DIFF
--- a/sentence_transformers/cross_encoder/CrossEncoder.py
+++ b/sentence_transformers/cross_encoder/CrossEncoder.py
@@ -19,7 +19,7 @@ logger = logging.getLogger(__name__)
 
 class CrossEncoder():
     def __init__(self, model_name:str, num_labels:int = None, max_length:int = None, device:str = None, tokenizer_args:Dict = {},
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

The CrossEncoder constructor on main now accepts `config_kwargs`, `model_kwargs`, and `processor_kwargs` which thread through to `AutoConfig.from_pretrained` (and friends). That covers the `use_auth_token` use case you described. Closing as resolved.

- Tom Aarsen

Issue #1425, opened 2022-02-16
[Bug] Non-terminating behavior for text-summarization example


dennlinger:
I've encountered a bug with the summarization example, where the underlying Lexrank function would not converge, and iterate forever.

The root cause of this behavior seems to be a problem of the assumption of Lexrank; any (similarity) matrix supplied must follow the simple principle of representing a "stochastic, irreducible and non-periodic [Markov chain] matrix." (see [the paper](https://arxiv.org/pdf/1109.2128.pdf), Algorithm 2).

While this holds true for similarity matrixes generated with graph degrees (as in the original paper), the "transition probabilities" in the case of `sentence-transformers` correspond to the cosine similarity -- which can be negative.
This negative value violates the reducibility and periodicity criteria, which can cause the iterative power method to diverge (which is used to compute the dominant eigenvaluem, eventually forming the centrality distribution).

I'm proposing a simple fix by normalizing slightly differently soon!
Best,
Dennis

PR #1426, opened 2022-02-16
Fix matrix normalization for summarization example (addresses #1425)

dennlinger:
Hi,
as described in #1425 , this is a quick fix to catch non-terminating behavior of Lexrank's eigenvalue computation.

To summarize the changes:
* Removed unused function which was taken from upstream (`graph_nodes_clusters()`).
* Introduced slightly different normalization behavior in the case of negative similarity scores being present (now uses `scipy.special.softmax` instead of division by row sum). One caveat is that this will generate different distributions than the standard method; however, given the continuous notion of similarities from `sentence-transformers`, I think that it is still an appropriate solution.
* To catch similar issues, the `while True:` loop in `_power_method)()` has been replaced with a `max_iter` value (default set to 10000), which will throw an appropriate error in case of non-terminating behavior.

### DIFF
--- a/examples/applications/text-summarization/LexRank.py
+++ b/examples/applications/text-summarization/LexRank.py
@@ -5,6 +5,8 @@ Source: https://github.com/crabcamp/lexrank/tree/dev
 
 import numpy as np
 from scipy.sparse.csgraph import connected_components
+from scipy.special import softmax
+
 
 def degree_centrality_scores(
     similarity_matrix,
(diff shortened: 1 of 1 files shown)


dennlinger:
To touch again on the issue of using `softmax` as a normalizing function: Instead, it would also be possible to scale the matrix accordingly (e.g., by adding +1, the range of `cos_sim` would be non-negative as well).

Given that we're technically not dealing with log probabilities, this would be a slightly more truthful approach to the original value distribution, but I'd love to hear others' thoughts on this.

PR #1434, opened 2022-02-21
Enable encoding with mixed embeddings

jglaser:
This change set generalizes the encoding of mixed batches, i.e. where different sentences in a batch may be in different languages and are encoded by different models.

Usage example for two languages:

```
from sentence_transformers.models import Transformer, Pooling, Asym
transformer_english = Transformer('bert-base-uncased')
transformer_german = Transformer('bert-base-german-cased')

combined = Asym({
    'english': [transformer_english, Pooling(transformer_english.get_word_embedding_dimension(),pooling_mode='cls')],
    'german': [transformer_german, Pooling(transformer_german.get_word_embedding_dimension(),pooling_mode='cls')],
})
transformer = SentenceTransformer(modules=[combined])
```

When there are multiple encodings per batch element, the sentence embeddings are concatenated.

```
>>> transformer_english.get_word_embedding_dimension() 
768
```

```
>>> transformer_german.get_word_embedding_dimension()
768
```

and 

```
>>> transformer.encode([{'english': ['Howdy'], 'german': ['Halloechen']}]).shape
(1, 1, 1536)
```

Of course, in the above example the combined model has to be fine-tuned first before reasonable results can be expected. I did not test fine-tuning of these models through `sentence-transformers`, my use case is that I have a protein and a ligand representation and I would like to get the combined embeddings of both for affinity prediction in an inference situation.

The combined embeddings can be passed through `Dense`, e.g., as an example for a classification head.

### DIFF
--- a/sentence_transformers/models/Asym.py
+++ b/sentence_transformers/models/Asym.py
@@ -72,7 +72,7 @@ class Asym(nn.Sequential):
         return batch
 
     def get_sentence_embedding_dimension(self) -> int:
-        raise NotImplementedError()
+        return {m: m.get_sentence_embedding_dimension() for m in self.sub_modules}
 
     def save(self, output_path):
         model_lookup = {}
--- a/sentence_transformers/models/Asym.py
+++ b/sentence_transformers/models/Asym.py
@@ -1,21 +1,25 @@
 from torch import Tensor
 from torch import nn
+import torch
 from typing import List, Dict
 import os
 import json
 from ..util import import_from_string
 from collections import OrderedDict
 from typing import List, Dict, Optional, Union, Tuple
+from ..util import batch_collator
 
 class Asym(nn.Sequential):
-    def __init__(self, sub_modules: Dict[str, List[nn.Module]], allow_empty_key: bool = True):
+    def __init__(self, sub_modules: Dict[str, List[nn.Module]],
+        allow_empty_key: bool = True):
         """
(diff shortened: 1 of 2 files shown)


jglaser:
For the real application, see here https://huggingface.co/jglaser/protein-ligand-mlp-1

nreimers:
Hi,
thanks for the PR.

Can you explain a bit more why you want to concatenate the different embeddings instead of having them individually?

jglaser:
> Hi, thanks for the PR.

You're welcome :)

> Can you explain a bit more why you want to concatenate the different embeddings instead of having them individually?

I have an existing model with custom inference code that takes the concatenated sequence output (first tokens) from two transformers and feeds it into an MLP. I would like to be able to harness the modularity of sentence-transformers to load/store the combined model conveniently, and offer it in the huggingface inference API, e.g..

Now, regarding the above implementation choice. One could, alternatively, keep them individually but then one would have to generalize all follow-on models, such as `Dense` that operate on `sentence_embedding`s, and make them accept `dict`s (of `dict`s). I don't think that would be difficult to implement.  In addition one would have to implement a `Concatenate` layer in `sentence-transformers`.

Do you have a preference for either approach?

Note that the concatenation inside `Asym` doesn't change the current behavior for input `dict`s that have only one key, since a concat of a list with one `sentence_embedding` is the embedding itself.

jglaser:
I don't have the time to maintain this branch

PR #1468, opened 2022-03-16
Changed incorrect mention of query to corpus

kunal-bhadra:
Changed the mention of chunk sizes in doc. By the usage, corpus_chunk_size should have been mentioned instead of query_chunk_size


nreimers:
Thanks for fixing this :)

PR #1484, opened 2022-03-28
Fix: typo (manhatten -> manhattan)

yk:
Changes the spelling of the Manhattan distance in various places from "manhatten" to "manhattan".


nreimers:
Thanks for this fix :)

Issue #1486, opened 2022-03-30
Loading saved SentenceTransformer model + Dense layer raises Error


ArvinZhuang:
To replicate the error:

```
from sentence_transformers import SentenceTransformer, models

word_embedding_model = SentenceTransformer('bert-base-uncased', cache_folder='cache')
Dense = models.Dense(768, 128)
model = SentenceTransformer(modules=[word_embedding_model, Dense], cache_folder='cache')
print(model)

"""
do training and others
"""

model.save('saved_bert')  # save model
loaded_model = SentenceTransformer('saved_bert')  # load saved model
```

The above code raises: 

```
Traceback (most recent call last):
  File "search.py", line 28, in <module>
    embedder = SentenceTransformer(model_name, cache_folder='cache')
  File "xxx/lib/python3.7/site-packages/sentence_transformers/SentenceTransformer.py", line 94, in __init__
    modules = self._load_sbert_model(model_path)
  File "xxx/lib/python3.7/site-packages/sentence_transformers/SentenceTransformer.py", line 831, in _load_sbert_model
    module = module_class.load(os.path.join(model_path, module_config['path']))
AttributeError: type object 'SentenceTransformer' has no attribute 'load'
```

Im using `sentence-transformers==2.2.0`

Reason: there is no `def load(input_path)` function in current SentenceTransformer class. The pull request provides a simple fix for this error.

nreimers:
I added this function.

However, it is rather unusual to add a dense layer this way. The recommended way is described here:
https://www.sbert.net/docs/training/overview.html#creating-networks-from-scratch

PR #1487, opened 2022-03-30
add load function to SentenceTransformer

ArvinZhuang:
to fix issue #1486


nreimers:
I merged it in the most recent commit using your code. Thanks for raising the issue & providing the fix

Issue #1021, opened 2021-06-23
More modular fit() and support for progress and logging callbacks


chiragjn:
Hello,
Is there a plan to make `.fit` more modular? 

For context, I am integrating the library in an async worker and I want to use python/tensorboard/wandb logging to log metrics, losses, etc every n steps or epochs. The fit function at the moment is not modular enough for me to inherit and override the right points. There is callback fn support but that only works if some evaluator is provided.
For e.g. libraries like fastai, pytorch lightning  provide callbacks for before/after batch/epoch etc

[1] https://docs.fast.ai/callback.progress.html
[2] https://pytorch-lightning.readthedocs.io/en/latest/extensions/callbacks.html

---
If acceptable I can help work on a PR for this :D

nreimers:
Yes, it is planned to integrate a more modular fit (and evaluators) in version 2.1. They should provide options to log to tensorboard / wandb or to your own custom code.

A PR on this would be really nice. I haven't started yet on this topic.

chiragjn:
Just saw 2.0 released a few days ago! :D I forked it and have started refactoring code to add events via callbacks. I'll update when I have something working. In the meantime, I kinda hacked the loss function to be able to do some logging

```python
import torch
from sentence_transformers import losses as st_losses
from timeit import default_timer as timer

class LoggedDenoisingAutoEncoderLoss(st_losses.DenoisingAutoEncoderLoss):
    def __init__(self,
                 model: st.SentenceTransformer,
                 decoder_name_or_path: str = None,
                 tie_encoder_decoder: bool = True,
                 record_every_n_steps: int = 1):
        super().__init__(model=model, decoder_name_or_path=decoder_name_or_path, tie_encoder_decoder=tie_encoder_decoder)
        self._record_every_n_steps = max(1, record_every_n_steps)
        self._step_counter = 0
        self._mtime = 0
        self._running_loss = torch.tensor(0.0)

    def _pre_forward(self):
        if self.encoder.training:
            if self._step_counter == 0 or self._step_counter % self._record_every_n_steps == 0:
                self._mtime = timer()
                self._running_loss = torch.tensor(0.0)

            self._step_counter += 1

    def _post_forward(self, loss: torch.Tensor):
        if self.encoder.training:
            self._running_loss += loss.item()

            if self._step_counter % self._record_every_n_steps == 0:
                metrics = {
[...]

Issue #510, opened 2020-10-20
can the loss or some information during training be shown?


823858275:
i use model.fit() to train the model, but it doesn't show the detail of training, such as loss. I don't find some parameter to show those information.can the loss or some information during training be shown?

nreimers:
You can pass an evaluator to the fit method, as shown here:
https://github.com/UKPLab/sentence-transformers/blob/master/examples/training/nli/training_nli.py

The evaluator is called every `evaluation_steps` and computes whatever scores you like. Usually I use it to compute scores (like accuracy, correlation) on some held-out dev set.

iamlockelightning:
Hi, @nreimers . Can it just return the training loss values during `fit`? (On progress bar, or logging to file/std, or Tensorboard support, ...)

I don't have a held-out dev set. Just want to monitor the training process.

Rj7:
@nreimers This would really useful for Unsupervised learning.

Issue #336, opened 2020-08-03
Can it show the loss?


aRookieMan:
My model is getting worse than pre-trained model on my own dataset, I want to see the loss to do some investigation. How can I can the loss? Will you introduce something like tensorboard to see the loss or the embedding?

pashok3d:
Without touching `fit` method, you can implement custom evaluator, where you can place your loss calculation (`with torch.no_grad()`) and return it instead of accuracy. Than you can set `evaluation_steps` in `fit` method according to how frequently you want to get loss. Additionally, you can return `1-loss` to save the model with the lowest loss. Another way is to accumulate loss during each epoch from `fit` method and then return average for example.

iamlockelightning:
> My model is getting worse than pre-trained model on my own dataset, I want to see the loss to do some investigation. How can I can the loss? Will you introduce something like tensorboard to see the loss or the embedding?

+1

iamlockelightning:
Hi, @nreimers , thanks for this awesome sentence-transformers. It's so elegant!!

We want to see the loss to do some investigation. How can we print the loss during `fit` (on the progress bar, logging to file/std, ...), instead of evaluating on a held-out dev set (because sometimes I don't have a dev set, just training the model for good emb for another task)?  is there an alternative method to achieve it now?

Besides, will you introduce something like tensorboard to see the loss and other metrics?

hchings:
@pashok3d Can you give an example of getting training loss in the customer evaluator without touching the fit method? Thanks.

duongkstn:
any update ?

nreimers:
Sadly no update yet

Mogady:
I have created a custom evaluator for tracking the evaluation loss and integrating tensorboard. so far it works for me would be nice to double check from anyone while trying it.

```python
import logging
import os
import csv
import numpy as np
from typing import List, Union
import math
from tqdm.autonotebook import trange

import torch
from torch import nn
from torch.utils.tensorboard import SummaryWriter

from sentence_transformers import InputExample
from sentence_transformers.evaluation import SentenceEvaluator


logger = logging.getLogger(__name__)

class LossEvaluator(SentenceEvaluator):

    def __init__(self, loader, loss_model: nn.Module = None, name: str = '', show_progress_bar: bool = False, write_csv: bool = True):
        
        """
        Evaluate a model based on the loss function.
        The returned score is loss value.
        The results are written in a CSV and Tensorboard logs.
[...]

PR #1532, opened 2022-04-25
Adding tensorboard loss tracking for training and evaluation

Mogady:
I saw multiple discussions on adding loss tracking and tensor-board integration, however no pull requests have been made in that direction. [1021](https://github.com/UKPLab/sentence-transformers/issues/1021#issue-927902758), [510](https://github.com/UKPLab/sentence-transformers/issues/510#issue-725261879), [336](https://github.com/UKPLab/sentence-transformers/issues/336#issue-671935207)   and recently I needed that also.

I added a loss tracking module for both training and evaluation.

For evaluation:
since the code allow for custom evaluators, a LossEvaluator class is added which allow for calculating the loss and writing to TFboard logs beside the CSV file.

For training:
- I had to change in the` .fit()` signature but not much, just adding a parameter (tensorboard: bool) for controlling it.
- I used the evaluation_steps parameter as a flag, since it is the best option for now to log the loss with the evaluation steps. However a better implementation will allow for using different parameter.

I understand this is not general enough, and i tried to not change the code signature that much, but hopefully it can serve as a reference or a basis for a better integration in the next versions.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -14,6 +14,7 @@ import torch
 from torch import nn, Tensor, device
 from torch.optim import Optimizer
 from torch.utils.data import DataLoader
+from torch.utils.tensorboard import SummaryWriter
 import torch.multiprocessing as mp
 from tqdm.autonotebook import trange
 import math
@@ -576,6 +577,7 @@ class SentenceTransformer(nn.Sequential):
             save_best_model: bool = True,
             max_grad_norm: float = 1,
             use_amp: bool = False,
+            tensorboard: bool = True,
             callback: Callable[[float, int, int], None] = None,
             show_progress_bar: bool = True,
             checkpoint_path: str = None,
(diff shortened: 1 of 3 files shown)


mi1ebsco:
Seems there are lots of whitespace differences that makes this PR difficult to review.  Perhaps remove the whitespace changes you made and this might expedite the review process?

Mogady:
> Seems there are lots of whitespace differences that makes this PR difficult to review. Perhaps remove the whitespace changes you made and this might expedite the review process?

Yes, sorry about that, it should be more clear now

Mogady:
> Thanks for this PR.
> 
> There is a PR #1606 that adds a generic way to track loss that is also compatible with e.g. W&B. My plan is to add this generic framework first and then have classes / callbacks that can be used to track with tensorboard / W&B etc.

Yes, W&B would be better definitely and more generic, however, the reason why I didn't start with it, is its enterprise license.
since it is not free and not open source, I recommend giving us access to the loss and the grads values besides adding W&B (similar to HuggingFace), so that if I can't use their dashboard then I can design any custom one.

Mogady:
> Thanks for this PR.
> 
> There is a PR #1606 that adds a generic way to track loss that is also compatible with e.g. W&B. My plan is to add this generic framework first and then have classes / callbacks that can be used to track with tensorboard / W&B etc.

Great, W&B is definitely better and easier, I couldn't use it because of the license, but if you are giving us access to the values to design any custom dashboard then it would be perfect.

PR #1551, opened 2022-05-11
dont move device in collate fn

dctelus:
This allows setting num_workers for the pytorch Dataloader. Needed this to support mutltiprocess data loading.


nreimers:
Thanks for adding this PR :)

PR #1554, opened 2022-05-12
"Multi-Lingual Models" languages used

graemenail:
Hi, 

I found that `pt` is duplicated in list of languages for the pretrained models.
This is also duplicated in the paper referred to in the text.

Might it also be worth mentioning the status of `en` here? 

Thanks


nreimers:
Thanks for the PR. I will add en to the list

Issue #1461, opened 2022-03-11
Model not automatically loaded from cache


DavidHuebner:
First, let me thank you for maintaining this great package!

I think that the model-caching logic in `SentenceTransformer.py` could be improved. When I specify a `model` and a `cache_dir` like so `SentenceTransformer("distiluse-base-multilingual-cased-v1", cache_folder="~/.models/")` , my hope (and expectation) was the following behaviour:

1. When executed for the first time, it should download the model into the `cache_folder` (this is ok!).
2. When executed for a second time, it should re-load the existing model from that cache folder (does not work!). Even worse, it will download the model every time.

To actually make the model reload from cache, one would need to specify the `model_name_or_path` like so: `SentenceTransformer(model_name_or_path="~/.models/sentence-transformers_distiluse-base-multilingual-cased-v1")`. This leaves me with two different calls (first call, later re-loads) to SentenceTransformer with some checking in between. 

The fix to this is rather simple. The loading code should check if the model was already downloaded to the cache_dir. The call to `snapshot_download` in https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L86 should only be executed if there is no existing model at `model_path` yet.

sadnoodles:
Sorry for incomplete PR.

https://github.com/UKPLab/sentence-transformers/pull/1923 is also not a fix.

```python
            if os.path.exists(model_name_or_path):
                #Load from path
                model_path = model_name_or_path
            else:
                ......
                if not os.path.exists(model_path):
                    # Download from hub with caching
                    snapshot_download(model_name_or_path,
```
is equal to

```python
            if os.path.exists(model_name_or_path):
                #Load from path
                model_path = model_name_or_path
            else:
                ......
                snapshot_download(model_name_or_path,
```

Which does not check anything.

I’m think check it's integrality.

As you can see from the following code, some models do not have `modules.json`

```python
            if os.path.exists(os.path.join(model_path, 'modules.json')):    #Load as SentenceTransformer model
                modules = self._load_sbert_model(model_path)
            else:   #Load with AutoModel
                modules = self._load_auto_model(model_path)
```

When the download is complete, add a file indicating that the download is complete. Verify integrity with this file.
if not completely, we go to the snapshot_download.

sadnoodles:
Sorry for incomplete PR.

https://github.com/UKPLab/sentence-transformers/pull/1923 is also not a fix. If download is not complete, this will go wrong.

[...]

PR #1565, opened 2022-05-24
Download from online only if not cached

sadnoodles:
Fix issue of  #1461 . This issue will cause error when cached & offline.


nreimers:
Thanks for this PR :)

PR #1589, opened 2022-06-12
Update train_bi-encoder_mnrl.py

mscham:
del ce_scores object - releases around 17gb of ram


nreimers:
Thanks for updating the example

PR #1591, opened 2022-06-12
Add data streaming example

NouamaneTazi:
This example demonstrates how to [stream data](https://huggingface.co/docs/datasets/stream) which is helpful in case you don't want to wait for an extremely large dataset to download, or if you want to limit the amount of memory used.


nreimers:
Thanks for adding this example :)

PR #1603, opened 2022-06-22
Added functionality to do community detection in batches

### LINKED ISSUE #1023: Batch cos_sim for community_detection?
I've been experimenting with the `community_detection` method but noticed I quickly get OOM errors if I use too large of embeddings.

Seeing how it uses `cos_sim` to computed all the embedding distances, do you think it would make sense to have the option for batching? I believe you will find other bottlenecks when iterating over the entries, but at least it will complete on larger embeddings.

YashGadhia:
This resolves #1023 
Modified community detection to process the embeddings in batches since if the number of embeddings is large, the CPU/GPU can run out of memory while finding cosine similarity between all pairs at once.

### DIFF
--- a/sentence_transformers/util.py
+++ b/sentence_transformers/util.py
@@ -9,6 +9,8 @@ import torch
 import numpy as np
 import queue
 import logging
+import math
+import gc
 
 from huggingface_hub import HfFolder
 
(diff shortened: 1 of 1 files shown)


nreimers:
Thanks for this PR. I merged the code together with #1450 and committed it. It will be part of the next release.

PR #1614, opened 2022-06-26
Replace broken relative links with sbert.net links

bminaiev:
Currently some docs (e.g. [https://github.com/UKPLab/sentence-transformers/tree/master/examples/training/sts](https://github.com/UKPLab/sentence-transformers/tree/master/examples/training/sts)) have broken links (e.g. [CosineSimilarityLoss](https://github.com/UKPLab/sentence-transformers/blob/master/docs/package_reference/losses.html#cosinesimilarityloss)) or links to not very informative github pages (e.g. [models overview](https://github.com/UKPLab/sentence-transformers/blob/master/docs/package_reference/models.md)).

Replaced them with corresponding links to sbert.net as in most other docs.


tomaarsen:
Hello!

To the best of my knowledge, each of these links work correctly in the built documentation. I don't think we should update them, as I prefer having relative links over these absolute links to sbert.net. So, I will be closing this PR.

- Tom Aarsen

PR #1629, opened 2022-07-10
Refactor save_to_hub, and add option to push training checkpoints to Hub

NimaBoscarino:
This PR contributes two things:

1. The `save_to_hub` method has been refactored and simplified to use [one of the new utility functions](https://huggingface.co/docs/huggingface_hub/how-to-upstream#managing-files-in-a-repo-without-git-with-the-createcommit-api) from the Hugging Face Hub client library, which uses the `create_commit` API.
2. A feature has been added to allow `.fit()` to push checkpoints to the Hub.

Pushing the checkpoints halts training (and there are no progress bars for the upload), which might not be ideal UX. Would it be better if the upload happened in the background while training continued?

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -5,6 +5,7 @@ import shutil
 import stat
 from collections import OrderedDict
 from typing import List, Dict, Tuple, Iterable, Type, Union, Callable, Optional
+from dataclasses import asdict, dataclass, field
 import requests
 import numpy as np
 from numpy import ndarray
@@ -31,24 +32,27 @@ from . import __version__
 logger = logging.getLogger(__name__)
 
 
+@dataclass
 class HubTrainingArguments:
-    def __init__(self, repo_name: str,
-                 organization: Optional[str] = None,
-                 private: Optional[bool] = None,
-                 local_model_path: Optional[str] = None,
-                 replace_model_card: Optional[bool] = True,
-                 train_datasets: Optional[List[str]] = None,
-                 local_repo_path: Optional[str] = None,
-                 ):
-        self.parameters = {
-            "repo_name": repo_name,
-            "organization": organization,
-            "private": private,
-            "local_model_path": local_model_path,
-            "replace_model_card": replace_model_card,
-            "train_datasets": train_datasets,
-            "local_repo_path": local_repo_path,
-            "exist_ok": True
+    repo_name: str = field()
+    organization: Optional[str] = field(default=None)
+    private: Optional[bool] = field(default=None, metadata={"help": "Specifies the model repo's visibility"})
+    local_model_path: Optional[str] = field(default=None, metadata={"help": "Local path to a model checkpoint"})
+    replace_model_card: Optional[bool] = field(default=None, metadata={"help": "Controls whether the existing model card is replaced with a new one"})
+    train_datasets: Optional[List[str]] = field(default=None, metadata={"help": "List of datasets used for training, included in the model card"})
+    local_repo_path: Optional[str] = field(default=None, metadata={"help": "Local path to use to initialize a repo"})
+
+    @property
+    def parameters(self):
+        return {
+            "repo_name": self.repo_name,
+            "organization": self.organization,
+            "private": self.private,
+            "local_model_path": self.local_model_path,
+            "replace_model_card": self.replace_model_card,
+            "train_datasets": self.train_datasets,
+            "local_repo_path": self.local_repo_path,
+            "exist_ok": True,
         }
 
 
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -36,7 +36,7 @@ class HubParameters:
                  organization: Optional[str] = None,
                  private: Optional[bool] = None,
                  local_model_path: Optional[str] = None,
-                 replace_model_card: bool = False,
+                 replace_model_card: Optional[bool] = True,
                  train_datasets: Optional[List[str]] = None,
                  local_repo_path: Optional[str] = None,
                  ):
@@ -131,6 +131,8 @@ class SentenceTransformer(nn.Sequential):
 
         self._target_device = torch.device(device)
 
+
+
     def encode(self, sentences: Union[str, List[str]],
                batch_size: int = 32,
                show_progress_bar: bool = None,
@@ -224,6 +226,8 @@ class SentenceTransformer(nn.Sequential):
 
         return all_embeddings
 
+
+
     def start_multi_process_pool(self, target_devices: List[str] = None):
         """
         Starts multi process to process the encoding with several, independent processes.
@@ -253,6 +257,7 @@ class SentenceTransformer(nn.Sequential):
             processes.append(p)
 
         return {'input': input_queue, 'output': output_queue, 'processes': processes}
+
 
     @staticmethod
     def stop_multi_process_pool(pool):
@@ -320,6 +325,8 @@ class SentenceTransformer(nn.Sequential):
                 results_queue.put([id, embeddings])
             except queue.Empty:
                 break
+
+
 
     def get_max_seq_length(self):
         """
@@ -428,6 +435,7 @@ class SentenceTransformer(nn.Sequential):
             if train_datasets is not None:
                 datasets_str = "datasets:\n"+"\n".join(["- " + d for d in train_datasets])
             model_card = model_card.replace("{DATASETS}", datasets_str)
+
 
             # Add dim info
             self._model_card_vars["{NUM_DIMENSIONS}"] = self.get_sentence_embedding_dimension()
@@ -570,6 +578,7 @@ class SentenceTransformer(nn.Sequential):
 
         return sentence_features, labels
 
+
     def _text_length(self, text: Union[List[int], List[List[int]]]):
         """
         Help function to get the length for the input text. Text can be either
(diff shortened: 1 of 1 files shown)


NimaBoscarino:
cc: @omarespejel @osanseviero

sentence_transformers/SentenceTransformer.py
```diff
@@ -2,14 +2,12 @@
...
 import shutil
-import stat
 from collections import OrderedDict
 from typing import List, Dict, Tuple, Iterable, Type, Union, Callable, Optional
-import requests
 import numpy as np
 from numpy import ndarray
 import transformers
-from huggingface_hub import HfApi, HfFolder, Repository, hf_hub_url, cached_download
+from huggingface_hub import HfApi, HfFolder, upload_folder
```
osanseviero, line 10: Unfortunately we can't use `create_commit`. It was just added in 0.8.1, but ST uses any version from https://github.com/UKPLab/sentence-transformers/blob/master/setup.py#L31, specially to support Python 3.6 which we don't support in hhub in latest versions (https://github.com/UKPLab/sentence-transformers/commit/39ab855919652846cb5e15c260dc9f07ecf52951)

sentence_transformers/SentenceTransformer.py
```diff
@@ -618,8 +588,12 @@ def fit(self,
         :param checkpoint_path: Folder to save checkpoints during training
         :param checkpoint_save_steps: Will save a checkpoint after so many steps
         :param checkpoint_save_total_limit: Total number of checkpoints to store
+        :param push_checkpoints_to_hub: If true, each checkpoint will be pushed to the Hugging Face Hub
```
osanseviero, line 591:
```suggestion
        :param push_checkpoints_to_hub: If True, each checkpoint will be pushed to the Hugging Face Hub
```

sentence_transformers/SentenceTransformer.py
```diff
@@ -797,6 +768,9 @@ def _save_checkpoint(self, checkpoint_path, checkpoint_save_total_limit, step):
                 old_checkpoints = sorted(old_checkpoints, key=lambda x: x['step'])
                 shutil.rmtree(old_checkpoints[0]['path'])
 
+        if save_to_hub:
+            self.save_to_hub(**self.hub_parameters, exist_ok=True, commit_message="Push new training checkpoint.")
```
osanseviero, line 772: What would go in `hub_parameters` and what would go in normal params? E.g. why is `exist_ok` separate?

osanseviero (commented):
Very nice! Do you have an example repo created with this?

sentence_transformers/SentenceTransformer.py
NimaBoscarino: Ah I should have documented it better. `hub_parameters` will be passed entirely to `save_to_hub`, so users should only put in parameters that make sense for `save_to_hub`. However, I have to override `exist_ok` so the call to `create_repo` in `save_to_hub` won't fail after the first checkpoint.

Otherwise I'm not quite sure what pattern to use to do it cleanly without making structural changes to the code. I could create a `HubParameters` class to hold the parameters, or I could have users pass all of the hub parameters when `.fit()` is called? `fit` already has a ton of parameters, so I didn't want to add more in there.

NimaBoscarino:
@osanseviero Yup! Here's an example repo: https://huggingface.co/NimaBoscarino/STPushToHub-test

sentence_transformers/SentenceTransformer.py
osanseviero: Makes sense to have a `HubParameters`

NimaBoscarino:
@osanseviero I refactored it and now have a HubParameters class to hold the params! Here's a repo where I tested it all out: https://huggingface.co/NimaBoscarino/July25Test/commits/main

If you think this looks good, I can write up some docstrings for it and I can add it to the Hugging Face section of the docs.

sentence_transformers/SentenceTransformer.py
```diff
@@ -30,6 +30,27 @@
 
 logger = logging.getLogger(__name__)
 
+
+class HubParameters:
```
osanseviero, line 34: Consider using a `dataclass` for this :) it will be cleaner and easier to use and to extend in the future. See https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py#L106 as an example

sentence_transformers/SentenceTransformer.py
```diff
@@ -30,6 +30,27 @@
...
+    def __init__(self, repo_name: str,
+                 organization: Optional[str] = None,
+                 private: Optional[bool] = None,
+                 local_model_path: Optional[str] = None,
+                 replace_model_card: bool = False,
```
osanseviero, line 39: Should this be `True` by default? I would also mark it as `Optional`

sentence_transformers/SentenceTransformer.py
```diff
@@ -106,8 +131,6 @@ def __init__(self, model_name_or_path: Optional[str] = None,
 
         self._target_device = torch.device(device)
 
-
```
osanseviero, line 109: Please remove changes unrelated to the PR (e.g. removing new lines)

sentence_transformers/SentenceTransformer.py
```diff
@@ -474,53 +493,51 @@ def save_to_hub(self,
...
+        repo_dir = local_repo_path
 
-            # Remove .git folder. On Windows, the .git folder might be read-only and cannot be deleted
-            # Hence, try to set write permissions on error
+        if local_repo_path is None:
+            repo_dir = tempfile.mkdtemp()
+
+        # First create the repo (and clone its content if it's nonempty).
+        logger.info("Create repository and clone it if it exists")
+        repo = Repository(repo_dir, clone_from=repo_url)
```
osanseviero, line 503: Not really related to your PR, but could we use `create_repo` with `exist_ok` before `Repository`? That way we know that the repository exists on the Hub. The behavior of `Repository` creating the repo is a bit strange and will be deprecated in the future.

sentence_transformers/SentenceTransformer.py
osanseviero, line 34: Maybe this should be renamed to `HubTrainingArguments`? My concern if not is that if you use `save_to_hub` with a `SentenceTransformers` model that was initialized with params for a different repo/org, then the behaviour becomes confusing. And if we do that maybe `HubTrainingArguments` should just be passed in `fit`. WDYT?

sentence_transformers/SentenceTransformer.py
```diff
@@ -673,7 +698,6 @@ def fit(self,
             optimizers.append(optimizer)
             schedulers.append(scheduler_obj)
 
-
```
osanseviero, line 676: Nit: please remove these non changes

osanseviero (commented):
Thanks a lot! This is looking quite good, I left some final comments

sentence_transformers/SentenceTransformer.py
NimaBoscarino: Cool, TIL! Will do 👍

sentence_transformers/SentenceTransformer.py
NimaBoscarino: Yup, it's already being called further up! https://github.com/UKPLab/sentence-transformers/blob/a83ce1629a1e6da8706ce8f6c5ce5c51ba06a947/sentence_transformers/SentenceTransformer.py#L486

I think the comment here is just confusing, so I'll change it.

sentence_transformers/SentenceTransformer.py
NimaBoscarino: Sure, that sounds good to me! I'll remove it from the SentenceTransformers constructor.

sentence_transformers/SentenceTransformer.py
```diff
@@ -449,6 +476,7 @@ def save_to_hub(self,
         :param exist_ok: If true, saving to an existing repository is OK. If false, saving only to a new repository is possible
         :param replace_model_card: If true, replace an existing model card in the hub with the automatically created model card
         :param train_datasets: Datasets used to train the model. If set, the datasets will be added to the model card in the Hub.
+        :param local_repo_path: Local path where the model repo will be stored. If not provided, a temporary directory will be created.
```
osanseviero, line 479: I'm not convinced by the behaviour of using a temporary directory by default. This would mean the repo is cloned every single save, which would make it very slow. Should it use something like "hub" by default maybe?

osanseviero (approved):
This looks good to me! 
* Feel free to close discussions a resolved if they are resolved :) 
* I left a comment about `local_repo_path` creating temporary directories

APart from that, this looks in good state. Pinging @nreimers :hugs: to see what they think

sentence_transformers/SentenceTransformer.py
NimaBoscarino: I'm okay with either way! I think we'd decided to keep the temp directory by default from this convo: https://huggingface.slack.com/archives/C03HVLC16D7/p1657743607849189?thread_ts=1657694774.984229&cid=C03HVLC16D7

osanseviero: Hmm good point :thinking: after thinking more about it I think having "hub" by default might make more sense to avoid too many redownloads

tomaarsen:
Hello!

This PR has been superseded by #2380. Also, Repository is now outdated. I do want to thank you for this time, and I'm sorry that it was never merged - I recognize that you (all) put time and effort into this.

- Tom Aarsen

PR #1636, opened 2022-07-12
Added SENTENCE_TRANSFORMERS_HOME to docs

boing102:
Added a reference to the `SENTENCE_TRANSFORMERS_HOME` env var to the API's docstring. 

I was a bit confused because `sentence-transformers` is not using the HF cache dir env var, but it defines its own. But the specifc env var is not mentioned anywhere in the docs page.


sentence_transformers/SentenceTransformer.py
```diff
@@ -10,6 +10,7 @@
 from numpy import ndarray
 import transformers
 from huggingface_hub import HfApi, HfFolder, Repository, hf_hub_url, cached_download
+from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
```
nreimers, line 13: I think this line is not needed. Could you remove it from the PR so that it can be merged?

nreimers:
Thanks for extending the docs. Added one comment, it could then be merged

boing102:
> Thanks for extending the docs. Added one comment, it could then be merged

@nreimers Good catch, fixed now.

nreimers:
Thanks for adding this :)

Issue #1643, opened 2022-07-20
Use config_sentence_transformers.json in the util functions


NimaBoscarino:
SentenceTransformer models can have an associated `config_sentence_transformers.json` file ([see this one for example](https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2/blob/main/config_sentence_transformers.json)) which contains information about the versions that were used when saving the model. It would be nice to extend the use of this file.

Since the ideal scoring function (dot vs. cos_sim) is known at training time, that could be written into the `config_sentence_transformers.json` file. Then, some of the util methods (`semantic_search`, `paraphrase_mining`, etc.) could read directly from that file to choose the appropriate scoring function.

PR #1649, opened 2022-07-26
Fetch score_function for semantic_search from model config if available

NimaBoscarino:
For issue: #1643 
This PR loads the score_function from the model config JSON file, only if the model is passed to the `semantic_search` function. I tested this by using a local model with an edited `config_sentence_transformers.json` file.

The other places where I think this could be relevant are:
* InformationRetrievalEvaluator
* SequentialEvaluator
* util.paraphrase_mining

If you think this looks okay, I can make sure this is tested in `test_util`, and then I can also implement it for those other ones that I listed above!

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -106,7 +106,9 @@ class SentenceTransformer(nn.Sequential):
 
         self._target_device = torch.device(device)
 
-
+    @property
+    def score_function(self):
+        return self._model_config.get("score_function", "cos_sim")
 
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

The model now carries `similarity_fn_name` in its config and exposes `model.similarity` / `model.similarity_pairwise` / `util.semantic_search` that pick it up automatically. The original ask ("let semantic_search read the score function from config") is now solved through the model rather than through `util.semantic_search` directly. Closing.

- Tom Aarsen

PR #1653, opened 2022-07-28
Allow specifying `num_cpus` for multiprocess encoding

MikulasZelinka:


tomaarsen:
Hello!

Hmm, I think this is fairly interesting, but users can also just provide `["cpu"] * num_cpus` themselves. After all, the `target_devices = ['cpu'] * 4` only exists as a default in case the user didnt specify devices. 
Although admittedly, starting multi-process encoding with e.g. `['cpu', 'cpu', 'cpu', 'cpu', 'cpu', 'cpu', 'cpu', 'cpu']` isn't exactly very intuitive.

- Tom Aarsen

tomaarsen:
Related to #2383

MikulasZelinka:
@tomaarsen As long as `["cpu"] * num_cpus` is explicitly documented and shown in an example, I think that's fine :)

tomaarsen:
I've included it here:
https://github.com/UKPLab/sentence-transformers/blob/056d9b4a77904d0d8b7ce179d6d82f660aa7eb05/sentence_transformers/SentenceTransformer.py#L317-L325

Feel free to let me know if you think that it might not be clear enough!

- Tom Aarsen

PR #1656, opened 2022-07-31
word embeddings from word2vec fails to load correctly

lambdaofgod:
Loading from word2vec format fails if txt file contains first line with number of words and dimensionality.

For example gensim exports to .txt file with the first line like this:
```
10000 50
```

In the current version loading will set dimensionality to 1 and fail to load vectors.

I fixed this by checking whether first line contains exactly two tokens, and treating the second one as dimensionality.

### DIFF
--- a/sentence_transformers/models/WordEmbeddings.py
+++ b/sentence_transformers/models/WordEmbeddings.py
@@ -106,11 +106,11 @@ class WordEmbeddings(nn.Module):
                 word = split[0]
 
                 if len(split) == 2:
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Thanks for your PR pointing me to this issue! I've resolved this issue by merging #1875 which also had a clean solution. As a result, I will me closing this, and the next version of Sentence Transformers should include word2vec support.

- Tom Aarsen

PR #1684, opened 2022-09-05
Fix TypeError: object is not subscriptable.

lauberto:
If the type of texts[0] is, for example, 'int', indexing the variable text_tuple will result in a TypeError. By changing the order of the if-else statement conditions, and converting to list at the 'else' step, this error could be prevented.

### DIFF
--- a/sentence_transformers/models/Transformer.py
+++ b/sentence_transformers/models/Transformer.py
@@ -86,9 +86,7 @@ class Transformer(nn.Module):
         Tokenizes a text and maps tokens to token-ids
         """
         output = {}
-        if isinstance(texts[0], str):
-            to_tokenize = [texts]
-        elif isinstance(texts[0], dict):
+        if isinstance(texts[0], dict):
             to_tokenize = []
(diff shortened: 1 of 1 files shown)


lauberto:
I had encountered this error:
```
  ...
  File "/env/lib/python3.9/site-packages/sentence_transformers/SentenceTransformer.py", line 319, in tokenize
    return self._first_module().tokenize(texts)
  File "/env/lib/python3.9/site-packages/sentence_transformers/models/Transformer.py", line 106, in tokenize
    batch1.append(text_tuple[0])
TypeError: 'datetime.datetime' object is not subscriptable
```

tomaarsen:
Hello!

I'm trying to figure out the use case where this is strictly necessary. It seems that you didn't feed the tokenizer with strings (or tuples or dictionaries with strings). It doesn't really make sense to me to support tokenizing integers or datetime instances. 

I will be closing this, I'm afraid. But feel free to reply if you indeed believe that this behaviour should be supported in some way.

- Tom Aarsen

PR #1685, opened 2022-09-07
support max_seq_length argument in SentenceTransformer

Jiaxin-Wen:


tomaarsen:
Hello!

This is already covered by passing `processing_kwargs={"model_max_length": 512}` when loading the model, or setting `model.max_seq_length = N` after load, passing `max_seq_length` directly to `Transformer(...)` when building from modules. The first one in particular should match your goals!

- Tom Aarsen

PR #1715, opened 2022-10-05
Fix issue of storing too many docs during IR-eval.: Maintain topk with heaps

kwang2049:
In the current version, the `InformationRetrievalEvaluator` and the `util.semantic_search` accumulates all the docs in each top-k retrieved chunk in the query results:
https://github.com/UKPLab/sentence-transformers/blob/a8cebb235066cacf533d073b8c8250e5e7b04c3d/sentence_transformers/evaluation/InformationRetrievalEvaluator.py#L162-L173

https://github.com/UKPLab/sentence-transformers/blob/a36e6f126cf0e333e1b00eb7cfcef2863f8919ad/sentence_transformers/util.py#L218-L222

In other words, the `queries_result_list` will hold #trunks * top-K docs for each query instead of just top-K. This will result in a very **severe memory burden for a large corpus, e.g. 60GB+ when evaluating on MS MARCO**. One could do some simulation about how large RAM it could use:

```python
import random
import psutil

def entry():
	return {'corpus_id': random.randint(0, 100000), 'score': random.random()}

msmarco_docs = 8800000
corpus_chunk_size = 50000
top_k = 1000
queries = 100
# queries = 7000

print("Available RAM (GB), before:", psutil.virtual_memory().available / 1024 / 1024 / 1024)

query_results = [entry() for _ in range(msmarco_docs // corpus_chunk_size * top_k * queries)]

print("Available RAM (GB), after:", psutil.virtual_memory().available / 1024 / 1024 / 1024)

# Available RAM (GB), before: 16.977794647216797
# Available RAM (GB), after: 11.836997985839844
```

This PR fixes the issue by maintaining exactly top-K docs with heaps efficiently during the whole retrieval process. A test has been made to make sure the new code yields the same final results / scores:
https://colab.research.google.com/drive/1ibA6hjfXKsl97L1wA_FlT1HnVFhnRw0L?usp=sharing

### DIFF
--- a/sentence_transformers/evaluation/InformationRetrievalEvaluator.py
+++ b/sentence_transformers/evaluation/InformationRetrievalEvaluator.py
@@ -174,7 +174,7 @@ class InformationRetrievalEvaluator(SentenceEvaluator):
                         if len(queries_result_list[name][query_itr]) < max_k:
                             heapq.heappush(queries_result_list[name][query_itr], (score, corpus_id))  # heaqp tracks the quantity of the first element in the tuple
                         else:
-                            heapq.heapreplace(queries_result_list[name][query_itr], (score, corpus_id))
+                            heapq.heappushpop(queries_result_list[name][query_itr], (score, corpus_id))
 
         for name in queries_result_list:
             for query_itr in range(len(queries_result_list[name])):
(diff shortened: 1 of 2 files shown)


kwang2049:
Due to the same reason mentioned in https://github.com/beir-cellar/beir/pull/117#issuecomment-1272034017, I have updated the PR with `heapq.heappushpop` instead of `heapq.heapreplace`. The colab in the PR desc has also been updated. [0cb2720](https://github.com/UKPLab/sentence-transformers/pull/1715/commits/0cb2720bc8894d415b04a39305e44db8bc01d573)

PR #1717, opened 2022-10-09
Fix Cuda out of memory issue in model.encode by allowing user to transfer to cpu

Quetzalcohuatl:
In issue #487 and issue #522, users were running into OOM issues when batch size is large, because the embeddings aren't offloaded onto cpu.

The PR that fixed this only fixes it `if convert_to_numpy`. That means if you have convert_to_numpy=False, then your problem still exists.

In this PR, I just added an extra flag that allows the embeddings to be offloaded to cpu. This gives the user the flexibility to save the embeddings (for example if they are saving  the SentenceTransformer embeddings to disk or keeping them in RAM for knn, which is often the case) instead of keeping all the embeddings on the gpu.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -114,6 +114,7 @@ class SentenceTransformer(nn.Sequential):
                output_value: str = 'sentence_embedding',
                convert_to_numpy: bool = True,
                convert_to_tensor: bool = False,
+               transfer_to_cpu = False,
                device: str = None,
                normalize_embeddings: bool = False) -> Union[List[Tensor], ndarray, Tensor]:
         """
@@ -125,6 +126,7 @@ class SentenceTransformer(nn.Sequential):
         :param output_value:  Default sentence_embedding, to get sentence embeddings. Can be set to token_embeddings to get wordpiece token embeddings. Set to None, to get all output values
         :param convert_to_numpy: If true, the output is a list of numpy vectors. Else, it is a list of pytorch tensors.
         :param convert_to_tensor: If true, you get one large tensor as return. Overwrites any setting from convert_to_numpy
+        :param transfer_to_cpu: If true, the output_value will be detached and moved to cpu. This helps avoid Cuda Out-of-memory issues.
         :param device: Which torch.device to use for the computation
         :param normalize_embeddings: If set to true, returned vectors will have length 1. In that case, the faster dot-product (util.dot_score) instead of cosine similarity can be used.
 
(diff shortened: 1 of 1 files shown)


paolorechia:
Hey, thanks for opening this, I'm facing this OOM issue. I'll checkout this PR and give it a try.

paolorechia:
Hey, I tried using this PR, however, I ran into the following error:

```
    con.model.fit(
  File "/home/paolo/dev/openimagegenius/gpu-code/gpu-node/lib/python3.10/site-packages/sentence_transformers/SentenceTransformer.py", line 749, in fit
    self._eval_during_training(evaluator, output_path, save_best_model, epoch, -1, callback)
  File "/home/paolo/dev/openimagegenius/gpu-code/gpu-node/lib/python3.10/site-packages/sentence_transformers/SentenceTransformer.py", line 781, in _eval_during_training
    score = evaluator(self, output_path=eval_path, epoch=epoch, steps=steps)
  File "/home/paolo/dev/openimagegenius/gpu-code/gpu-node/lib/python3.10/site-packages/sentence_transformers/evaluation/EmbeddingSimilarityEvaluator.py", line 77, in __call__
    embeddings1 = model.encode(self.sentences1, batch_size=self.batch_size, show_progress_bar=self.show_progress_bar, convert_to_numpy=True)
  File "/home/paolo/dev/openimagegenius/gpu-code/gpu-node/lib/python3.10/site-packages/sentence_transformers/SentenceTransformer.py", line 201, in encode
    all_embeddings = np.asarray([emb.numpy() for emb in all_embeddings])
  File "/home/paolo/dev/openimagegenius/gpu-code/gpu-node/lib/python3.10/site-packages/sentence_transformers/SentenceTransformer.py", line 201, in <listcomp>
    all_embeddings = np.asarray([emb.numpy() for emb in all_embeddings])
TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```

Here's how I'm calling the encode function
```
    def encode(self, X):
        return self.model.encode(X, convert_to_numpy=True, transfer_to_cpu=True)
```

Quetzalcohuatl:
@paolorechia It looks like the problem is in the file `/home/paolo/dev/openimagegenius/gpu-code/gpu-node/lib/python3.10/site-packages/sentence_transformers/evaluation/EmbeddingSimilarityEvaluator.py` specifically on Line 77. You need to edit it to have `transfer_to_cpu=True`, because by default it's False.

Here is an example of it working correctly.
image

paolorechia:
Hey, @Quetzalcohuatl thanks for the advice. Fortunately, I managed to solve my problem in another way.

tomaarsen:
Hello!

The OOM path you targeted has been addressed: `encode` with `convert_to_numpy=True` already moves outputs to CPU as it accumulates, and `convert_to_numpy=False` now respects the `device` argument. Closing as resolved.

- Tom Aarsen

PR #1743, opened 2022-11-02
Fix save_to_hub for last huggingface_hub version

### LINKED ISSUE #1742: RepositoryNotFoundError when saving to Huggingface
When calling `save_to_hub` on a model with the parameter `private=True` the following error is thrown:
```
RepositoryNotFoundError: 401 Client Error. (Request ID: Nz1KbaaqnLXON_gxODABt)

Repository Not Found for url: https://huggingface.co/api/repos/create.
Please make sure you specified the correct `repo_id` and `repo_type`.
If the repo is private, make sure you are authenticated.
Unauthorized - Unauthorized
```

It seems to be because of [`huggingface-hub`](https://pypi.org/project/huggingface-hub/#history) version `0.10.1`, as it does not happen on version `0.9.1`. I'm unsure why this is a problem, but I started getting it recently when using Google Colab.

Code to reproduce:
```
model = SentenceTransformer('my_private_model', use_auth_token="x")

model.save_to_hub("my_new_private_model", private=True)
```

osanseviero:
Fixes #1742 

This should work both with old and new `huggingface_hub` versions

`organization` and `name` parameters have been deprecated since some releases ago, and it was announced they were going to be deprecated in favour of `repo_id`. At the same time, positional parameters are no longer supported, which led to users not being able to push with the latest version of `huggingface_hub`

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -464,9 +464,12 @@ class SentenceTransformer(nn.Sequential):
                 raise ValueError("You passed and invalid repository name: {}.".format(repo_name))
 
         endpoint = "https://huggingface.co"
+        repo_id = repo_name
+        if organization:
+          repo_id = f"{organization}/{repo_id}"
         repo_url = HfApi(endpoint=endpoint).create_repo(
-                token,
-                repo_name,
+                repo_id=repo_id,
+                token=token,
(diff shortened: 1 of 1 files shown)


osanseviero:
cc @nreimers

nreimers:
Thanks :)

PR #1744, opened 2022-11-03
changes to run clover

eunseojo:


tomaarsen:
Hello!

5975b02014669f4b8fb398facea3b5a354f02430 is a bit of an odd commit, it seems to include a bunch of regressions, so I can't tell what you actually wanted to edit/change here.
f6a0082aff1344d518745c4799265f7595286f7c is also a bit odd
013ba80af22568024788e1593fd35310e3f886a6 is interesting, but these files should not be in the root of the repo & they don't run.

I don't think I can merge this. That said, I am quite curious about your findings. Do you want to share what you discovered? E.g. what worked well & what didn't?

- Tom Aarsen

PR #1748, opened 2022-11-08
Fix typo: emeddings -> embeddings

victorjmarin:
I thought that "metrices" was also a typo. I was glad to learn that it is a less common plural of metric after seeing it throughout the library :smiley:.


nreimers:
Thanks :)

PR #1749, opened 2022-11-09
Fixed bug of loading DPRContextEncoder to the Transformer,

mosheber:
This allows the proper loading of DPRContextEncoder pretrained models.


tomaarsen:
Hello!

The `self.auto_model = AutoModel.from_pretrained(model_name_or_path, config=config, cache_dir=cache_dir, **model_args)` line right after the one that you proposed should also correctly load a `DPRContextEncoder` model if the model uses that architecture in its configuration. 

Beyond that, a DPR encoder model doesn't seem to be compatible with the rest of Sentence Transformers, so I don't think this should be merged as it stands right now. Feel free to respond if you disagree, though!

- Tom Aarsen

PR #1751, opened 2022-11-10
Add truncation argument

shirayu:
Currently truncation method is fixed to `longest_first`.
This PR enables use of other methods like `only_first`.

- Reference: https://huggingface.co/docs/transformers/pad_truncation

## Context

Tokenized sequences for the same text will vary depending on other inputs if `batch_size > 1`.
This is because the embeddings generated for the same text are not always the same.
I would like to fix it in an app.

### Example

```python
model = SentenceTransformer('model_name')
texts = [
               "short text",
               "short text",
               "long long long long long long text",
               "short text",
]
model.encode(texts, batch_size=2)
```

The embeddings for `short text` are not the same.

### DIFF
--- a/sentence_transformers/cross_encoder/CrossEncoder.py
+++ b/sentence_transformers/cross_encoder/CrossEncoder.py
@@ -18,7 +18,7 @@ logger = logging.getLogger(__name__)
 
 
 class CrossEncoder():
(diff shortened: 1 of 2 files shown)


nreimers:
This issue is not connected to truncation and will not be solved by that.

It comes from a bug in HF transformers that padding tokens influence the attentions. But the impact is not relevant, returned embeddings are really close and no impact on downstream tasks.

PR #1754, opened 2022-11-11
Add return type hints to util methods

zachschillaci27:
Adding return type hints to the main methods of `sentence_transformers/util.py`


tomaarsen:
Hello Zach!

Thanks for this PR! I believe all of your proposed typehints are correct. I've merged it with main & will be merging :)

- Tom Aarsen

PR #1765, opened 2022-11-28
change save latest model to save best model

kubapok:
According to sentence_transformers/cross_encoder/CrossEncoder.py  https://github.com/UKPLab/sentence-transformers/blob/e1157ce45811df59f8ae817c1a10e12ded7ff51d/sentence_transformers/cross_encoder/CrossEncoder.py#L116 the save_best_model: bool = True, is the default parameter in CrossEncoder.fit method. So the best model will be saved, not the last.


nreimers:
Here it is actually the last model. The best model will not automatically be loaded from disc in the fit method

PR #1775, opened 2022-12-02
Fix error when encoding empty set of strings with convert_to_tensor=True

oToToT:


lsyin (approved):

sentence_transformers/SentenceTransformer.py
```diff
@@ -192,7 +192,10 @@ def encode(self, sentences: Union[str, List[str]],
         all_embeddings = [all_embeddings[idx] for idx in np.argsort(length_sorted_idx)]
 
         if convert_to_tensor:
-            all_embeddings = torch.stack(all_embeddings)
+            if len(all_embeddings) == 0:
+                all_embeddings = torch.Tensor()
+            else:
+                all_embeddings = torch.stack(all_embeddings)
```
tomaarsen, line 198:
```suggestion
            if len(all_embeddings):
                all_embeddings = torch.stack(all_embeddings)
            else:
                all_embeddings = torch.Tensor()
```

tomaarsen:
Hello!

Thanks for proposing this fix! I changed up the order, but otherwise kept everything as it was.

- Tom Aarsen

PR #1812, opened 2023-01-19
Added move_to_cpu flag in encode method to avoid OOM for token embeddings.

bodin-e:
Currently the solution to resolve OOM during prediction (see #522 and #487) is only applicable to sentence embeddings, not token embeddings. This change generalizes the previous solution to also be applicable to token embeddings via the use of the added 'move_to_cpu' flag.


tomaarsen:
Hello!

The OOM path you targeted has been addressed: `encode` with `convert_to_numpy=True` already moves outputs to CPU as it accumulates, and `convert_to_numpy=False` now respects the `device` argument. Closing as resolved.

- Tom Aarsen

PR #1822, opened 2023-02-03
Add gradient accumulation to SentenceTransformer

vsuarezpaniagua:
This PR adds gradient accumulation to SentenceTransformer training, which might be very useful when using larger models or training with larger batch sizes.

*Based on #1092

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -579,6 +579,7 @@ class SentenceTransformer(nn.Sequential):
             steps_per_epoch = None,
             scheduler: str = 'WarmupLinear',
             warmup_steps: int = 10000,
+            accumulation_steps: int = 1,
             optimizer_class: Type[Optimizer] = torch.optim.AdamW,
             optimizer_params : Dict[str, object]= {'lr': 2e-5},
             weight_decay: float = 0.01,
@@ -605,6 +606,7 @@ class SentenceTransformer(nn.Sequential):
         :param steps_per_epoch: Number of training steps per epoch. If set to None (default), one epoch is equal the DataLoader size from train_objectives.
         :param scheduler: Learning rate scheduler. Available schedulers: constantlr, warmupconstant, warmuplinear, warmupcosine, warmupcosinewithhardrestarts
         :param warmup_steps: Behavior depends on the scheduler. For WarmupLinear (default), the learning rate is increased from o up to the maximal learning rate. After these many training steps, the learning rate is decreased linearly back to zero.
+        :param accumulation_steps: Number of steps to accumulate before performing a backward pass
         :param optimizer_class: Optimizer
         :param optimizer_params: Optimizer parameters
         :param weight_decay: Weight decay for model parameters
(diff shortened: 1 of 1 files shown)


nreimers:
Thanks for the PR. Sadly gradient acc doesn't provide any benefits for training embedding models

vsuarezpaniagua:
Thanks for the quick reply. However, performance in training SBERT is clearly increasing when I use larger batches for some Contrastive Learning tasks.

nreimers:
Yes, there it works as you have more negatives in a batch. But with grad acc you miss these additional negatives. There I don't expect any impact at all.

vsuarezpaniagua:
grad acc integrates more examples in the batch, so it actually adds more negative examples. As far as I know, having a batch size of N samples performs the same as having batches of N/M samples with M grad acc steps.

nreimers:
Sadly not. The loss is computed with N/M negatives and then just averaged.
Eg a batch size of 512 works great, while batch size of 1 with 512 grad acc steps doesn't work at all, as you have zero negatives in each batch.

You need more clever ways to scale to larger batches for contrastive training

vsuarezpaniagua:
In the batch size of 1 with 512, the loss is calculated for each sample and then average.

vsuarezpaniagua:
In both cases, the final loss is calculated for each individual sample and then averaged.

vsuarezpaniagua:
Hi @nreimers ,
I was thinking about what you said and I think there was a bit of misunderstanding. If the gradient accumulation is applied to loss functions that are pairwise computed, like ContrastiveLoss, CosineSimilarityLoss, or TripletLoss among others, it will work great because the loss calculation will be the same having M steps since it averages the individual examples. So, having the gradient accumulation implemented it would be an additional feature to the SBERT repository for training larger models or low memory GPUs. A note could be added in the comments saying that it will only work for pairwise computed losses (batch independent).

shhshn:
Hi. Thanks for the PR!
CrossEncoder PR #1092 actually [worked for me](https://github.com/CyberAgentAILab/japanese-nli-model).
Obtained better results compared with no accumulation when you have very limited GPU memory, say 12 GB.

I am not sure whether it will work for non-regression objectives as discussed above, but can we give it a try?

sentence_transformers/SentenceTransformer.py
```diff
@@ -710,24 +712,29 @@ def fit(self,
...
-                        optimizer.step()
+                        if (i + 1) % accumulation_steps == 0:
+                            torch.nn.utils.clip_grad_norm_(loss_model.parameters(), max_grad_norm)
+                            optimizer.step()
+                            optimizer.zero_grad()
 
-                    optimizer.zero_grad()
 
-                    if not skip_scheduler:
+                    if not skip_scheduler and (i + 1) % accumulation_steps == 0::
```
FajnyKarton, line 737: There is a typo: '::'

FajnyKarton:
Hi @vsuarezpaniagua 
Thanks for the PR. I am training my ST model with gradient accumulation and TripletLoss. It really helped me achieve the desired result. I definitely opt for adding it.

thearod5:
I support this being added in. Or at least allowing an option to easily be able to toggle this on and off. Currently have to override the SentenceTransformer class and a Copy-pasted method with gradient accumulation. Would be nice to not have this be that case.

tomaarsen:
Hello!

Since the v3.0 release, you can now use gradient accumulation with Sentence Transformers via the `gradient_accumulation_steps` [Training Argument](https://sbert.net/docs/sentence_transformer/training_overview.html#training-arguments), so I'll be closing this PR.
I do want to note that Nils' comment still holds true: gradient accumulation will not improve your performance with losses that employ in-batch negatives as there will not be more negatives to choose from. That said, it can still be used for the normal reasons.

- Tom Aarsen

PR #1833, opened 2023-02-21
add ONet transfer learning for Job Embedding

ginofft:
ONet dataset have many jobs titles referring to the same occupation.
For example, `Computer Engineer`, `Computer Software Engineer` and `Developer` all refers to `Software Developers`

Given this dataset, we hope to train an embedding that:
- Minimize distance between similar jobs
- Maximize distance between different jobs
- Can be used to deal with out-of-dataset job title, and embed them into similar jobs in the ONet Dataset.

This example features:
- Usage of SentenceLabelDataset
- Usage of Triplet Loss (set margin = 0.3167, default margin lead to model collapse)
- Usage of Triplet Evaluator
- some test cases


tomaarsen:
Hello!

The v3.0 Sentence Transformers release has modernized the training loop to no longer use `model.fit`. Although that's still possible to keep old scripts working, I'd like to avoid it in the `examples`.
With other words, I won't accept this PR as is. I'll close it for now.

- Tom aarsen

PR #1845, opened 2023-02-27
Add offline mode

### LINKED ISSUE #1725: Offline mode
Hello! 👋 

Would it be possible to have an offline mode similar to Hugginface's `TRANSFORMERS_OFFLINE=1` environment variable documented [here](https://huggingface.co/docs/transformers/installation#offline-mode)?

### Example
For example, if I call `SentenceTransformer(model_name)`, it downloads to cache. If I then turn off my Internet connection and re-run, I get something like `requests.exceptions.ConnectionError: HTTPSConnectionPool(host='huggingface.co', port=443): Max retries exceeded with url: /api/models/sentence-transformers/all-MiniLM-L12-v2 (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x176176130>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))`.

I would like to tell `sentence-transformers` to go straight to the cache.

### Workaround
I can do the following workaround:
```py
SentenceTransformer(
    f"/Users/joseph/.cache/torch/sentence_transformers/sentence-transformers_{model_name}"
    if os.environ.get("TRANSFORMERS_OFFLINE") else model_name
)
```
But with `SENTENCE_TRANSFORMERS_HOME`, `TORCH_HOME` and `XDG_CACHE_HOME` in mind, it would be a lot more robust to support that inside `SentenceTransformer()`.

danielzgtg:
Fixes #1725

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -82,7 +82,9 @@ class SentenceTransformer(nn.Sequential):
 
                 model_path = os.path.join(cache_folder, model_name_or_path.replace("/", "_"))
                 
-                if not os.path.exists(os.path.join(model_path, 'modules.json')):
+                if not os.path.exists(model_path):
+                    if transformers.utils.is_offline_mode():
+                        raise ValueError("Offline mode but model {} not downloaded".format(model_name_or_path))
                     # Download from hub with caching
                     snapshot_download(model_name_or_path,
                                         cache_dir=cache_folder,


tomaarsen:
Hello!

As of #2345, if there is no internet access but the model is in the cache, then the cache model will be used without any crash. That should hopefully prevent the need for further support with the `transformers` offline mode, so I'll be closing this. I intend to include these changes in the upcoming 2.3.0 release, which should hopefully release within the next week or 2!
I do want to thank you for taking the time to prepare this PR!

- Tom Aarsen

Issue #1494, opened 2022-04-01
How to use early stopping?


ycy3675001:
Hi, there. Is there any way that early stopping can be used in this framework? I haven't found a function that can get the model's loss? Thanks a lot.

nreimers:
It is currently not implemented

david-waterworth:
With the v3.0.0 release you can use the huggingface `transformers.EarlyStoppingCallback`

tomaarsen:
@david-waterworth is right, there's more details here: https://huggingface.co/docs/transformers/en/main_classes/callback#transformers.EarlyStoppingCallback

- Tom Aarsen

DogitoErgoSum:
@tomaarsen Do you have a working example? I am currently trying with the InformationRetrieval Class and it does not work

haukelicht:
@DogitoErgoSum: Here an example when fine-tuning with `SentenceTransformerTrainer` using `TripletLoss`:

Prerequisites: You need a train and eval data split because when using early stopping when to stop finetuning depends on the development of your eval metrics obtained in held-out data.

**Step 1:** Set up an [*evaluator*](https://sbert.net/docs/package_reference/sentence_transformer/evaluation.html) that outputs evaluation metrics from applying the fine-tuned model to the held-out eval dataset. When using `TripletLoss`, the appropriate evaluator is [`TripletEvaluator`](https://sbert.net/docs/package_reference/sentence_transformer/evaluation.html#tripletevaluator).

```python
from sentence_transformers.evaluation import TripletEvaluator

dev_evaluator = TripletEvaluator(
    anchors=dev_dataset['anchor'],
    positives=dev_dataset['positive'],
    negatives=dev_dataset['negative'],
    main_distance_function='cosine', # note: not required
)
```

**Step 2:** In `SentenceTransformerTrainingArguments`, you need to define what metric to track to assess how your model performs in the eval data split while fine-tuning step by step. For example, `TripletEvaluator` outputs `"cosine_accuracy"` and we can use it in the training arguments by setting argument `metric_for_best_model` to `"cosine_accuracy"`.

```python
from sentence_transformers import SentenceTransformerTrainingArguments
training_args = SentenceTransformerTrainingArguments(
    # ... add more arguments if needed
    eval_strategy="steps",
    eval_steps=10, # you can increase this if needed
    metric_for_best_model="cosine_accuracy",
    greater_is_better=True,
   # ... add more arguments if needed
)
```

[...]

PR #1851, opened 2023-02-28
Add early stopping criteria per epoch

vsuarezpaniagua:
This PR adds the early stopping criteria per epoch (It could be implemented per steps instead) to solve #1494.


tomaarsen:
Hello!

Apologies for never getting around to reviewing this PR. Instead, this behaviour has been implemented in the v3.0 update. See https://sbert.net/docs/sentence_transformer/training_overview.html#callbacks and https://huggingface.co/docs/transformers/en/main_classes/callback#transformers.EarlyStoppingCallback for more details.

- Tom Aarsen

PR #1852, opened 2023-03-02
Simplify the smart_batching_collate function

vsuarezpaniagua:


tomaarsen:
Hello!

This is looking great! Thanks for the simplification. I added some type hints as well, and will be merging when the tests go green.

- Tom Aarsen

PR #1857, opened 2023-03-06
Optimized Community Detection using Binary Search

arpanbiswas97:
For finding the new cluster, we are currently doing a linear search to find all the elements greater than threshold
https://github.com/UKPLab/sentence-transformers/blob/3e1929fddef16df94f8bc6e3b10598a98f46e62d/sentence_transformers/util.py#L384
Since the array is sorted, I am using binary search here to find the last element which is greater than the threshold. This is causing a noticeable improvement in performance in some cases. Thanks to @arunpurohit3799 for the help.

### DIFF
--- a/sentence_transformers/util.py
+++ b/sentence_transformers/util.py
@@ -381,11 +381,22 @@ def community_detection(embeddings, threshold=0.75, min_community_size=10, batch
                     sort_max_size = min(2 * sort_max_size, len(embeddings))
                     top_val_large, top_idx_large = cos_scores[i].topk(k=sort_max_size, largest=True)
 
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Thanks for proposing this PR! It showed me that there is interest/benefit in trying to improve the community detection. I realised that sorting is not even necessary, because we can use `torch` slicing to quickly get the "cutoff". I've implemented this in #2381. According to my tests, the fix in that PR is slightly faster than the one in this PR, so I will be closing this PR.

Semi-related, I also improved the performance heavily if you're using GPUs to compute the communities.

- Tom Aarsen

PR #1875, opened 2023-03-23
Acccept word2vec formats too.

mokha:
The difference between word2vec and glove embeddings (in text format) is just that in word2vec the first line indicates the size of the vocabulary and the embeddings dimensions. With this commit, the line is ignored and WordEmbeddings class is now able to read embeddings in both formats.

### DIFF
--- a/sentence_transformers/models/WordEmbeddings.py
+++ b/sentence_transformers/models/WordEmbeddings.py
@@ -103,6 +103,10 @@ class WordEmbeddings(nn.Module):
             iterator = tqdm(fIn, desc="Load Word Embeddings", unit="Embeddings")
             for line in iterator:
                 split = line.rstrip().split(item_separator)
+                
+                if not vocab and len(split) == 2: # Handle Word2vec format
+                    continue
+
                 word = split[0]
 
                 if embeddings_dimension == None:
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

I didn't realise about the difference in formats, but I was under the impression that word2vec was supported. For example:
```py
word_embedding_model = models.WordEmbeddings.from_text_file('GoogleNews-vectors-negative300.txt.gz')
```

Do you have an example of a common file that doesn't currently work?

- Tom Aarsen

tomaarsen:
I've prepared a word2vec file via gensim and its `model.wv.save_word2vec_format`, and it indeed showed me the issue here. I also found #1656, but I prefer the fix from this PR.

sentence_transformers/models/WordEmbeddings.py
```diff
@@ -103,6 +103,10 @@ def from_text_file(embeddings_file_path: str, update_embeddings: bool = False, i
             iterator = tqdm(fIn, desc="Load Word Embeddings", unit="Embeddings")
             for line in iterator:
                 split = line.rstrip().split(item_separator)
+                
+                if not vocab and len(split) == 2: # Handle Word2vec format
+                    continue
+
```
tomaarsen, line 109:
```suggestion

                if not vocab and len(split) == 2: # Handle Word2vec format
                    continue

```

Issue #1713, opened 2022-10-04
[FR] Add progress bar in paraphrase mining


PhilipMay:
This also needs much time when using paraphrase mining:

https://github.com/UKPLab/sentence-transformers/blob/a8cebb235066cacf533d073b8c8250e5e7b04c3d/sentence_transformers/util.py#L161

add a progress bar like here:

https://github.com/UKPLab/sentence-transformers/blob/a8cebb235066cacf533d073b8c8250e5e7b04c3d/sentence_transformers/SentenceTransformer.py#L159

PhilipMay:
What do you think @nreimers ? I might provide a PR.

nreimers:
If you could provide a PR that would be great

uahmad235:
Hello @PhilipMay . Are you planning to create a PR for this anytime soon or should I go ahead and do that?

PhilipMay:
I can create one...

uahmad235:
Cool. Please go ahead!

Marlon154:
I created the PR @nreimers

PR #1879, opened 2023-03-28
Add a progress bar for community detection

Marlon154:
Add a progress bar for community detection.
By default the progress bar is disabled, so the behavior does not change.
No additional dependencies since tqdm is already used.

#1713


tomaarsen:
👍 I'm in favor of this. I will get back to this next week.

- Tom Aarsen

Marlon154:
@tomaarsen do you have any news?

sentence_transformers/util.py
```diff
@@ -361,7 +361,7 @@ def community_detection(embeddings, threshold=0.75, min_community_size=10, batch
     min_community_size = min(min_community_size, len(embeddings))
     sort_max_size = min(max(2 * min_community_size, 50), len(embeddings))
 
-    for start_idx in range(0, len(embeddings), batch_size):
+    for start_idx in tqdm(range(0, len(embeddings), batch_size), desc="Find Cluster", disable=not show_progress_bar):
```
tomaarsen, line 364:
```suggestion
    for start_idx in tqdm(range(0, len(embeddings), batch_size), desc="Finding clusters", disable=not show_progress_bar):
```
I think this might be a bit clearer, next to the progress bar. Other than that, this looks great!

tomaarsen:
Thanks for this! I started looking at all the PRs from most to least recent, and I hadn't quite made it down to this PR yet :)
But thanks for the reminder & thanks for the PR! I hope to release this within the next 2 weeks or so.

- Tom Aarsen

Issue #1873, opened 2023-03-23
Project dependencies may have API risk issues


PyDeps:
Hi, In **sentence-transformers**, inappropriate dependency versioning constraints can cause risks.

Below are the dependencies and version constraints that the project is using
```
transformers>=4.6.0<5.0.0
tqdm
numpy
scikit-learn
scipy
nltk
```

The version constraint **==** will introduce the risk of dependency conflicts because the scope of dependencies is too strict.
The version constraint **No Upper Bound** and **\*** will introduce the risk of the missing API Error because the latest version of the dependencies may remove some APIs.

After further analysis, in this project,
The version constraint of dependency **nltk** can be changed to *>=3.2.3,<=3.8.1*.

The above modification suggestions can reduce the dependency conflicts as much as possible, 
and introduce the latest version as much as possible without calling Error in the projects.

The invocation of the current project includes all the following methods.

In version **nltk-3.2.2**, the API **nltk.tokenize.treebank.TreebankWordDetokenizer** and **nltk.tokenize.treebank.TreebankWordDetokenizer.detokenize** whch is used by the current project in **sentence_transformers/datasets/DenoisingAutoEncoderDataset.py** is missing.
    
<img width="737" alt="image" src="https://user-images.githubusercontent.com/109138844/227142561-8e9a0a1e-0ee9-4681-bcea-05093a697c94.png">

<details>
<summary>The calling methods from the nltk</summary>
<pre>
nltk.tokenize.treebank.TreebankWordDetokenizer
nltk.tokenize.treebank.TreebankWordDetokenizer.detokenize
</pre>
</details>

    
<details>
<summary>The calling methods from the  all methods</summary>
<pre>
sentence_transformers.models.Pooling
self.MultipleNegativesRankingLoss.super.__init__
conv
sentence_transformers.datasets.append
self._first_module
self._load_sbert_model
cos_scores_top_k_idx.cpu.tolist.cpu
fOutTest.write
ctx.Queue.put
self.decoder
bulk_data.append
torch.manual_seed
pytrec_eval.RelevanceEvaluator
args.negs_to_use.split
collections.defaultdict
torch.cuda.amp.GradScaler.unscale_
self.get_embeddings
self._last_module.get_pooling_mode_str
loss_model.backward
torch.save
src_ind.source_sentences.replace
numpy.ix_
Asym
coloredlogs.install
len
super.__init__
labels.to.append
[...]

PR #1880, opened 2023-03-30
Fix Python dependency API risk issue

PyDeps:
Fix issue https://github.com/UKPLab/sentence-transformers/issues/1873 by change dependency


tomaarsen:
Hello!

Apologies for the delay. The NLTK mandatory dependency has been removed, so this is no longer needed.

- Tom Aarsen

Issue #1644, opened 2022-07-20
Necessity of torchvision dependency


niazarak:
Hi, thanks for your work, we've been using sbert for quite some time.

We noticed that `sentence-transformers` depend on `torchvision`, and while it was necessary for CLIP implementation in previous versions, now `sentence-transformers` use CLIP from HF ([commit](https://github.com/UKPLab/sentence-transformers/commit/5ab198556c6122dc9e943fcb54af4d5f645a6365)).

Is there any reason to keep this dependency now?

nreimers:
I think the HF clip model also requires torchvision

iamhatesz:
How about moving torchvision to extras dependencies? I have a completely non-CV setup and this only complicates things for me. See e.g.: https://github.com/pytorch/vision/issues/5393

ecly:
Agreed with @iamhatesz that having it as optional would be ideal, as `torchvision` sometimes can be a headache.

glemaitre:
@tomaarsen I think this issue can be closed after the merge of https://github.com/UKPLab/sentence-transformers/pull/1881

PR #1881, opened 2023-03-30
Remove unused torchvision dependency

dvruette:
Torchvision is no longer used in the ST codebase, but it's still listed as a dependency. My guess is that whatever was using `torchvision` has since been replaced by `transformers` but the dependency was never  removed.

This has also been pointed out by #1644

### DIFF
--- a/examples/applications/image-search/README.md
+++ b/examples/applications/image-search/README.md
@@ -6,7 +6,7 @@ SentenceTransformers provides models that allow to embed images and text into th
 
 
 ## Installation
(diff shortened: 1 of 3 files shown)


scribu:
+1 for removing this. The torchvision dependency prevents upgrading from torch 1.9.0.

gkumbhat:
+1 for removing this dependency. Can we remove this?

dvruette:
@nreimers can we get some movement on this? It seems any CLIP-related `torchvision` dependency is no longer needed due to `transformers` replacing it.

phgz:
Is there anything preventing this branch from merging? I can try to contribute to the PR if there are some issues.

tomaarsen:
There is not. This PR is ready as far as I'm concerned. I'd like to get this merged in the coming days.

- Tom Aarsen

tomaarsen:
Thanks a bunch for this!

PR #1892, opened 2023-04-16
Add num_cycles for warmupcosinewithhardrestarts

datorresb:
* This commit adds a new parameter, num_cycles, to the fit method for the warmupcosinewithhardrestarts scheduler. 
* This change solves the problem of not being able to customize the number of cycles in the warmupcosinewithhardrestarts scheduler.
* The _get_scheduler method has also been modified to include the new num_cycles parameter when calling the get_cosine_with_hard_restarts_schedule_with_warmup function.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -579,6 +579,7 @@ class SentenceTransformer(nn.Sequential):
             steps_per_epoch = None,
             scheduler: str = 'WarmupLinear',
             warmup_steps: int = 10000,
+            num_cycles: int = 1,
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Apologies for the delay. Since the v3.0 release, you can get this behaviour through the `lr_scheduler_type` and `lr_scheduler_kwargs` [Training Arguments](https://sbert.net/docs/package_reference/sentence_transformer/training_args.html), as the SentenceTransformerTrainingArguments subclasses the [TrainingArguments from transformers](https://huggingface.co/docs/transformers/main/en/main_classes/trainer#transformers.TrainingArguments).

- Tom Aarsen

PR #1910, opened 2023-05-01
Expose model_args and tokenizer_args

edwardpwtsoi:
Exposed model_args and tokenizer_args in Transformer at SentenceTransformer


NouamaneTazi (approved):
LGTM! This would be very nice to have!

debanjum:
The ability to pass `model_args`, `tokenizer_args` to Transformer from SentenceTransformer seems to be supported now. Should this PR be closed if it's gone stale @edwardpwtsoi?

PR #1920, opened 2023-05-13
empty GPU cache after moving data to CPU (numpy) in each worker

ulf1:


tomaarsen:
Hello!

To my knowledge, it is [not recommended](https://discuss.pytorch.org/t/about-torch-cuda-empty-cache/34232) to use `torch.cuda.empty_cache()`, and I don't see any motivation of why it would be necessary here. So, I will be closing this PR. Feel free to let me know why you thought it was necessary/useful, though!

- Tom Aarsen

PR #1923, opened 2023-05-18
use cache even if module.json is not present in the cache. 

CharlesJu1:
Fix issus #1461. Pull #1565 checks whether module.json exist in the cache, which causes downloading from hub every time as there is no module.json from the hub.


tomaarsen:
Hello!

#2345 has refactored the model loading, so this PR is no longer relevant I'm afraid. I'll be closing it. I do want to thank you for taking the time to make this PR, it was part of the inspiration for #2345.

- Tom Aarsen

PR #2111, opened 2023-05-24
Fix indexing of lasttoken pooling for longest sequence

ssharpe42:
Sequences that span the entire seq length of the batch will have attention mask of all 1's. 

The current logic results in retrieving the **first** token embedding instead of the last since `torch.min(torch.tensor([1,1,1 ... , 1]))` will be 0. 

The new logic uses the `attention_mask.shape[1]-1` as the index to retrieve where `min(attention_mask, 1)==1`


izhx:
Nice work! Also found this bug.

sentence_transformers/models/Pooling.py
```diff
@@ -139,7 +139,9 @@ def forward(self, features: Dict[str, Tensor]):
             # attention_mask shape: (bs, seq_len)
             # Get shape [bs] indices of the last token (i.e. the last token for each batch item)
             # argmin gives us the index of the first 0 in the attention mask; We get the last 1 index by subtracting 1
-            gather_indices = torch.argmin(attention_mask, 1, keepdim=False) - 1 # Shape [bs]
+            # Any sequence where min == 1, we use the entire sequence lenth since argmin = 0
+            values, indices = torch.min(attention_mask, 1, keepdim = False)
+            gather_indices = torch.where(values==0, indices, attention_mask.shape[1]) - 1 # Shape [bs]
```
izhx, line 144: line 144, `attention_mask.shape[1]` could be replaced by the unused `seq_len` at line 138.

ssharpe42: good point, added in 581fe58939377595df0cfb5698f207bc1e777b19

tomaarsen:
Hello!

Thanks to you both for spotting this! You've provided a very elegant fix as well - I appreciate it! I'll merge when this goes green.

- Tom Aarsen

PR #2237, opened 2023-06-17
Feat: Adding CTranslate2 for accelerated inference

michaelfeil:
This PR adds CTranslate2 for accelerated inference.

Benefits:
- identical features in inference
- no breaking changes
- speedup of 1.2-1.5x
- float32 and float16 without accuracy loss (though inference in C++), unit test on stsbenchmark
- float32 with near identical embedding (up to tolerance)), added unit test against to original bert models
- memory improvement of 2x to 4x (int8, compared to float32)

Be aware of:
- Installing CTranslate2 as optional dependency
- not all models are supported, for now only those inheriting from huggingface-transformers: BertModel / XLM-RoBERTa

Please let me know, if there is something you missing. I know, the core maintainers are pretty busy, thanks for reviewing the PR.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -9,6 +9,7 @@ on:
   pull_request:
     branches:
       - master
+      - v*-release
       - v*-pre
   workflow_dispatch:
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,6 +8,7 @@ on:
   pull_request:
     branches:
       - master
+      - v*-release
   workflow_dispatch:
 
 jobs:


michaelfeil:
Not sure how much support there will be, going forward, for Ctranslate2.

tomaarsen:
Are you referring to the side of the CTranslate2 maintainers? Or from the side of this project? I want to point out that I'm considering your various issues and PRs closely @michaelfeil. In particular, templating & sped up runtimes (BetterTransformer, ONNX, CTranslate2, etc.).

The optimum folks pointed out that `transformers` will (or have?) switched to `scaled_dot_product_attention` which was one of the main benefits of `BetterTransformer`. 

- Tom Aarsen

michaelfeil:
So in regards to CTranslate2, there are two limiting factors:
- Limited Model Support: Mostly XLM-RoBERTa and Bert
- Questionable correctness: 
- Small community, especially around encoder models, which is not the primary strength. Also it does not depend on torch generally, but this PR is so intertwined with sentence-transformers, so that this does not matter.
- no support for flash-attention - I would argue its in range of the `C++ reimplementation`  of the `F.scaled_dot_product_attention` magnitude, which might be sufficent for Bert.

Pro:
- It's best in class in embeddings-speed in int8 for CPU. 
- Output seem to be functionally correct, given the benchmarks!

PR #2243, opened 2023-06-23
chore; update make_multilingual.py

rufimelo99:
Update paths for parallel dataset

Paths were no longer valid


tomaarsen:
Hello!

Well spotted! Thanks for this. I noticed there were some other places that referred to a now-removed dataset. Instead, I point to the parallel sentences corpus.

- Tom Aarsen

PR #2265, opened 2023-07-29
Saving and loading weights

jan-nicklaus:
While using sentence transformers, I noticed the following "problem": When using colab, it is advisable to chunk your dataset into multiple parts to train on, creating a loop of model.fit(), saving and loading. But when calling model.fit() for the second time, the weights are appearantly dropped. This can be fixed by saving and loading the torch state dict for model and optimizers, which is what I implemented here. I also updated the section on continuing training in the docs. I believe this small addition of a few lines of code could help out users, especially if training on colab hardware.

### DIFF
--- a/docs/training/overview.md
+++ b/docs/training/overview.md
@@ -166,6 +166,28 @@ model.fit(train_objectives=[(train_dataloader, train_loss)],
           output_path=model_save_path)
 ```
 
+However, this does not load the model and optimizer state dict from the previous training. If you wish to continue with the training weights from the previous fine-tuning, please use the save_state_dict and state_dict_path parameters:
+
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Apologies for the delay! Resuming training has been integrated in the v3.0 update: [`SentenceTransformerTrainer.train`](https://sbert.net/docs/package_reference/sentence_transformer/trainer.html#sentence_transformers.trainer.SentenceTransformerTrainer.train) accepts a `resume_from_checkpoint` parameter where you can pass a path to a training checkpoint with the trainer state. You can now resume without any loss of training progress.

- Tom Aarsen

PR #2277, opened 2023-08-18
Fix a bug in the util.community_detection algorithm 

dyaaalbakour:
One line change to fix a bug in the  community_detection algorithm.

The bug is that currently each community is sorted by the indexes. However, the first item in each community should be the centroid (as per the documentation).


tomaarsen:
Hello!

Well spotted. This indeed clashes with the docstring, and there's no reason to sort the clusters by indices: we rather keep them sorted them by similarity to the centroid (i.e. the first element). 962b2295a12a02946d7df13504a6c967f18e19e8 introduced this bug.

- Tom Aarsen

PR #2284, opened 2023-08-24
Add eval_batch_size argument to cross encoder evaluators

yunyu:


fkdosilovic:
This is a useful addition as far as I'm concerned. @yunyu Would you also like to add an option to pass the `show_progress_bar` argument in the same PR?

cc @tomaarsen

tomaarsen:
Apologies for the delay! I would indeed be interested in `batch_size` and `show_progress_bar` options for all CE Evaluators.

- Tom Aarsen

tomaarsen:
Resolved by #3643 and related. Apologies for the delay and not tackling this sooner!

- Tom Aarsen

PR #2285, opened 2023-08-24
Add support for bfloat16 automatic mixed precision

yunyu:
BF16 mixed precision can provide slightly better performance in some cases, especially with models previously trained with BF16 mixed precision (i.e. T5) due to the lack of requirement for gradient scaling. This PR adds a bf16_amp parameter to SentenceTransformer and CrossEncoder which allows users with Ampere+ GPUs to use BF16 for autocasting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ nr_*/
 /docs/make.bat
 /docs/Makefile
 /examples/training/quora_duplicate_questions/quora-IR-dataset/
+.venv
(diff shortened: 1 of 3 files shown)


tomaarsen:
Hello!

Apologies for not getting to this for so long. `bf16` support has been implemented in the v3.0 update via the [Training Arguments](https://sbert.net/docs/sentence_transformer/training_overview.html#training-arguments). I'll be closing this PR as a result.

(Note: It hasn't been implemented for CrossEncoders yet, but I'll create a Trainer for that too, rather than updating the `model.fit`)

- Tom Aarsen

PR #2301, opened 2023-09-08
Add support for asymmetric models in NoDuplicatesDataLoader

eranhirs:
Asymmetric models have a different format in which the text is actually a dictionary and not a string, as documented [here](https://www.sbert.net/docs/package_reference/models.html?highlight=asym#sentence_transformers.models.Asym). 

An error is thrown when using this format with `NoDuplicatesDataLoader`, as the data loader analyzes the text to avoid duplications, but expects a string and not a dictionary. This pull request add supports for both formats.

### DIFF
--- a/sentence_transformers/datasets/NoDuplicatesDataLoader.py
+++ b/sentence_transformers/datasets/NoDuplicatesDataLoader.py
@@ -24,14 +24,14 @@ class NoDuplicatesDataLoader:
 
                 valid_example = True
                 for text in example.texts:
-                    if text.strip().lower() in texts_in_batch:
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Both the targets here have moved on: `Asym` was deprecated in favor of `Router`, and `NoDuplicatesDataLoader` is gone in favor of `NoDuplicatesBatchSampler` on the v3 trainer path. The current sampler dedupes by hashing each row's column values or doing set instance checks with stringified inputs, so dict-shaped inputs should work correctly. Closing as no longer applicable.

- Tom Aarsen

PR #2315, opened 2023-09-25
[Fix typo] `ElasticSearch` -> `Elasticsearch`

leemthompo:
Just a case of using the wrong case. 😄


tomaarsen:
Hello!

This has been superseded by #2356. However, I've noticed several other cases of "ElasticSearch" being used instead of "Elasticsearch", so I'll replace those in a new PR! Thanks for making this PR :)

- Tom Aarsen

PR #2343, opened 2023-11-03
feat: Add `ignore_files_when_downloading` argument to `SentenceTransformer`

saattrupdan:
When using, say, the model intfloat/multilingual-e5-large, it will download both the `safetensors` and `ONNX` files from the Hugging Face Hub, which takes a long time. Currently you are ignoring a hardcoded list which types of files, and this small PR just makes it possible for the users to change that.


nreimers:
Yes, this is sadly a known problem. @tomaarsen will work on a fix that solves this without manually needing to define which files to ignore.

saattrupdan:
> Yes, this is sadly a known problem. @tomaarsen will work on a fix that solves this without manually needing to define which files to ignore.

I see! Good to hear that it's already on the pipeline. If you need help at all then let me know - closing this PR 🙂

PR #2345, opened 2023-11-06
`[refactor]` model loading - no more unnecessary file downloads

tomaarsen:
Hello!

## Pull Request overview
* Refactor the model loading;
  * No longer download the full model repository.
  * Update cache format to git style via `hf_hub_download`.
  * No longer use deprecated `cached_download`.
  * Soft deprecation of `use_auth_token` in favor of `token` as required by recent `transformers`/`huggingface_hub` versions.
* Add test to ensure that correct/appropriate files are downloaded.

## Details
In short, model downloading has moved from greedy full repository downloading to lazy per-module downloading, where no files are downloaded for `Transformers` modules.

### Original model loading steps
1. **Greedily** load the full model repository to the cache folder.
2. Check if `modules.json` exists.
3. If so, load all modules individually using the local files downloaded in the last step.
4. If not, load `Transformer` using the local files downloaded in the last step + `Pooling`.
5. Done

### New model loading steps
1. Check if `modules.json` exists locally or on the Hub.
2. If so,
  a. Download the ST configuration files (`'config_sentence_transformers.json'`, `'README.md'`, `'modules.json'`) if they're remote.
  b. For each module, if it is **not** transformers, then download (if necessary) the directory with configuration/weights for that module. If it **is** transformers, then do not download & load the model using the `model_name_or_path`.
3. If not, load `Transformer` using the `model_name_or_path` + `Pooling`.
4. Done

With this changed setup, we defer downloading any `transformers` data to `transformers` itself. In a [test model](https://huggingface.co/sentence-transformers-testing/stsb-bert-tiny/tree/main) that I uploaded with **both** `pytorch_model.bin` and `model.safetensors`, only the safetensors file is loaded. This is verified in the attached [test case](https://github.com/tomaarsen/sentence-transformers/blob/31646a9eead0b3433c8fb8f180dc817276aa0e6c/tests/test_sentence_transformer.py#L15).

## Additional changes
As required by `huggingface_hub`, we now use `token` instead of `use_auth_token`. If `use_auth_token` is still provided, then `token = use_auth_token` is set, and a warning is given. I.e. a soft deprecation.

- Tom Aarsen


sentence_transformers/SentenceTransformer.py
```diff
@@ -38,17 +39,27 @@ class SentenceTransformer(nn.Sequential):
     :param modules: This parameter can be used to create custom SentenceTransformer models from scratch.
     :param device: Device (like 'cuda' / 'cpu') that should be used for computation. If None, checks if a GPU can be used.
     :param cache_folder: Path to store models. Can be also set by SENTENCE_TRANSFORMERS_HOME enviroment variable.
-    :param use_auth_token: HuggingFace authentication token to download private models.
+    :param token: HuggingFace authentication token to download private models.
```
bwanglzu, line 42: seems `use_auth_token` is still in the constructor, no need to delete the docstring.

sentence_transformers/SentenceTransformer.py
```diff
@@ -38,17 +39,27 @@ class SentenceTransformer(nn.Sequential):
...
     """
     def __init__(self, model_name_or_path: Optional[str] = None,
                  modules: Optional[Iterable[nn.Module]] = None,
                  device: Optional[str] = None,
                  cache_folder: Optional[str] = None,
-                 use_auth_token: Union[bool, str, None] = None
+                 token: Optional[Union[bool, str]] = None,
+                 use_auth_token: Optional[Union[bool, str]] = None,
```
bwanglzu, line 49: maybe `Optional` is not needed? a default `False` would be better

sentence_transformers/SentenceTransformer.py
tomaarsen: I've deleted the docstring because `use_auth_token` will be softly deprecated. I'd rather not have deprecated arguments in the docstring.

sentence_transformers/SentenceTransformer.py
tomaarsen: `Optional` is needed whenever `None` is a valid value to pass, which it is. I've kept `str` because the `token`/`use_auth_token` is passed directly to `transformers` and `huggingface_hub`, who use `Optional[Union[str, bool]]` as the typing:

https://github.com/huggingface/transformers/blob/f1185a4a73a03d238afce1b40456588d22520dd2/src/transformers/modeling_utils.py#L2303

Note also that you can either pass:
* `str`: The token to use as HTTP bearer authorization for remote files.
* `True`: will use the token generated when running `huggingface-cli login` (stored in `~/.huggingface`).
* `False`: will not use any token.
* `None`: Same as `True` (so: not same as `False`)

sentence_transformers/util.py
```diff
@@ -419,86 +419,43 @@ def community_detection(embeddings, threshold=0.75, min_community_size=10, batch
 ######################
 
 
+def is_sentence_transformer_model(model_name_or_path: str, token: Optional[Union[bool, str]] = None) -> bool:
+    if os.path.exists(model_name_or_path):
+        return os.path.exists(os.path.join(model_name_or_path, "modules.json"))
```
bwanglzu, line 424: not sure if `modules.json` is deterministic, maybe `sentence_bert_config` is better? https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2/blob/main/sentence_bert_config.json

bwanglzu (commented):
left some very minor comments, do you think it make sense, at some point to refactor tests in pytests? i personally find it much more effective than `unittest`

sentence_transformers/SentenceTransformer.py
bwanglzu: i see, thanks for the clarification!

tomaarsen:
I also prefer `pytest`. I would indeed like to fully refactor the tests and heavily improve them. The current coverage is quite low for my tastes! Thanks for the review by the way!

- Tom Aarsen

sentence_transformers/util.py
tomaarsen: I think `modules.json` must always be there, while `sentence_bert_config.json` is only *usually* there. For example, this code makes me believe that there may be models that don't use `sentence_bert_config.json`:
https://github.com/UKPLab/sentence-transformers/blob/c5f93f70eca933c78695c5bc686ceda59651ae3b/sentence_transformers/models/Transformer.py#L137-L141

And I can find some older models that don't even use Transformers that do still have `modules.json`: https://huggingface.co/sentence-transformers/average_word_embeddings_levy_dependency/tree/main

I appreciate you looking into this though!

bwanglzu: i see it make sense. To be frankly the way it current handle config is not so beautiful lol, maybe later a better design is needed :)

tomaarsen: I agree completely! I've already started brainstorming some options, though I think my main idea would be infeasible due to some breaking changes here and there. It involves subclassing `SentenceTransformer` as `PreTrainedModel` rather than `nn.Sequential`. Then it could use more of the functionality from `transformers`, e.g. `load_in_8bit`, PEFT, etc.

The `modules.json` and `sentence_bert_config.json` would be removed in favor of placing that information inside of `config_sentence_transformers.json`, and the other folders with e.g. `..._Pooling`, `..._Dense` or `..._Normalize` could be removed as well. The configuration for that would live inside of the 1 config file: `config_sentence_transformers.json`, and the weights (e.g. for Dense) would be stored natively by `transformers` via `save_pretrained` because the SentenceTransformer is now a special subclass of `PreTrainedModel`.

My primary concerns are models that don't use `transformers`, but those are few and far between.

bwanglzu: i'm not familiar as you on the transofmrers source code, let me read a bit the `class:PretrainedModel` and get back to you with some proper thoughts

Sirri69:
Somebody for the love of god, please merge this and update pypi

Sirri69:
THANK YOU

tomaarsen:
@Sirri69 I'm on it 😉 Give it a few days.

I made updates to introduce better support if Internet is unavailable. Now, we can run the following script under various settings:
```py

model = SentenceTransformer("sentence-transformers/all-MiniLM-L6-v2")
embeddings = model.encode("This is a test sentence", normalize_embeddings=True)
print(embeddings.shape)
```

These are now the outputs under the various settings:




|          | Internet                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               | No Internet                                                                                                                                                                                                                                                                                                                                                                                            |
|----------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Cache    | `(384,)`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 | `(384,)`                                                                                                                                                                                                                                                                                                                                                                                                 |
| No Cache | `modules.json: 100%\|█████████████████████████████████\| 349/349 [00:00<?, ?B/s]`<br>`config_sentence_transformers.json: 100%\|████████████\| 116/116 [00:00<?, ?B/s]`<br>`README.md: 100%\|████████████████████████████████\| 10.6k/10.6k [00:00<?, ?B/s]`<br>`sentence_bert_config.json: 100%\|██████████████████\| 53.0/53.0 [00:00<?, ?B/s]`<br>`config.json: 100%\|██████████████████████████████████\| 612/612 [00:00<?, ?B/s]`<br>`pytorch_model.bin: 100%\|████████████████\| 90.9M/90.9M [00:06<00:00, 14.9MB/s]`<br>`tokenizer_config.json: 100%\|████████████████████████\| 350/350 [00:00<?, ?B/s]`<br>`vocab.txt: 100%\|██████████████████████████\| 232k/232k [00:00<00:00, 1.36MB/s]`<br>`tokenizer.json: 100%\|█████████████████████\| 466k/466k [00:00<00:00, 4.97MB/s]`<br>`special_tokens_map.json: 100%\|██████████████\| 112/112 [00:00<00:00, 90.1kB/s]`<br>`1_Pooling/config.json: 100%\|████████████████████████\| 190/190 [00:00<?, ?B/s]`<br>`(384,)` | OSError: We couldn't connect to 'https://huggingface.co' to load this file, couldn't find it in the cached files and it looks like sentence-transformers/all-MiniLM-L6-v2 is not the path to a directory containing a file named config.json. Checkout your internet connection or see how to run the library in offline mode at 'https://huggingface.co/docs/transformers/installation#offline-mode'. |

This is exactly what I would hope to get. 

cc: @nreimers as we discussed this.

- Tom Aarsen

peiyangL:
@tomaarsen 

Hi, I appreciate this update to support model loading without an internet connection.

However, I find that loading the model is very slow without an internet connection. My testing code is as follows:

```python
import time
start = time.time()

model = SentenceTransformer("nomic-ai/nomic-embed-text-v1", trust_remote_code=True, device='cpu')
emb = model.encode(["hello world"])
print(emb.shape)
print('time:', time.time()-start)
```

The output is as follows:

```
# without internet
<All keys matched successfully>
(1, 768)
time: 376.90756702423096

# with internet
<All keys matched successfully>
(1, 768)
time: 15.75501823425293
```

Additionally, I found that adding the `local_files_only=True` parameter speeds up model loading without an internet connection, but it is still quite slow.

```python
import time
start = time.time()

model = SentenceTransformer("nomic-ai/nomic-embed-text-v1", trust_remote_code=True, device='cpu', local_files_only=True)
emb = model.encode(["hello world"])
print(emb.shape)
print('time:', time.time()-start)

# output:
# <All keys matched successfully>
# (1, 768)
# time: 145.69492316246033
```

PR #2348, opened 2023-11-06
Edit transformers requirement to be <4.35.0

claudiatang-git:
With safetensors serialization being the default in [`transformers v4.35.0`](https://github.com/huggingface/transformers/releases), `transformers v4.35.0` is currently not compatible with the `sentence-transformers` library as is. `sentence-transformers` cannot load safetensor models as of now, only pytorch_model.bin (pickle)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ The following publications are integrated in this framework:
 
 ## Installation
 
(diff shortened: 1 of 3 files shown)


tomaarsen:
Hello!

SentenceTransformers can still load models with safetensors, I don't think there is a need to restrict the maximum `transformers` version. See also #2345, which should improve the model loading such that loading a SentenceTransformer model no longer requires the full repository (potentially with multiple model files) to be downloaded.

- Tom Aarsen

PR #2356, opened 2023-11-12
Update Elasticsearch example to use ANN

maxjakob:
Approximate nearest neighbor search was added to Elasticsearch in version 8:
https://www.elastic.co/blog/introducing-approximate-nearest-neighbor-search-in-elasticsearch-8-0

This PR updates the example script and the documentation.


examples/applications/semantic-search/semantic_search_quora_elasticsearch.py
```diff
@@ -67,13 +72,15 @@
...
-                  "dims": 768
+                  "dims": 768,
+                  "index": True,
+                  "similarity": "cosine"
                 }
               }
             }
         }
 
-        es.indices.create(index='quora', body=es_index, ignore=[400])
```
maxjakob, line 76: This argument is deprecated and I would argue that a bad request should fail hard here anyways.

examples/applications/semantic-search/semantic_search_quora_elasticsearch.py
```diff
@@ -21,13 +22,17 @@
...
 import csv
 import time
 import tqdm.autonotebook
 
 
-
-es = Elasticsearch()
+es = Elasticsearch(
+    hosts=["https://localhost:9200"],
+    basic_auth=("elastic", os.environ["ELASTIC_PASSWORD"],  # displayed at ES server startup
```
tomaarsen, line 33:
```suggestion
    basic_auth=("elastic", os.environ["ELASTIC_PASSWORD"]),  # displayed at ES server startup
```
Presumably this was missing?

tomaarsen (commented):
I haven't been able to run this yet, but these changes seem nice at a glance.

examples/applications/semantic-search/semantic_search_quora_elasticsearch.py
maxjakob: right, thanks! when is moving something last minute ever a good idea? :thinking: :smile:

maxjakob:
Awesome, thanks @tomaarsen! Do we need to make another change to see the documentation update reflected here?
https://sbert.net/examples/applications/semantic-search/README.html#elasticsearch

tomaarsen:
I will need to update the documentation, which I'll do following the next release. I intend for that to be within a week or 2.

maxjakob:
Great, thanks!

Issue #2336, opened 2023-10-23
Official procedure for adding new trained models


saskiabosma:
I use a pretrained CLIP [model](https://huggingface.co/patrickjohncyh/fashion-clip/tree/main) (base architecture is [CLIP-ViT-B-32 model](https://huggingface.co/laion/CLIP-ViT-B-32-laion2B-s34B-b79K)) that wasn't fine-tuned using the sentence-transformers library. It looks like it is possible to integrate it by adding files `modules.json` and `config_sentence_transformers.json` copied from `sentence-transformers/clip-ViT-B-32` in my model artifacts directory, and setting the correct `path`  in `module.json`.

Could you please confirm that this procedure is correct ? If so, it might be worth to add it to the docs, as other users may have similar needs !

saskiabosma:
Thanks a lot ! (cc @vinid)

saskiabosma:
@tomaarsen, what do you think of adding a few lines to the docs (specifically [hugging_face.md](https://github.com/UKPLab/sentence-transformers/blob/master/docs/hugging_face.md)) to guide users who want to use private local models ? Can I do a MR ?

tomaarsen:
Feel free! I'll be glad to review it.

- Tom Aarsen

PR #2360, opened 2023-11-28
Added doc for importing private model.

saskiabosma:
Follow up to the discussion on [issue-2336](https://github.com/UKPLab/sentence-transformers/issues/2336): adding details on how to import a local private model following HuggingFace base models/architecture but not hosted remotely on HuggingFace.

### DIFF
--- a/docs/hugging_face.md
+++ b/docs/hugging_face.md
@@ -82,6 +82,10 @@ You can automatically add to the Hub's model card a list of datasets you used to
 model.save_to_hub("my_new_model", train_datasets=["GEM/wiki_lingua", "code_search_net"])
 ```
 
+## Using local private models
+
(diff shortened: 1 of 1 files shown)


saskiabosma:
@tomaarsen here's the PR !

tomaarsen:
Hello!

I did some more digging around this PR & #2336, and I think the documentation should mention the (recommended) code-only procedure for converting a non-Sentence Transformer CLIP model to Sentence Transformers. Like always in Sentence Transformers, the model name can also just be a path to a local model. I've implemented this in #2369. I think this is more intuitive & useful for users than this PR. What do you think?

- Tom Aarsen

saskiabosma:
Hi ! I think your version is cleaner but I'd replace `from the Hugging Face Hub` by `from the Hugging Face Hub or a local path` to make clear the fact that the model doesn't need to be hosted remotely on HuggingFace (in case of private fine tuned models for example) and just needs to follow the HuggingFace structure.

tomaarsen:
@saskiabosma That's a great idea! I've added it in my PR [here](https://github.com/UKPLab/sentence-transformers/pull/2369/commits/6b9755c1a624b4ad8e7efef543abb531a1e767e2) & mentioned you as a co-author of the commit. 
I'll close this one, and merge that one! Thanks for all the help.

PR #2373, opened 2023-12-07
[`deps`] Add safetensors dependency, see #2372

### LINKED ISSUE #2372: Unable to load local trained cross encoder model
hi, I tried to load a cross encoder model I trained, but I got an Error message:
`OSError: Error no file named pytorch_model.bin, tf_model.h5, model.ckpt.index or flax_model.msgpack found in directory`

Yes, model is saved as `model.safetensors` not as `pytorch_model.bin`. I found that my virtual environment is missing the library `safetensors`. So fix was easy and fast:

`pip install safetensors`

If my problem is not local, please add lib to dependencies.

Thanks

tomaarsen:
Resolves #2372

Hello!

## Pull Request overview
* Add safetensors dependency.

## Details
#2372 mentions a case where the lack of this dependency prevents models from being loaded. Related is #2345 which should improve model loading, e.g. when a model is stored as a `model.safetensors` file.

- Tom Aarsen

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,4 @@ scipy
 nltk
 sentencepiece
 huggingface-hub
+safetensors
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,8 @@ setup(
         'scipy',
         'nltk',
         'sentencepiece',
-        'huggingface-hub>=0.4.0'
+        'huggingface-hub>=0.4.0',
+        'safetensors',
     ],
     classifiers=[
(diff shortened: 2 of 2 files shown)


tomaarsen:
`transformers` & `huggingface_hub` are being incremented in #2376

tomaarsen:
`transformers` will need to be at least 4.32.0 due to #2345, and 4.32.0 onwards has `safetensors` as a dependency, so we don't need to explicitly specify it.

- Tom Aarsen

PR #2383, opened 2023-12-17
Use all the CPU's instead of just 4 CPU's

manjunathshiva:
Currently CPU are hardcoded to use only 4 which means even if we have 64 cores it used only 4 cores which is not efficient. Use all the cores based on the number of available cores.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -239,8 +239,9 @@ class SentenceTransformer(nn.Sequential):
             if torch.cuda.is_available():
                 target_devices = ['cuda:{}'.format(i) for i in range(torch.cuda.device_count())]
             else:
-                logger.info("CUDA is not available. Starting 4 CPU workers")
-                target_devices = ['cpu']*4
+                num_cpus = os.cpu_count()
+                logger.info("CUDA is not available. Starting " +  str(num_cpus) + " CPU worker")
+                target_devices = ['cpu']*num_cpus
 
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Although using more CPUs will likely improve performance to an extent, more is not always strictly better. For example, too many CPUs often leads to worse latency, higher memory usage, and higher start-up time. Beyond simply processing time, starting up the pool with e.g. 24 CPUs is much slower than just with 4.

See here the graph of my findings:

image

In this one experiment, 8 is optimal for my setup, but weaker PCs might get their peak performance at 4. This is a fairly standard number to do this kind of processing at, so I would like to keep it as the default. After all, anyone can just provide the target devices to `start_multi_process_pool` if they want to use more/less CPUs.

- Tom Aarsen

vpanghalCohesity:
@tomaarsen  What is the unit of the y-axis (minutes/second) on your graph? What is the size of the sentence you used for benchmarking? The sentence encoder is pretty slow in my local setup, processing 1KB size sentences with ~40-50 sentences per second on 8 CPUs.

tomaarsen:
Hello!

I have to say that my experiments were very simple to get the graph relatively quickly. To answer your questions:
1. The y-axis is seconds.
2. The "sentence" is just "Hello there!", I likely should have used a more "real" sentence.
3.  I used `sentence-transformers-testing/stsb-bert-tiny-safetensors`, i.e. a tiny model.

In short, this was not exactly a normal use - a really short text with a really tiny model.

- Tom Aarsen

PR #2396, opened 2024-01-04
Add `classifier_dropout` param for CrossEncoder.

PhilipMay:
This PR adds the option to set / change the `classifier_dropout` value. This way it is possible to set the dropout value of the classification head. This can be a very important hyperparameter for classification tasks (and also regression).

**This is a non breaking change:** When the `classifier_dropout` is `None` (the default) then nothing happens.

For more background please also see:
- https://github.com/huggingface/transformers/pull/12794
- https://github.com/huggingface/transformers/issues/12781
- https://github.com/huggingface/transformers/issues/12792

PR #2398, opened 2024-01-08
[`feat`] Add `trust_remote_code` parameter to SentenceTransformer init

tomaarsen:
Hello!

## Pull Request overview
* Add `trust_remote_code` parameter to SentenceTransformer
* Disallow models from setting `trust_remote_code` via configuration

## Details
Any HF Hub model that requires `trust_remote_code` could not be reasonably loaded, except if the `sentence_bert_config.json` contains `"model_args": {"trust_remote_code": true}`. That said, you could argue that this isn't particularly reasonable either: we don't want models to specify this parameter.

Hence, I make 2 changes:
1. Add `trust_remote_code`. We'll want this parameter as a keyword argument in `SentenceTransformer` even after a potential refactor, so I'm good with already adding it.
2. Remove`trust_remote_code` from the loaded configuration options.

E.g.:
```py

model = SentenceTransformer("jinaai/jina-embeddings-v2-base-en", trust_remote_code=True)
embeddings = model.encode(['How is the weather today?', 'What is the current weather like today?'])

print(cos_sim(*embeddings))
# => tensor([[0.9341]])
```

cc: @bwanglzu

- Tom Aarsen

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -64,12 +64,16 @@ class SentenceTransformer(nn.Sequential):
     :param device: Device (like "cuda", "cpu", "mps") that should be used for computation. If None, checks if a GPU
         can be used.
     :param cache_folder: Path to store models. Can be also set by SENTENCE_TRANSFORMERS_HOME environment variable.
+    :param trust_remote_code: Whether or not to allow for custom models defined on the Hub in their own modeling files.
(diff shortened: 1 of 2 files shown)


tomaarsen:
Test failures are also described here: https://github.com/UKPLab/sentence-transformers/pull/2396#issuecomment-1877088716
They'll be resolved with #2396.

PR #2426, opened 2024-01-19
[`V3`] Loading & Saving refactor

### LINKED ISSUE #2435: Refactor model loading & saving
Hello!

## Context
This issue describes a refactor that I am planning for a v3 release of Sentence Transformers.
See #2426 for the related PR.

## Details
Sentence Transformer models can be converted to and from `PreTrainedModel` instances. We can rely on this to save all model weights (even Dense, CNN, LSTM, etc. weights) in the same `model.safetensors` file. Additionally, we can join all configuration files into one: `config_sentence_transformers.json` (or `sentence_transformers_config.json`?).

When loading such a model, we can use `AutoModel.from_pretrained` and its beneficial parameters such as `torch_dtype`, `device_map`, `load_in_8bit`, `load_in_4bit`, `low_cpu_mem_usage` and `quantization_config`.

## Pro's
* One file with model weights and one file with configuration is notably cleaner, more convenient, and more in line with how models are commonly stored on Hugging Face - this implies a lower risk of breaking changes in the future.
* We can now load models in 8bit, on specific devices, with quantization, etc.
* We can keep full backwards compatibility by allowing "old" models to still be loaded as usual.
* Models can still be loaded via `transformers`, although there might now be a (sensible) warning if some weights are ignored.

## Potential Pro's
This is a list of potential benefits that I haven't looked into far enough to know if they would really be beneficial:

- [ ] We might be able to integrate more easily with Optimum for ONNX due to `SentenceTransformers.to_pretrained_model()`.
- [ ] We might be able to integrate more easily with PEFT.
- [ ] It might be possible to update the existing Sentence Transformer models such that they can be loaded by v2 and with the new parameters on v3.

## Cons
* The new parameters (`torch_dtype`, etc.) would not be accessible for models saved in the "older" format, unless we 1) load via the old method, 2) save to a temporary folder via the new method, 3) load from the temporary folder with the extra parameters.

See example models with this new updated format here:
* [tomaarsen/all-MiniLM-L6-v2](https://huggingface.co/tomaarsen/all-MiniLM-L6-v2)
* [tomaarsen/distiluse-base-multilingual-cased-v2](https://huggingface.co/tomaarsen/distiluse-base-multilingual-cased-v2)

cc @bwanglzu @ir2718 @johneckberg

- Tom Aarsen

tomaarsen:
Resolves #2435

Hello!

## Pull Request overview
* Refactor model saving
  * If a model is `transformers`-based, then upon saving a model, inject all modules into the `transformers` `PreTrainedModel` & call `save_pretrained` on it.
  * If a model is `transformers`-based, then save a `config_sentence_transformers.json` file with type & config information on which modules are saved in the model weights.
* Refactor model loading
  * If a v3 model is detected (a model has `config_sentence_transformers.json` but not `modules.json`), then we inject the (empty) modules into a `PreTrainedModel` `__init__`. Then we load the model with `from_pretrained` & separate it back into the standard `Transformer` + additional modules.
* Add `to_pretrained_model()` that returns a `PreTrainedModel` version of the full SentenceTransformer model, modules included.

## Details
This experimental PR reworks how models are saved & loaded. I want to preface this by saying that the old methods for saving & loading still exist & work. For saving, they will be used for all non-Transformers based models and for loading they will be used for loading <v3 models.

### Why these changes?
The intention is that if the model weights are saved in one location & using the format that `transformers` expects, then we can take advantage of the additional `transformers` model loading functionality:
* `torch_dtype`
* `device_map`
* `load_in_8bit`
* `load_in_4bit`
* `low_cpu_mem_usage`
* `quantization_config`
etc.

Additionally, we might be able to integrate more easily with Optimum (for ONNX) & PEFT. I still have to look into this.

## Backwards incompatibility
To my knowledge, this refactor should **not** cause any backwards incompatibility. All old models can still be loaded, and additionally, `transformers` can still be used to load these refactored models. The only difference is that `transformers` might raise some warnings if the model weights have e.g. dense weights & transformers won't use these. This is a very sensible warning.

## TODO:
- [ ] Look into PEFT
- [ ] Look into Optimum support
- [ ] Look into updating existing Sentence Transformers models in such a way that users on v2 can still use the models.
- [ ] Deprecate `save_to_hub` in favor of `push_to_hub`. 
- [ ] Much, much more tests

## Example models in the v3 format
I've converted some models into the new v3 saved format here:
* [tomaarsen/distiluse-base-multilingual-cased-v2](https://huggingface.co/tomaarsen/distiluse-base-multilingual-cased-v2) (compared to [sentence-transformers/distiluse-base-multilingual-cased-v2](https://huggingface.co/sentence-transformers/distiluse-base-multilingual-cased-v2))
* [tomaarsen/all-MiniLM-L6-v2](https://huggingface.co/tomaarsen/all-MiniLM-L6-v2) (compared to [sentence-transformers/all-MiniLM-L6-v2](https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2))

Note in particular:
* The updated `config_sentence_transformers.json` file.
* The lack of `modules.json`.
* The lack of `1_Pooling` & `2_Dense` folders & their config and `pytorch_model.bin` files.

cc: @osanseviero @LysandreJik @younesbelkada

- Tom Aarsen


sentence_transformers/configuration.py
```diff
@@ -0,0 +1,60 @@
...
+
+from transformers import PretrainedConfig
+
+
+logger = logging.getLogger(__name__)
+
+
+class SentenceTransformerConfig(PretrainedConfig):
+    model_type = "sentence-transformers"
+    is_composition = True
```
LysandreJik, line 17: Ah, fun to depend on the PretraindConfig!

sentence_transformers/SentenceTransformer.py
```diff
@@ -175,27 +181,37 @@ def __init__(
...
+                # TODO: pass kwargs as model_args to Transformer?
+                modules = self._load_auto_model(model_name_or_path, **hub_kwargs)
 
         if modules is not None and not isinstance(modules, OrderedDict):
             modules = OrderedDict([(str(idx), module) for idx, module in enumerate(modules)])
 
+        if self.config is None:
+            self.config = SentenceTransformerConfig()
+
+        self.calling_init = True
```
LysandreJik, line 212: It's a bit hard to follow around why we need this `calling_init` and if it wouldn't be simpler to pass arguments around rather than keep a state of the init calls

sentence_transformers/configuration.py
```diff
@@ -0,0 +1,60 @@
...
+        Args:
+            save_directory (`str` or `os.PathLike`):
+                Directory where the configuration JSON file will be saved (will be created if it does not exist).
+            push_to_hub (`bool`, *optional*, defaults to `False`):
+                Whether or not to push your model to the Hugging Face model hub after saving it. You can specify the
+                repository you want to push to with `repo_id` (will default to the name of `save_directory` in your
+                namespace).
+            kwargs (`Dict[str, Any]`, *optional*):
+                Additional key word arguments passed along to the [`~utils.PushToHubMixin.push_to_hub`] method.
+        """
```
LysandreJik, line 53: It seems like there are two documentation styles in this PR :grin: 

I see the above that looks very close to `transformers` style, but the `save` method in the `SentenceTransformer.py` module has a very different style:

```
        """
        Saves all elements for this seq. sentence embedder into different sub-folders
        :param path: Path on disc
        :param model_name: Optional model name
        :param create_model_card: If True, create a README.md with basic information about this model
        :param train_datasets: Optional list with the names of the datasets used to to train the model
        """
```

I understand this may be legacy, but it'll probably be simpler to choose one and stick with one

sentence_transformers/SentenceTransformer.py
```diff
@@ -936,79 +771,404 @@ def _save_checkpoint(self, checkpoint_path, checkpoint_save_total_limit, step):
...
+        logger.warning(
+            "`SentenceTransformer._target_device` has been removed, please use `SentenceTransformer.device` instead.",
+        )
+        return self.device
+
+    @_target_device.setter
+    def _target_device(self, device: Optional[Union[int, str, torch.device]] = None) -> None:
+        self.to(device)
+
+    def to_pretrained_model(self) -> PreTrainedModel:
```
LysandreJik, line 856: That's cool!

sentence_transformers/SentenceTransformer.py
```diff
@@ -936,79 +771,404 @@ def _save_checkpoint(self, checkpoint_path, checkpoint_save_total_limit, step):
...
+                model_card = model_card.replace(name, str(value))
+
+        if model_name is not None:
+            model_card = model_card.replace("{MODEL_NAME}", model_name.strip())
+
+        with open(os.path.join(path, "README.md"), "w", encoding="utf8") as fOut:
+            fOut.write(model_card.strip())
+
+    @save_to_hub_args_decorator
+    def save_to_hub(
```
LysandreJik, line 1037: Would be awesome to eventually have `push_to_hub` here to align with all other libraries!

LysandreJik (commented):
Looks great! I mostly have nits, the overall design looks sound. Very cool to take advantage so much of the `PretrainedConfig` and `PretrainedModel`

tomaarsen:
Perhaps I may revisit this in the future to put all configuration in one file, but I deemed the `from_pretrained` support much too hacky: injecting code and hooks is not a clean enough solution.

- Tom Aarsen

PR #2427, opened 2024-01-19
Add '@k' at the end of csv file name

milistu:
## Description:
The **RerankingEvaluator** class has been updated to include the `@k` value in the output CSV filenames. This change prevents data inconsistencies that can arise when performing multiple evaluations with different `@k` values for Mean Reciprocal Rank (MRR).

## Why This Matters:
Without specifying `@k` in the filename, there's a risk of incorrectly logging data under the wrong `@k` parameter. 
For instance, without this update, running an evaluation first with `mrr_at_k=10` and then with `mrr_at_k=5` would result in both sets of results being saved in the same file (e.g., `RerankingEvaluator_results.csv`) and under the same column named "MRR@10", leading to a misleading representation of the MRR@5 results as MRR@10. 
This could lead to a situation where MRR@5 results are mistakenly interpreted as MRR@10, leading to incorrect analysis and conclusions.

## Solution Implemented:
Now, the output filenames reflect the `@k` value used in the evaluation (e.g., `RerankingEvaluator_results_@10.csv` for MRR@10 and `RerankingEvaluator_results_@5.csv` for MRR@5), ensuring clarity and accuracy in data recording and analysis.

### DIFF
--- a/sentence_transformers/evaluation/RerankingEvaluator.py
+++ b/sentence_transformers/evaluation/RerankingEvaluator.py
@@ -32,9 +32,15 @@ class RerankingEvaluator(SentenceEvaluator):
         batch_size: int = 64,
         show_progress_bar: bool = False,
         use_batched_encoding: bool = True,
+        mrr_at_k: Optional[int] = None,
     ):
         self.samples = samples
         self.name = name
+        if mrr_at_k is not None:
+            logger.warning(f"The `mrr_at_k` parameter has been deprecated; please use `at_k={mrr_at_k}` instead.")
+            at_k = mrr_at_k
+        else:
+            at_k = at_k
         self.at_k = at_k
         self.similarity_fct = similarity_fct
         self.batch_size = batch_size
--- a/sentence_transformers/evaluation/RerankingEvaluator.py
+++ b/sentence_transformers/evaluation/RerankingEvaluator.py
@@ -5,7 +5,7 @@ import os
 import csv
 from ..util import cos_sim
 import torch
-from sklearn.metrics import average_precision_score
+from sklearn.metrics import average_precision_score, ndcg_score
 import tqdm
 
 logger = logging.getLogger(__name__)
@@ -16,7 +16,7 @@ class RerankingEvaluator(SentenceEvaluator):
     This class evaluates a SentenceTransformer model for the task of re-ranking.
 
     Given a query and a list of documents, it computes the score [query, doc_i] for all possible
-    documents and sorts them in decreasing order. Then, MRR@10 and MAP is compute to measure the quality of the ranking.
+    documents and sorts them in decreasing order. Then, MRR@10, NDCG@10 and MAP is compute to measure the quality of the ranking.
 
     :param samples: Must be a list and each element is of the form: {'query': '', 'positive': [], 'negative': []}. Query is the search query,
      positive is a list of positive (relevant) documents, negative is a list of negative (irrelevant) documents.
@@ -25,7 +25,7 @@ class RerankingEvaluator(SentenceEvaluator):
     def __init__(
         self,
         samples,
-        mrr_at_k: int = 10,
+        at_k: int = 10,
         name: str = "",
         write_csv: bool = True,
         similarity_fct=cos_sim,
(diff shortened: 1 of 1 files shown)


milistu:
## Addition of NDCG Metric in RerankingEvaluator Class

This update integrates the NDCG (Normalized Discounted Cumulative Gain) metric into the `RerankingEvaluator` class for the SentenceTransformer models, alongside the existing MRR (Mean Reciprocal Rank) and MAP (Mean Average Precision) metrics. 

### Importance of Including NDCG:
- **Handling Multiple Relevant Documents:** In many real-world scenarios, such as when using Retrieval-Augmented Generation (RAG) systems, the answer to a query is often distributed across multiple documents. NDCG is particularly valuable in these contexts as it evaluates the entire ranking list, taking into account the varying degrees of relevance among these documents.
- **Positional Weighting:** NDCG emphasizes the importance of the position of relevant documents in the ranking list, which aligns with practical needs where the top-ranked documents are often more critical.

### MRR vs. NDCG:
- **MRR Focus:** MRR measures how well a model identifies the most relevant document, emphasizing the rank of the first relevant item.
- **NDCG Focus:** NDCG assesses the quality of the entire list of documents based on their relevance and order, making it a more comprehensive metric for evaluating ranking systems.

### Implementation:
The NDCG metric is calculated alongside MRR and MAP at the same `@k` value, ensuring consistent and comprehensive evaluation. The results are included in the output CSV file with distinct columns for each metric: MAP, MRR@k, and NDCG@k. This enhancement significantly improves the utility of the `RerankingEvaluator` for in-depth performance analysis of ranking systems.

tomaarsen:
Hello!

At a glance, this is looking great. I am in favor of both changes, and the implementations look good. I also appreciate the detailed descriptions. I do have one comment to prevent breaking changes: I don't want to fully remove the `mrr_at_k` parameter, as some people might use that in their scripts. I'll propose a solution.

Thanks for this work!

- Tom Aarsen

sentence_transformers/evaluation/RerankingEvaluator.py
```diff
@@ -35,7 +35,7 @@ def __init__(
     ):
         self.samples = samples
         self.name = name
```
tomaarsen, line 37:
```suggestion
        mrr_at_k: Optional[int] = None,
    ):
        self.samples = samples
        self.name = name
        if mrr_at_k is not None:
            logger.warning(f"The `mrr_at_k` parameter has been deprecated; please use `at_k={mrr_at_k}` instead.")
            at_k = mrr_at_k
        else:
            at_k = at_k
```

sentence_transformers/evaluation/RerankingEvaluator.py
```diff
@@ -35,7 +35,7 @@ def __init__(
...
-        self.mrr_at_k = mrr_at_k
+        self.at_k = at_k
```
tomaarsen, line 38:
```suggestion
```

sentence_transformers/evaluation/RerankingEvaluator.py
```diff
@@ -5,7 +5,7 @@
 import csv
 from ..util import cos_sim
 import torch
-from sklearn.metrics import average_precision_score
+from sklearn.metrics import average_precision_score, ndcg_score
 import tqdm
```
tomaarsen, line 10:
```suggestion
from typing import Optional

```

tomaarsen (changes requested):
I'd like to see the `__init__` updated roughly like what I've proposed in my comments. For some reason, GitHub wouldn't let me make it all 1 comment, apologies for that.
The idea is that if someone still  has some old code that uses `mrr_at_k`, then their code won't suddenly crash.

sentence_transformers/evaluation/RerankingEvaluator.py
milistu: Thank you for the fast feedback and your suggestions! This completely makes sense.

milistu:
I added on top of your proposed solution small changes to at_k assigning, there was a missing `self.`. Also, as you advised me, I ran the `make style` command, and everything should be working now.

Thank you again!

tomaarsen:
Awesome! I'll give this another look on Monday and then merge it.

tomaarsen:
This is looking good, I think. Thanks a lot for this work!

PR #2434, opened 2024-01-23
Added more saved parameter in Transformer

zinccat:
Added more saved parameter in Transformer to make saving/loading models with parameters like 'trust_remote_code=True' to work correctly

### DIFF
--- a/sentence_transformers/models/Transformer.py
+++ b/sentence_transformers/models/Transformer.py
@@ -29,7 +29,7 @@ class Transformer(nn.Module):
         tokenizer_name_or_path: str = None,
     ):
         super(Transformer, self).__init__()
-        self.config_keys = ["max_seq_length", "do_lower_case"]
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

To my knowledge, a lot of the tokenizer/model args will also be saved in their respective configuration (e.g. if you specify `bos_token` in the `tokenizer_args`, it should be included in e.g. `tokenizer_config.json`). The `trust_remote_code` option is indeed an exception, but that one is purposefully excluded from working via configuration:
https://github.com/UKPLab/sentence-transformers/blob/0116721cc9862bfcc63ce1e736d7e9a706a645bc/sentence_transformers/models/Transformer.py#L184-L186

After all, this parameter exists specifically to warn users to have a look at custom models before adding `trust_remote_code=True`. If it could be set via configuration, that would defeat the purpose.

So, I'm kind of lost with other parameters for which it could be useful to include these additional saved parameters.

- Tom Aarsen

zinccat:
Got it, thanks!

Issue #2443, opened 2024-01-23
Improve documentation of loss functions


tomaarsen:
Hello!

## Context
This issue describes documentation improvements that I would like to see.

## Details
Sentence Transformers, and the embedding space as a whole, features a lot of different loss functions, each designed with different purposes and with different inputs. Ideally, I would like to update the docstrings of all loss functions such that they contain more structured data. E.g. they have a table with the valid input formats, like for `MultipleNegativesRankingLoss`:

| Texts | Label |
|-|-|
| (anchor, positive) pairs | - |
| (anchor, positive, negative) triplets | - |

Or for `SoftmaxLoss`:
| Texts | Label |
|-|-|
| (sentence_A, sentence_B) pairs | class |

Some ideas:
- [ ] Input data format.
- [ ] Relations to other loss functions (e.g. "equivalent to `MultipleNegativesRankingLoss` but with caching" or "equivalent to `MSELoss` but with a margin")
- [ ] Purpose, if relevant.
- [ ] A code snippet.

With other words, currently the [loss docs](https://sbert.net/docs/package_reference/losses.html) can't be feasibly understood by someone who is experimenting with the project. I would like to change that.

cc @bwanglzu @ir2718 @johneckberg @aamir-s18 as I know you're interested in my TODO list. 

- Tom Aarsen

ir2718:
@tomaarsen Could you elaborate a bit more on what you mean by purpose?

tomaarsen:
Definitely!
Some loss functions were specifically introduced for certain purposes, e.g. for knowledge distillation or NLI. I don't strictly need "purpose" as that might indeed be too vague or restrictive (after all, some losses may have been introduced for NLI, but work wonders for IR as well).

So I'll explain the goal that I see with it instead: for users to be able to filter out loss functions that certainly won't be useful for them. For example, if you don't have hard negatives or similarity scores.

- Tom Aarsen

PR #2447, opened 2024-01-24
[`docs`] Refactor & improve the loss documentation

ir2718:
## PR Overview
- Adding input data format tables for each loss
- Adding relations for each loss
- Adding requirements for using each loss
- Adding working examples for each of the losses
- Fixing a few typos

## Details
This is the pull request for improving the documentations of loss functions https://github.com/UKPLab/sentence-transformers/issues/2443. The only thing that's different than the original ideas is the requirements section. I thought this was better suited for people to filter out what they can and can't try out. 

Disclaimer: I'm not an expert on all of these losses, so in case something is incorrect or something should be done a bit different, please point it out.


tomaarsen:
This is looking excellent at a glance! I will look into this with much more detail soon. Thanks already!

- Tom Aarsen

tomaarsen:
Thanks a bunch for spearheading this @ir2718!
I've done some more reformatting; there was a weird css rule that prevented some parts of the docs from looking right. I've resolved it & then applied some small adjustments throughout. Now all loss documentation should have these clear "sections": 
image

Additionally, I've added a large table at the top of the loss documentation. It should help users determine which loss functions may be relevant for them, or perhaps rather, allow them to quickly narrow down their search by ignoring the ones that are not relevant.

image

I hope that this keeps the loss functions *somewhat* parse-able as we inevitably implement more loss functions over time.

cc: @osanseviero though not related to the docs sprint; this improvement should make the docs more useful for advanced users in particular. Improving the docs for newer users is still on my TODO.

- Tom Aarsen

PR #2467, opened 2024-02-03
Correct Semantic Search Code Snippet

DhruvSrikanth:
Add the correct `model_name` for the example in the docs on **Semantic Search**. Align evaluation function (dot product) with model trained toward dot products for multi-QA task code block shown.

### DIFF
--- a/docs/pretrained_models.md
+++ b/docs/pretrained_models.md
@@ -28,7 +28,7 @@ The following models have been specifically trained for **Semantic Search**: Giv
 ```python
 from sentence_transformers import SentenceTransformer, util
 
-model = SentenceTransformer("multi-qa-MiniLM-L6-cos-v1")
+model = SentenceTransformer("multi-qa-MiniLM-L6-dot-v1")
 
 query_embedding = model.encode("How big is London")
 passage_embedding = model.encode([


DhruvSrikanth:
I would agree that in this case, it is okay to use both cosine similarity and dot product as the similarity function since the model chosen produces normalized embeddings. However, I still think the model name should align with the similarity function used for the following reason:
- The purpose of the documentation here is to showcase an example of how the model can be used to perform semantic search. Therefore the goal here is to provide clarity to the through an example. Hence, I believe that the model chosen and similarity function should align. If I was a newly minted NLP enthusiast, I would be confused by the misalignment between the two thereby hindering the purpose of the documentation and example.

Thoughts?

tomaarsen:
That is a very valid argument indeed. That said, I think the [multi-qa-MiniLM-L6-cos-v1](https://huggingface.co/sentence-transformers/multi-qa-MiniLM-L6-cos-v1) model is recommended over the [multi-qa-MiniLM-L6-dot-v1](https://huggingface.co/sentence-transformers/multi-qa-MiniLM-L6-dot-v1) model, even if due to the model metadata provided in the model card.
As a result, I would like to keep showing off the `cos` model.

- Tom Aarsen

DhruvSrikanth:
Thats totally fair :)

Issue #2440, opened 2024-01-23
Shorthand for combining losses at various weights


tomaarsen:
Hello!

## Context
This issue describes a feature that I am planning to be included in a release before v3, or alternatively, in v3 of Sentence Transformers.

## Details
When training embedding models, it is possible to want to include multiple losses for various datasets, e.g. see our [multi-task learning example](https://github.com/UKPLab/sentence-transformers/blob/0116721cc9862bfcc63ce1e736d7e9a706a645bc/examples/training/other/training_multi-task.py#L104-L107). Here we use SoftmaxLoss for NLI data and CosineSimilarityLoss for STSbenchmark data. However, it may also be interesting to use multiple loss functions for the same data. In this case, you can create your own loss function that internally calls the other loss functions & applies some weights to the losses, but it may be beneficial to create a weighted shorthand for this.

This does get somewhat tricky as different loss functions may require different inputs (positive pairs, text-positive-negative triplets, sentence_A-sentence_B-score triplets), etc.

This requires some more additional thought & experimentation. I haven't done enough multi-task testing myself to see if a weighted combined loss would be sufficiently valuable.

cc @bwanglzu @ir2718 @johneckberg @aamir-s18 as I know you're interested in my TODO list. 

- Tom Aarsen

PR #2471, opened 2024-02-05
AnglE loss

### LINKED ISSUE #2437: Implement AnglE loss
Hello!

## Context
This issue describes a feature that I am planning to be included in a release before v3, or alternatively, in v3 of Sentence Transformers.

## Details
Some [recent work by Li & Li, 2023](https://arxiv.org/abs/2309.12871) introduces an AnglE objective. Its inclusion in Sentence Transformers would be beneficial to allow our users to also train using this new loss function. The first author, @SeanLee97, has implemented the objective in https://github.com/SeanLee97/AnglE.

@johneckberg has expressed an interest in contributing this loss.

cc @bwanglzu @ir2718 @johneckberg @aamir-s18 as I know you're interested in my TODO list. John will lead this development, and I will review his PR. Additionally, @SeanLee97 may also have a look if he finds the time.

- Tom Aarsen

johneckberg:
### PR Overview:

- Implement AnglELoss and its corresponding distance metric (utils.pairwise_angle_sim) from https://arxiv.org/abs/2309.12871v6
- Resolve #2437 

### Details:
- @ir2718 pairwise_angle_sim is implemented in utils and has the same method signature as pairwise_cos_sim.
- I'm not happy with the readability of utils.pairwise_angle_sim as the math requires a handful of temp variables. If anyone has suggestions on this please leave a comment!
- @SeanLee97 if you have the time to review this, please comment if you see any issues!

### DIFF
--- a/docs/package_reference/losses.md
+++ b/docs/package_reference/losses.md
@@ -13,7 +13,7 @@ Feel free to consider the following tables to help narrow down your choice of lo
 | `(sentence_A, sentence_B) pairs`              | `class`                                                       | <a href="#softmaxloss">`SoftmaxLoss`</a>                                                                                                                                                                                                                                                                                         |
 | `(anchor, positive) pairs`                    | `none`                                                        | <a href="#cachedmultiplenegativesrankingloss">`CachedMultipleNegativesRankingLoss`</a><br><a href="#multiplenegativesrankingloss">`MultipleNegativesRankingLoss`</a><br><a href="#multiplenegativessymmetricrankingloss">`MultipleNegativesSymmetricRankingLoss`</a><br><a href="#megabatchmarginloss">`MegaBatchMarginLoss`</a> |
 | `(anchor, positive/negative) pairs`           | `1 if positive, 0 if negative`                                | <a href="#contrastiveloss">`ContrastiveLoss`</a><br><a href="#onlinecontrastiveloss">`OnlineContrastiveLoss`</a>                                                                                                                                                                                                                 |
-| `(sentence_A, sentence_B) pairs`              | `float similarity score`                                      | <a href="#cosentloss">`CoSENTLoss`</a><br><a href="#cosinesimilarityloss">`CosineSimilarityLoss`</a>                                                                                                                                                                                                                             |
+| `(sentence_A, sentence_B) pairs`              | `float similarity score`                                      | <a href="#cosentloss">`CoSENTLoss`</a><br><a href="#angleloss">`AnglELoss`</a><br><a href="#cosinesimilarityloss">`CosineSimilarityLoss`</a>                                                                                                                                                                                     |
 | `(anchor, positive, negative) triplets`       | `none`                                                        | <a href="#cachedmultiplenegativesrankingloss">`CachedMultipleNegativesRankingLoss`</a><br><a href="#multiplenegativesrankingloss">`MultipleNegativesRankingLoss`</a><br><a href="#tripletloss">`TripletLoss`</a>                                                                                                                 |
 
 Note that you can often convert one training data format into another, allowing more loss functions to be viable for your case. For example, (sentence_A, sentence_B) pairs with classes can be converted into (anchor, positive, negative) by sampling sentences with the same or different classes.
@@ -71,6 +71,12 @@ For example, when finetuning a small model to behave more like a larger & strong
 
 ```eval_rst
 .. autoclass:: sentence_transformers.losses.CoSENTLoss
+```
+
+## AnglELoss
+
+```eval_rst
+.. autoclass:: sentence_transformers.losses.AnglELoss
 ```
 
 ## CosineSimilarityLoss
--- a/sentence_transformers/losses/AnglELoss
+++ b/sentence_transformers/losses/AnglELoss
@@ -1,0 +1,58 @@
+from sentence_transformers import losses, SentenceTransformer, util
+
+
+class AnglELoss(losses.CoSENTLoss):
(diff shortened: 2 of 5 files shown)


johneckberg:
@tomaarsen Do you have an idea of why only the ubuntu unit tests are failing?

SeanLee97:
@johneckberg, many thanks for your implementation! Could it combine with contrastive loss?

tomaarsen:
@johneckberg I'm not sure, no. The logs are very confusing too, it seems like the runners just die. I'll investigate it more.

@SeanLee97 do you mean the regular cosine loss (the cosent one) and in-batch negatives? I think maybe it makes most sense to have AnglELoss as only the angle-optimized objective, and to allow users to mix and match losses themselves to reproduce your final loss function?

- Tom Aarsen

johneckberg:
@tomaarsen that's strange, thanks for the clarification!

@SeanLee97, there is an open issue #2440 regarding combining losses. Different loss functions in the library require different input formats, so combining contrastive loss/MNR Loss with AnglE Loss is not possible yet. Please leave a comment if you have any ideas for a solution to this!

johneckberg:
Hey @tomaarsen! I also noticed this performance differential between CoSENT and AnglE when performing informal tests. This is somewhat visible in the [ablation study](https://arxiv.org/pdf/2309.12871v1.pdf) in the AnglE paper, noting a small (.13%) performance increase when using just CoSENT over just AnglE. How did you combine MNR with AnglE and CoSENT?

tomaarsen:
I'll have to dive back into the ablation study!

I used:
```python
class FullAngleLoss(nn.Module):

    def __init__(self, model) -> None:
        super().__init__()
        self.angle_loss = losses.AnglELoss(model=model)
        self.cosent_loss = losses.CoSENTLoss(model=model)
        self.ibn = losses.MultipleNegativesRankingLoss(model=model)

    def forward(self, sentence_features, labels):
        return self.angle_loss(sentence_features, labels) + self.cosent_loss(sentence_features, labels) + self.ibn(sentence_features, labels)

train_loss = FullAngleLoss(model=model)
```

tomaarsen:
Just had another look at the ablation study - the findings mirror mine quite closely!

SeanLee97:
hi @tomaarsen @johneckberg , thanks for your testing!

Here are some suggestions from UAE training to achieve good performance:

- Combine the angle loss with MultipleNegativesRankingLoss (MNRL), without combining it with CoSENT.
- When working with small batch sizes, consider increasing the weight of MNRL. In the case of UAE training, we set the batch size to 64, with a weight of 20 for MNRL and a weight of 1.0 for the angle loss.

As for NLI (multinli + snli), we just used the entailment (label 1) and contradict (label 0) data for training.

tomaarsen:
Very useful information! I will try to run some extra experiments.

tomaarsen:
@johneckberg Oh, you're super right. It makes no sense to apply MNRL on the `training_stsbenchmark.py` example as it doesn't exclusively have anchor-positive pairs or anchor-positive-negative triplets. That's an oversight on my part.

- Tom Aarsen

johneckberg:
@tomaarsen No worries, glad I could be a second set of eyes!

I have been thinking about that sort of data formatting problem in relation to issue #2440, and can't think of any solid ways around it. In the AnglE repo, @SeanLee97 combines losses by always conforming to the y_true, y_pred input convention. Each input pair is just the first two in y_pred, the third and forth in y_pred, and so on. Then when using MNR loss, a [target matrix](https://github.com/SeanLee97/AnglE/blob/main/angle_emb/angle.py) gets created to filter the pairs by label. This is a good solution, but wouldn't work for ST.

tomaarsen:
An interesting solution is to keep the losses separate. E.g. in the current ST codebase that would entail 2 dataloaders & 2 losses (one with AnglE + CoSENT and one with MNRL) that fire round-robin style. This might be less performant, though.

tomaarsen:
In `training_stsbenchmark.py` I've changed up my custom loss function to:
```python
class FullAngleLoss(nn.Module):

    def __init__(self, model) -> None:
        super().__init__()
        self.angle_loss = losses.AnglELoss(model=model)
        self.cosent_loss = losses.CoSENTLoss(model=model)
        self.ibn = losses.MultipleNegativesRankingLoss(model=model)

    def forward(self, sentence_features, labels):
        positive_pairs = [
            {key: value[labels >= 0.8] for key, value in features.items()}
            for features in copy.deepcopy(sentence_features)
        ]
        loss = (
            self.angle_loss(sentence_features, labels)
            + self.cosent_loss(sentence_features, labels)
            + 3 * self.ibn(positive_pairs, labels)
        )
        return loss


train_loss = FullAngleLoss(model=model)
```
And this reaches a competitive 0.8486 Spearman correlation coefficient on the test set with batch_size of 64. 
For comparison, CoSENT + AnglE reaches 0.8419, pure AnglE reaches 0.8371, pure CoSENT reaches 0.8425, pure Cosine reaches 0.7918.

tomaarsen:
I believe this might be ready. Any last comments or suggestions before I move forward with this @johneckberg @SeanLee97?

- Tom Aarsen

johneckberg:
@tomaarsen I don't have any!

SeanLee97:
> I believe this might be ready. Any last comments or suggestions before I move forward with this @johneckberg @SeanLee97?
> 
> * Tom Aarsen

No more comments.

tomaarsen:
Much appreciated to you both, this is a very exciting addition.

- Tom Aarsen

PR #2477, opened 2024-02-09
[`feat`] Add prompt templates

### LINKED ISSUE #2439: Implement prompt/instruction templates
Hello!

## Context
This issue describes a feature that I am planning to be included in a release before v3, or alternatively, in v3 of Sentence Transformers.

## Details
Many recent works, e.g. [Wang et al., 2024](https://arxiv.org/pdf/2401.00368.pdf), [Li & Li, 2023](https://arxiv.org/abs/2309.12871), [Xiao et al., 2023](https://arxiv.org/pdf/2309.07597.pdf), and many more, use instructions/prompts to improve their model performance and instruct the models on the specific task at hand.

Ideally, Sentence Transformers should support this more easily by allowing prompt/instruction templates to be stored in the model configuration. For example, we could include the following two options in the configuration (e.g. `config_sentence_transformers.json`):
```json
{
    ...
    "prompts": {
        "classification": "Classify the following text:",
        "retrieval": "Retrieve semantically similar text:",
        "clustering": "Identify the topic or theme based on the text:",
    },
    "default_prompt_name": "classification",
}
```

And then the `SentenceTransformers.encode` method would also support `prompt` and `prompt_name` arguments:
```python
# Using a custom prompt
embeddings = model.encode(texts, prompt="Identify the topics:")
```
```python
# Using a prompt from the config
embeddings = model.encode(texts, prompt_name="clustering")
```
```python
# Using the default config, if one is defined
embeddings = model.encode(texts)
```

I am still very unsure about the names of all of these arguments - I think they're not amazing. Additionally, I'm considering whether the prompt should include `{}`, which will be filled with `prompt.format(text)`. This would allow `"Classify this text: {}. That was all."` or something, but then the end of the text will be cut off in the case of truncation, which is not great.

I'm definitely open to suggestions or ideas here!

cc @bwanglzu @ir2718 @johneckberg @aamir-s18 as I know you're interested in my TODO list. 
cc @intfloat

- Tom Aarsen

tomaarsen:
Closes #2439

Hello!

## Pull Request overview
* Add prompt templates

## Details
This PR adds prompt templates to SentenceTransformer models and their configuration.

### Initialization
```python

model = SentenceTransformer(
    "intfloat/multilingual-e5-large",
    prompts={
        "classification": "Classify the following text: {}",
        "retrieval": "Retrieve semantically similar text: {}",
        "clustering": "Identify the topic or theme based on the text: {}",
    },
    default_prompt_name="retrieval",
)
```
* `prompts` is a new (optional) argument that accepts a dictionary of prompts with prompt names to prompt format-strings.
* `default_prompt_name` is a new (optional) argument that determines the default prompt to be used. If None, then no prompt is used by default.

## Inference
1. Explicit `prompt` in `model.encode`:
   ```python
   embeddings = model.encode("How to bake a strawberry cake", prompt="Retrieve semantically similar text: {}")
   ```
2. Explicit `prompt_name` in `model.encode` by relying on the `prompts` loaded from a) initialization or b) the model config:
   ```python
   embeddings = model.encode("How to bake a strawberry cake", prompt_name="retrieval")
   ```
3. Via `model.default_prompt_name`, that prompt will be applied by default already:
   ```python
   embeddings = model.encode("How to bake a strawberry cake")
   ```

## Saved models
Upon saving a model (`model.save`), the `config_sentence_transformers.json` will now contain the `prompts` & `default_prompt_name`:
```python
{
  "__version__": {
    "sentence_transformers": "2.0.0",
    "transformers": "4.6.1",
    "pytorch": "1.8.1"
  },
  "prompts": {
    "classification": "Classify the following text: {}",
    "retrieval": "Retrieve semantically similar text: {}",
    "clustering": "Identify the topic or theme based on the text: {}"
  },
  "default_prompt_name": "retrieval"
}
```

## Loading models
If a model with the above configuration is loaded, then the prompts and `default_prompt_name` will be loaded directly into the model. This means that model authors can set the configuration & their users will automatically be able to use the prompts designed by the model authors.

## Todos
- [x] Documentation
- [x] Flag to exclude prompt tokens from pooling
- [x] INSTRUCTOR support

---

- Tom Aarsen

### DIFF
--- a/docs/pretrained_models.md
+++ b/docs/pretrained_models.md
@@ -132,6 +132,59 @@ We further provide this multilingual text-image model:
 
 ## Other Models
 
+### INSTRUCTOR models
+As of Sentence Transformers 2.4.0, some INSTRUCTOR models such as [hkunlp/instructor-large](https://huggingface.co/hkunlp/instructor-large) are natively supported in Sentence Transformers. These models are special, as they are trained with instructions in mind. Notably, the primary difference between normal Sentence Transformer models and Instructor models is that the latter do not include the instructions themselves in the pooling step.
+
+The following models work out of the box:
+* [hkunlp/instructor-base](https://huggingface.co/hkunlp/instructor-base)
+* [hkunlp/instructor-large](https://huggingface.co/hkunlp/instructor-large)
+* [hkunlp/instructor-xl](https://huggingface.co/hkunlp/instructor-xl)
+
+You can use these models like so:
+
+```python
+from sentence_transformers import SentenceTransformer
+
+model = SentenceTransformer("hkunlp/instructor-large")
+embeddings = model.encode(
+    [
+        "Dynamical Scalar Degree of Freedom in Horava-Lifshitz Gravity",
+        "Comparison of Atmospheric Neutrino Flux Calculations at Low Energies",
+        "Fermion Bags in the Massive Gross-Neveu Model",
+        "QCD corrections to Associated t-tbar-H production at the Tevatron",
+    ],
+    prompt="Represent the Medicine sentence for clustering: ",
+)
+print(embeddings.shape)
+# => (4, 768)
+```
+
+For example, for information retrieval:
+```python
+from sentence_transformers import SentenceTransformer
+from sentence_transformers.util import cos_sim
+
+model = SentenceTransformer("hkunlp/instructor-large")
+query = "where is the food stored in a yam plant"
+query_instruction = (
+    "Represent the Wikipedia question for retrieving supporting documents: "
+)
+corpus = [
+    'Yams are perennial herbaceous vines native to Africa, Asia, and the Americas and cultivated for the consumption of their starchy tubers in many temperate and tropical regions. The tubers themselves, also called "yams", come in a variety of forms owing to numerous cultivars and related species.',
+    "The disparate impact theory is especially controversial under the Fair Housing Act because the Act regulates many activities relating to housing, insurance, and mortgage loansâ€”and some scholars have argued that the theory's use under the Fair Housing Act, combined with extensions of the Community Reinvestment Act, contributed to rise of sub-prime lending and the crash of the U.S. housing market and ensuing global economic recession",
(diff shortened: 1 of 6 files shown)


tomaarsen:
I'm leaving the "Flag to exclude prompt tokens from pooling" and INSTRUCTOR support out of this PR. It doesn't seem feasible without tokenizing the prompt twice, which I'd like to avoid.

tomaarsen:
@osanseviero I'm curious to get your thoughts on this PR before I move forward with merging it, for example regarding the names of the arguments.

- Tom Aarsen

tomaarsen:
@Rocketknight1 I would definitely appreciate your look as well. This prompt templating is much different than `transformers`', but your feedback would be quite useful still.

- Tom Aarsen

Rocketknight1:
This looks like a clean implementation to me! The only potential issue is that I'm not sure how easily it'll extend to more complex prompt formats. The reason we used `jinja` for our prompt templates was that we needed the ability to, for example, loop over lists of messages and apply conditional formatting based on message roles.

As long as you're sure you won't need anything like that, this seems like a much simpler and cleaner solution!

tomaarsen:
I think Jinja would be overkill for embedding model prompt templates. For reference, embedding models often work with very small sequence lengths (256, 512), so prompts often stay small. Additionally, I've only seen "prompt prefixing". The current implementation allows for the input text to be included in the middle of the prompt as well, which is already a step further than any model has gone.

So at least for the time being, it doesn't look like very complex prompts would be used any time soon. I'll be using Jinja for generating model cards during training, though. But that's for another PR.

Thanks for having a look at this.

- Tom Aarsen

osanseviero:
No strong opinion, but with decoder models such as https://huggingface.co/Salesforce/SFR-Embedding-Mistral and https://huggingface.co/intfloat/e5-mistral-7b-instruct being used for embeddings, I wonder if we should explore using `transformers` template under the hood rather than creating a new template mechanism. That way we leverage `transformers` implementation and the templates live in the repos rather than on the users' side

tomaarsen:
I will investigate the viability of adopting this functionality from `transformers`. I'm concerned that it could get too messy.
Will update more.

- Tom Aarsen

tomaarsen:
I've chosen to add INSTRUCTOR support, as the recent GRIT also started excluding the prompts in their pooling. I will likely simplify the prompt code to only allow prefixes, i.e. no more `"{}"` either. This matches how prompting is done at this time. With the suffix, the INSTRUCTOR-style prompt ignoring won't work if the input is truncated.

PR #2485, opened 2024-02-19
[`feat`] Add Matryoshka loss + examples + docs

tomaarsen:
Hello!

## Pull Request overview
* Implement Matryoshka loss based on https://arxiv.org/abs/2205.13147
* plus example code
* plus docs

## Details
Dense embedding models normally produce embeddings with a fixed size, such as 768 or 1024. All further computations (clustering, classification, semantic search, retrieval, reranking, etc.) must then be done on these full embeddings. [Matryoshka Representation Learning](https://arxiv.org/abs/2205.13147) revisits this idea, and proposes a solution to train embedding models whose embeddings are still useful after truncation to much smaller sizes. This allows for considerably faster (bulk) processing.

### Use Cases
A particularly interesting use cases are to split up processing into two steps: 1) pre-processing with much smaller vectors and then 2) processing the remaining vectors as full size (also called "shortlisting and reranking"). Additionally, Matryoshka models will allow you to scale your embedding solutions to your desired storage cost, processing speed and performance.

### Training

Training using Matryoshka Representation Learning (MRL) is quite elementary: rather than applying some loss function on only the full-size embeddings, we also apply that same loss function on truncated portions of the embeddings. For example, if a model has an embedding dimension of 768 by default, it can now be trained on 768, 512, 256, 128, 64 and 32. Each of these losses will be added together, optionally with some weight:

```python
from sentence_transformers.losses import CoSENTLoss, MatryoshkaLoss

base_loss = CoSENTLoss(model=model)
loss = MatryoshkaLoss(model=model, loss=base_loss, matryoshka_dims=[768, 512, 256, 128, 64])
```

### Inference

When a model has been trained using a Matryoshka loss, then you can run inference with it using `SentenceTransformers.encode`. You must then truncate the resulting embeddings, and it is recommended to renormalize the embeddings.

```python

model = SentenceTransformer("nomic-ai/nomic-embed-text-v1.5", trust_remote_code=True)

matryoshka_dim = 64
embeddings = model.encode(
    [
        "search_query: What is TSNE?",
        "search_document: t-distributed stochastic neighbor embedding (t-SNE) is a statistical method for visualizing high-dimensional data by giving each datapoint a location in a two or three-dimensional map.",
        "search_document: Amelia Mary Earhart was an American aviation pioneer and writer.",
    ]
)
embeddings[..., :matryoshka_dim]  # Shrink the embedding dimensions

similarities = cos_sim(embeddings[0], embeddings[1:])
# => tensor([[0.7839, 0.4933]])
```
As you can see, the similarity between the search query and the correct document is much higher than that of an unrelated document, despite the very small matryoshka dimension applied. Feel free to copy this script locally, modify the `matryoshka_dim`, and observe the difference in similarities.

---

## Implementation
In the implementation, I override the `forward` method of the first module in the Sentence Transformer. Usually this'll be a `Transformer` instance, but it doesn't have to be. I expect that this method is called with the same batches every time that the "real" `loss` is called, and I abuse that with a caching mechanism. With other words, only for the first set of matryoshka dimensions do I actually compute the `forward` call, in all subsequent occasions I simply use the cached outputs. Afterwards, I truncate the `forward` outputs or the cache outputs to the correct matryoshka dimension.

### Training efficiency
As far as I can tell, this trains roughly as quickly as just applying the loss with the maximum embedding dimension.

### Note
This may not work on models that use Dense modules, and it may also cause issues with `CachedMultipleRankedNegativesLoss`.

cc: @Xenova for your interest in Matryoshka embedding models
cc: @zanussbaum as you've also implemented Matryoshka learning + because I mention Nomic in the documentation.

- Tom Aarsen

### DIFF
--- a/examples/training/matryoshka/README.md
+++ b/examples/training/matryoshka/README.md
@@ -46,6 +46,8 @@ similarities = cos_sim(embeddings[0], embeddings[1:])
 ```
 As you can see, the similarity between the search query and the correct document is much higher than that of an unrelated document, despite the very small matryoshka dimension applied. Feel free to copy this script locally, modify the `matryoshka_dim`, and observe the difference in similarities.
 
+**Note**: Despite the embeddings being smaller, training and inference of a Matryoshka model is not faster, not more memory-efficient, and not smaller. Only the processing and storage of the resulting embeddings will be faster and cheaper.
+
 ## Code Examples
 
 See the following scripts as examples of how to apply the [`MatryoshkaLoss`](../../../docs/package_reference/losses.html#matryoshkaloss) in practice:
--- a/examples/training/matryoshka/README.md
+++ b/examples/training/matryoshka/README.md
@@ -53,4 +53,5 @@ As you can see, the similarity between the search query and the correct document
 See the following scripts as examples of how to apply the [`MatryoshkaLoss`](../../../docs/package_reference/losses.html#matryoshkaloss) in practice:
 
 * **[matryoshka_nli.py](matryoshka_nli.py)**: This example uses the MultipleNegativesRankingLoss with MatryoshkaLoss to train a strong embedding model using Natural Language Inference (NLI) data. It is an adaptation of the [NLI](../nli/README) documentation.
+* **[matryoshka_nli_reduced_dim.py](matryoshka_nli_reduced_dim.py)**: This example uses the MultipleNegativesRankingLoss with MatryoshkaLoss to train a strong embedding model with a small maximum output dimension of 256. It trains using Natural Language Inference (NLI) data, and is an adaptation of the [NLI](../nli/README) documentation.
 * **[matryoshka_sts.py](matryoshka_sts.py)**: This example uses the CoSENTLoss with MatryoshkaLoss to train an embedding model on the training set of the STSBenchmark dataset. It is an adaptation of the [STS](../sts/README) documentation.
--- a/examples/training/matryoshka/README.md
+++ b/examples/training/matryoshka/README.md
@@ -1,6 +1,6 @@
 # Matryoshka Embeddings
 
-Dense embedding models normally produce embeddings with a fixed size, such as 768 or 1024. All further computations (clustering, classification, semantic search, retrieval, reranking, etc.) must then be done on these full embeddings. [Matryoshka Representation Learning](https://arxiv.org/abs/2205.13147) revisits this idea, and proposes a solution to train embedding models whose embeddings are still useful after truncation to much smaller sizes. This allows for considerably faster (bulk) processing.
+Dense embedding models typically produce embeddings with a fixed size, such as 768 or 1024. All further computations (clustering, classification, semantic search, retrieval, reranking, etc.) must then be done on these full embeddings. [Matryoshka Representation Learning](https://arxiv.org/abs/2205.13147) revisits this idea, and proposes a solution to train embedding models whose embeddings are still useful after truncation to much smaller sizes. This allows for considerably faster (bulk) processing.
 
 ## Use Cases
 
--- a/examples/training/matryoshka/README.md
+++ b/examples/training/matryoshka/README.md
@@ -1,0 +1,54 @@
+# Matryoshka Embeddings
+
+Dense embedding models normally produce embeddings with a fixed size, such as 768 or 1024. All further computations (clustering, classification, semantic search, retrieval, reranking, etc.) must then be done on these full embeddings. [Matryoshka Representation Learning](https://arxiv.org/abs/2205.13147) revisits this idea, and proposes a solution to train embedding models whose embeddings are still useful after truncation to much smaller sizes. This allows for considerably faster (bulk) processing.
+
+## Use Cases
+
+A particularly interesting use cases are to split up processing into two steps: 1) pre-processing with much smaller vectors and then 2) processing the remaining vectors as full size (also called "shortlisting and reranking"). Additionally, Matryoshka models will allow you to scale your embedding solutions to your desired storage cost, processing speed and performance.
+
+## Training
+
+Training using Matryoshka Representation Learning (MRL) is quite elementary: rather than applying some loss function on only the full-size embeddings, we also apply that same loss function on truncated portions of the embeddings. For example, if a model has an embedding dimension of 768 by default, it can now be trained on 768, 512, 256, 128, 64 and 32. Each of these losses will be added together, optionally with some weight:
+
+```python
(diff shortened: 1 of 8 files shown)


examples/training/matryoshka/README.md
```diff
@@ -0,0 +1,54 @@
+# Matryoshka Embeddings
+
+Dense embedding models normally produce embeddings with a fixed size, such as 768 or 1024. All further computations (clustering, classification, semantic search, retrieval, reranking, etc.) must then be done on these full embeddings. [Matryoshka Representation Learning](https://arxiv.org/abs/2205.13147) revisits this idea, and proposes a solution to train embedding models whose embeddings are still useful after truncation to much smaller sizes. This allows for considerably faster (bulk) processing.
```
xenova, line 3:
```suggestion
Dense embedding models typically produce embeddings with a fixed size, such as 768 or 1024. All further computations (clustering, classification, semantic search, retrieval, reranking, etc.) must then be done on these full embeddings. [Matryoshka Representation Learning](https://arxiv.org/abs/2205.13147) revisits this idea, and proposes a solution to train embedding models whose embeddings are still useful after truncation to much smaller sizes. This allows for considerably faster (bulk) processing.
```

tomaarsen: Much better indeed!

examples/training/matryoshka/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Use Cases
+
+A particularly interesting use cases are to split up processing into two steps: 1) pre-processing with much smaller vectors and then 2) processing the remaining vectors as full size (also called "shortlisting and reranking"). Additionally, Matryoshka models will allow you to scale your embedding solutions to your desired storage cost, processing speed and performance.
```
xenova, line 7:
```suggestion
A particularly interesting use case is to split processing into two steps: 1) pre-processing with much smaller vectors and then 2) processing the remaining vectors as full size (also called "shortlisting and reranking"). Additionally, Matryoshka models will allow you to scale your embedding solutions to your desired storage cost, processing speed and performance.
```

examples/training/matryoshka/README.md
```diff
@@ -0,0 +1,54 @@
...
+model = SentenceTransformer("microsoft/mpnet-base")
+
+base_loss = CoSENTLoss(model=model)
+loss = MatryoshkaLoss(model=model, loss=base_loss, matryoshka_dims=[768, 512, 256, 128, 64])
+```
+* **Reference**: [`MatryoshkaLoss`](../../../docs/package_reference/losses#matryoshkaloss)
+
+## Inference
+
+When a model has been trained using a Matryoshka loss, then you can run inference with it using [`SentenceTransformers.encode`](../../../docs/package_reference/SentenceTransformer.html#sentence_transformers.SentenceTransformer.encode). You must then truncate the resulting embeddings, and it is recommended to renormalize the embeddings.
```
xenova, line 26:
```suggestion
After a model has been trained using a Matryoshka loss, you can then run inference with it using [`SentenceTransformers.encode`](../../../docs/package_reference/SentenceTransformer.html#sentence_transformers.SentenceTransformer.encode). You must then truncate the resulting embeddings, and it is recommended to renormalize the embeddings.
```

examples/training/matryoshka/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+matryoshka_dim = 64
+embeddings = model.encode(
+    [
+        "search_query: What is TSNE?",
+        "search_document: t-distributed stochastic neighbor embedding (t-SNE) is a statistical method for visualizing high-dimensional data by giving each datapoint a location in a two or three-dimensional map.",
+        "search_document: Amelia Mary Earhart was an American aviation pioneer and writer.",
+    ]
+)
+embeddings[..., :matryoshka_dim]  # Shrink the embedding dimensions
```
xenova, line 42: It might be good to mention the additional steps (as stated in the [model card](https://huggingface.co/nomic-ai/nomic-embed-text-v1.5#sentence-transformers)):
1. Applying layer normalization after encoding (before truncation)
2. (Optional) Normalizing the truncated embeddings. Luckily since we're performing cosine similarity in this example, it doesn't matter, but just so the user is aware of it in case they're going to store the embedding at this point.

xenova (commented):
Super cool! 🚀 Looking forward to playing around with this myself too!

examples/training/matryoshka/README.md
xenova: (for reference)
```py
import torch.nn.functional as F

matryoshka_dim = 512

model = SentenceTransformer("nomic-ai/nomic-embed-text-v1.5", trust_remote_code=True)
sentences = ['search_query: What is TSNE?', 'search_query: Who is Laurens van der Maaten?']
embeddings = model.encode(sentences, convert_to_tensor=True)
embeddings = F.layer_norm(embeddings, normalized_shape=(embeddings.shape[1],))
embeddings = embeddings[:, :matryoshka_dim]
embeddings = F.normalize(embeddings, p=2, dim=1)
print(embeddings)
```

tomaarsen: Hmm, the tricky part is that this is a Nomic-specific thing I believe, not just for Matryoshka models.

zanussbaum: yeah this is a Nomic Specific thing but not necessary in general. If you want to include the Nomic Model, you should use the layernorm

tomaarsen: Hmm, that makes it a bit tricky. I'd love to show a general example that works with essentially all models.

tomaarsen: I'm seeing extremely minimal differences in performance when using layer norm vs no layer norm. 
| | 768 | 512 | 256 | 128 | 64 |
| --- | ---- |  --- | ---- | --- | ---- |
| Layer norm | 0.84247 |0.83947 | **0.83743** | **0.83185**|0.82377 |
| No layer norm | 0.84247 | 0.83947| 0.83725| 0.83168|**0.82386** |
| Ratio | 1.00000221 | 0.99999807 |  1.00020698 | 1.00020339 | 0.9998845 |

I would rather not recommend the layer norm in these docs, as they're just for general inference (not just Nomic), so either:
1. I use Nomic v1.5 without Layer Norm (should perform roughly equivalently, but not identically as you can see)
2. I use some other (unknown) model, like https://huggingface.co/tomaarsen/mpnet-base-all-nli-matryoshka

My preference is the first option, but if you think that I shouldn't list Nomic without layer norm, then I'll do the second.

Issue #2410, opened 2024-01-15
Request on a few more evaluation metrics


mucun1988:
In [evaluation metrics](https://github.com/UKPLab/sentence-transformers/tree/08a57b4a19ddaf7cccda51cd0c2c8af7bbc339a3/sentence_transformers/cross_encoder/evaluation), could we also add a few more? 

- F1 score for binary classification 
- micro-F1 score for multi-class classification
- macro-F1 score for multi-class classification

tomaarsen:
Hello!

The `CEBinaryClassificationEvaluator` does report the F1 score for binary classification, although the evaluator admittedly only returns the average precision. That said, I think these evaluators can indeed be expanded.

- Tom Aarsen

tempdeltavalue:
@mucun1988 Hi, I'm not sure what is useful for you fyi, you write your own metrics inside of custom callback 

https://github.com/UKPLab/sentence-transformers/issues/2413

mucun1988:
@tempdeltavalue I think that would help the completeness of the package.

ir2718:
@mucun1988 

For now, a workaround solution can be a custom evaluator similar to `CESoftmaxAccuracyEvaluator`. This should do the trick:
```
from typing import List
from sklearn.metrics import f1_score
import numpy as np

class CEF1Evaluator:

    def __init__(self,  problem_type: str, sentence_pairs: List[List[str]], labels: List[int], name: str = "", write_csv: bool = True):
        self.sentence_pairs = sentence_pairs
        self.labels = labels
        self.name = name

        if problem_type == "binary":
            self.f1_callables = [
                ("Binary F1 score", lambda x, y: f1_score(x, y, average="binary")),
            ]
        elif problem_type == "multiclass":
            self.f1_callables = [
                ("Macro F1 score", lambda x, y: f1_score(x, y, average="macro")),
                ("Micro F1 score", lambda x, y: f1_score(x, y, average="micro")),
                ("Weighted F1 score", lambda x, y: f1_score(x, y, average="weighted")),
            ]
        else:
            raise Exception(f"The problem_type should be either 'binary' or 'multiclass', but you used '{problem_type}'.")
            
        self.csv_file = "CEF1Evaluator" + ("_" + name if name else "") + "_results.csv"
        self.csv_headers = ["epoch", "steps"] + [name for name, _ in self.f1_callables]
        self.write_csv = write_csv

    @classmethod
[...]

PR #2493, opened 2024-02-20
Add F1 score evaluator for CrossEncoder.

fkdosilovic:
#### Related Issue

Closes https://github.com/UKPLab/sentence-transformers/issues/2410

#### Description

- Add F1 score evaluator (micro, macro, and weighted) for CrossEncoder
- Update `training_nli.py` to use SequentialEvalutor

#### Additional comments

Minor changes to [the code](https://github.com/UKPLab/sentence-transformers/issues/2410#issuecomment-1892896867) by @ir2718.

### DIFF
--- a/sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
+++ b/sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
@@ -4,6 +4,7 @@ import os
 from typing import List
 
 import numpy as np
+from .. import CrossEncoder
 from ... import InputExample
 from sklearn.metrics import f1_score
 
@@ -14,7 +15,9 @@ class CEF1Evaluator:
     """
     CrossEncoder F1 score based evaluator for binary and multiclass tasks.
 
-    The task type (binary or multiclass) is determined from the labels array.
+    The task type (binary or multiclass) is determined from the labels array. For
+    binary tasks the returned metric is binary F1 score. For the multiclass tasks
+    the returned metric is weighted F1 score.
 
     :param sentence_pairs: A list of sentence pairs, where each pair is a list of two strings.
     :type sentence_pairs: list[list[str]]
@@ -36,7 +39,7 @@ class CEF1Evaluator:
         labels: List[int],
         *,
         batch_size: int = 32,
-        show_progress_bar: bool = None,
+        show_progress_bar: bool = False,
         name: str = "",
         write_csv: bool = True,
     ):
@@ -47,32 +50,28 @@ class CEF1Evaluator:
         self.name = name
         self.write_csv = write_csv
 
-        n_unique = np.unique(labels)
-        if n_unique == 2:
-            problem_type = "binary"
-        elif n_unique > 2:
-            problem_type = "multiclass"
-        else:
-            raise ValueError(
-                "Got only one distinct label. Please make sure there are at least two labels in the `labels` array."
-            )
+        n_unique = np.unique(labels).size
 
-        if problem_type == "binary":
+        if n_unique == 2:
             self.f1_callables = [
                 ("Binary F1 score", lambda x, y: f1_score(x, y, average="binary")),
             ]
-        elif problem_type == "multiclass":
+        elif n_unique > 2:
             self.f1_callables = [
                 ("Macro F1 score", lambda x, y: f1_score(x, y, average="macro")),
                 ("Micro F1 score", lambda x, y: f1_score(x, y, average="micro")),
                 ("Weighted F1 score", lambda x, y: f1_score(x, y, average="weighted")),
             ]
+        else:
+            raise ValueError(
+                "Got only one distinct label. Please make sure there are at least two labels in the `labels` array."
+            )
 
         self.csv_file = "CEF1Evaluator" + (f"_{name}" if name else "") + "_results.csv"
         self.csv_headers = ["epoch", "steps"] + [metric_name for metric_name, _ in self.f1_callables]
 
     @classmethod
-    def from_input_examples(cls, problem_type: str, examples: List[InputExample], **kwargs):
+    def from_input_examples(cls, examples: List[InputExample], **kwargs):
         sentence_pairs = []
         labels = []
 
@@ -80,11 +79,11 @@ class CEF1Evaluator:
             sentence_pairs.append(example.texts)
             labels.append(example.label)
 
-        return cls(problem_type, sentence_pairs, labels, **kwargs)
+        return cls(sentence_pairs, labels, **kwargs)
 
     def __call__(
         self,
-        model,
+        model: CrossEncoder,
         output_path: str = None,
         epoch: int = -1,
         steps: int = -1,
@@ -112,7 +111,7 @@ class CEF1Evaluator:
         for f1_name, f1_fn in self.f1_callables:
             f1_val = f1_fn(pred_labels, self.labels)
             save_f1.append(f1_val)
-            logger.info(f"{f1_name:20s}: {f1_val:.4f} \n")
+            logger.info(f"{f1_name:20s}: {f1_val:.4f}")
 
         if output_path is not None and self.write_csv:
             csv_path = os.path.join(output_path, self.csv_file)
--- a/sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
+++ b/sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
@@ -111,7 +111,7 @@ class CEF1Evaluator:
         for f1_name, f1_fn in self.f1_callables:
             f1_val = f1_fn(pred_labels, self.labels)
             save_f1.append(f1_val)
-            logger.info(f"{f1_name:20s}: {f1_val:.4f}")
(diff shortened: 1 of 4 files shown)


sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
```diff
@@ -0,0 +1,128 @@
...
+        if problem_type == "binary":
+            self.f1_callables = [
+                ("Binary F1 score", lambda x, y: f1_score(x, y, average="binary")),
+            ]
+        elif problem_type == "multiclass":
+            self.f1_callables = [
+                ("Macro F1 score", lambda x, y: f1_score(x, y, average="macro")),
+                ("Micro F1 score", lambda x, y: f1_score(x, y, average="micro")),
+                ("Weighted F1 score", lambda x, y: f1_score(x, y, average="weighted")),
+            ]
```
tomaarsen, line 69:
```suggestion
        n_unique = np.unique(labels)
        if n_unique == 2:
            self.f1_callables = [
                ("Binary F1 score", lambda x, y: f1_score(x, y, average="binary")),
            ]
        elif n_unique > 2:
            self.f1_callables = [
                ("Macro F1 score", lambda x, y: f1_score(x, y, average="macro")),
                ("Micro F1 score", lambda x, y: f1_score(x, y, average="micro")),
                ("Weighted F1 score", lambda x, y: f1_score(x, y, average="weighted")),
            ]
        else:
            raise ValueError(
                "Got only one distinct label. Please make sure there are at least two labels in the `labels` array."
            )
```

We don't have to use the temporary variable `problem_type` I think.

sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
```diff
@@ -0,0 +1,128 @@
...
+
+        self.csv_file = "CEF1Evaluator" + (f"_{name}" if name else "") + "_results.csv"
+        self.csv_headers = ["epoch", "steps"] + [metric_name for metric_name, _ in self.f1_callables]
+
+    @classmethod
+    def from_input_examples(cls, problem_type: str, examples: List[InputExample], **kwargs):
```
tomaarsen, line 75:
```suggestion
    def from_input_examples(cls, examples: List[InputExample], **kwargs):
```
I think these are from an earlier version of your code perhaps?

sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
```diff
@@ -0,0 +1,128 @@
...
+        sentence_pairs = []
+        labels = []
+
+        for example in examples:
+            sentence_pairs.append(example.texts)
+            labels.append(example.label)
+
+        return cls(problem_type, sentence_pairs, labels, **kwargs)
```
tomaarsen, line 83:
```suggestion
        return cls(sentence_pairs, labels, **kwargs)
```

sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
```diff
@@ -0,0 +1,128 @@
...
+        write_csv: bool = True,
+    ):
+        self.sentence_pairs = sentence_pairs
+        self.labels = labels
+        self.batch_size = batch_size
+        self.show_progress_bar = show_progress_bar
+        self.name = name
+        self.write_csv = write_csv
+
+        n_unique = np.unique(labels)
```
tomaarsen, line 50:
```suggestion
        n_unique = np.unique(labels).size
```

sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
```diff
@@ -0,0 +1,128 @@
...
+
+        for example in examples:
+            sentence_pairs.append(example.texts)
+            labels.append(example.label)
+
+        return cls(problem_type, sentence_pairs, labels, **kwargs)
+
+    def __call__(
+        self,
+        model,
```
tomaarsen, line 87:
```suggestion
        model: CrossEncoder,
```
Just a personal preference (I see not all evaluators do this, that's fine).

(Plus the import at the top)

sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
```diff
@@ -0,0 +1,128 @@
...
+    :type write_csv: bool, optional
+    """
+
+    def __init__(
+        self,
+        sentence_pairs: List[List[str]],
+        labels: List[int],
+        *,
+        batch_size: int = 32,
+        show_progress_bar: bool = None,
```
tomaarsen, line 39:
```suggestion
        show_progress_bar: bool = False,
```
I think it's good to allow users to choose whether to show the progress bar! Most of the other evaluators use `False` here though, so for consistency it might be best to set this to False as the default.

sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+        pred_labels = np.argmax(pred_scores, axis=1)
+
+        assert len(pred_labels) == len(self.labels)
+
+        save_f1 = []
+        for f1_name, f1_fn in self.f1_callables:
+            f1_val = f1_fn(pred_labels, self.labels)
+            save_f1.append(f1_val)
+            logger.info(f"{f1_name:20s}: {f1_val:.4f} \n")
```
tomaarsen, line 115:
```suggestion
            logger.info(f"{f1_name:20s}: {f1_val:.4f}")
```
The logger will already add the newlines

sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
tomaarsen: Otherwise we get the numpy array, we want the number of labels instead.

sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
```diff
@@ -0,0 +1,128 @@
...
+            output_file_exists = os.path.isfile(csv_path)
+            mode = "a" if output_file_exists else "w"
+            with open(csv_path, mode=mode, encoding="utf-8") as f:
+                writer = csv.writer(f)
+                if not output_file_exists:
+                    writer.writerow(self.csv_headers)
+
+                writer.writerow([epoch, steps, *save_f1])
+
+        return save_f1[0]
```
tomaarsen, line 128: So we return either the binary F1 or the Macro F1? Perhaps we can write that in the docstring at the top (below `"The task type (binary or multiclass) ..."`)?

tomaarsen (changes requested):
Hello!

This is looking great already! I do have a few comments of changes that I needed to do to make this work + some nitpicks/preferences.

Also, I think it would be valuable if we updated e.g. this line: 
https://github.com/UKPLab/sentence-transformers/blob/ecdda29859d6a651326addd11bd237870b39866a/examples/training/cross-encoder/training_nli.py#L64-L65

And replaced it with:
```python
# During training, we use CESoftmaxAccuracyEvaluator and CEF1Evaluator to measure the performance on the dev set
accuracy_evaluator = CESoftmaxAccuracyEvaluator.from_input_examples(dev_samples, name="AllNLI-dev")
f1_evaluator = CEF1Evaluator.from_input_examples(dev_samples, name="AllNLI-dev")
evaluator = SequentialEvaluator([accuracy_evaluator, f1_evaluator])
```
Even though this is a bit overkill because accuracy will be the same as the Micro F1 score in this example, so in a real scenario we would only run `CEF1Evaluator`, but I'd like to show how to use the `CEF1Evaluator` as well as `CESoftmaxAccuracyEvaluator` in one of the training examples. 

- Tom Aarsen

sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
fkdosilovic: Good catch. Thanks.

sentence_transformers/cross_encoder/evaluation/CEF1Evaluator.py
fkdosilovic: I've decided to set the default to None because this is the [default value](https://github.com/UKPLab/sentence-transformers/blob/20056c64c2583ff08fca8df31ff81ccc3f93b882/sentence_transformers/cross_encoder/CrossEncoder.py#L284) for CrossEncoder's predict method. This way, the decision is [delegated](https://github.com/UKPLab/sentence-transformers/blob/20056c64c2583ff08fca8df31ff81ccc3f93b882/sentence_transformers/cross_encoder/CrossEncoder.py#L317), as implemented in CrossEncoder's predict method, to the settings of a logger.

Ok, I'll switch to False, but it might be a good idea to switch to None for all evaluators in the future; not sure though.

tomaarsen: We can definitely reconsider this in the future indeed. For now I'm a fan of keeping the consistency :)

PR #2500, opened 2024-02-23
Add version range for numpy to requirements

fkdosilovic:
#### Description

I ran into problems while trying to use `cross-encoder/nli-deberta-v3-xsmall` model with CrossEncoder. By default, sentence-transfomers will install latest numpy available from pip, unfortunately this introduces some breaking changes.
Specifically, while trying to fit the CrossEncoder(`cross-encoder/nli-deberta-v3-xsmall`), I'm getting the `np.int` depreciation error.

#### Additional comments

`np.int` was depricated in numpy 1.20 and removed in numpy 1.24. numpy 1.17 is the minimum version required for HF Transformers and that's why it is set here as well as the minimum version.

#### Fix

Fixed by specifying the version range for numpy.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 transformers>=4.32.0,<5.0.0
 tqdm
 torch>=1.11.0
-numpy
+numpy>=1.17,<1.24.0
 scikit-learn
 scipy
 huggingface-hub>=0.15.1


tomaarsen:
Hello!

Perhaps instead of preventing users from using numpy 1.24, we can follow the deprecation guidelines and update `np.int` with `int`? Do we use `np.int` in Sentence Transformers or is it used in a third party library?

- Tom Aarsen

fkdosilovic:
> Perhaps instead of preventing users from using numpy 1.24, we can follow the deprecation guidelines and update np.int with int? Do we use np.int in Sentence Transformers or is it used in a third party library?

The issue is actually in combination of transfomers==4.20 (here np.int is used) and the latest sentence-transformers.

fkdosilovic:
Perhaps instead of preventing users from using numpy 1.24, we can follow the deprecation guidelines and update np.int with int? Do we use np.int in Sentence Transformers or is it used in a third party library?

The issue is actually in combination of transfomers==4.20 (here np.int is used) and the latest sentence-transformers. Will close PR for now.

PR #2506, opened 2024-02-26
[`loss`] Add AdaptiveLayerLoss; 2d Matryoshka loss modifiers

tomaarsen:
Hello!

## Pull Request overview
* Add `AdaptiveLayerLoss`, should be combinable with `MatryoshkaLoss` to get 2dMatryoshka loss.

## Details
This is still a work in progress. Here are some scripts to interact with it:

<details><summary>Training</summary>

```python
import math
from sentence_transformers import models, losses, datasets
from sentence_transformers import LoggingHandler, SentenceTransformer, util, InputExample
from sentence_transformers.evaluation import EmbeddingSimilarityEvaluator, SimilarityFunction
import logging
from datetime import datetime
import sys
import os
import gzip
import csv
import random

#### Just some code to print debug information to stdout
logging.basicConfig(
    format="%(asctime)s - %(message)s", datefmt="%Y-%m-%d %H:%M:%S", level=logging.INFO, handlers=[LoggingHandler()]
)
#### /print debug information to stdout

model_name = sys.argv[1] if len(sys.argv) > 1 else "distilroberta-base"
train_batch_size = 128  # The larger you select this, the better the results (usually). But it requires more GPU memory
max_seq_length = 75
num_epochs = 1

# Save path of the model
model_save_path = (
    "output/matryoshka_nli_" + model_name.replace("/", "-") + "-" + datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
)


# Here we define our SentenceTransformer model
word_embedding_model = models.Transformer(model_name, max_seq_length=max_seq_length)
pooling_model = models.Pooling(word_embedding_model.get_word_embedding_dimension(), pooling_mode="mean")
model = SentenceTransformer(modules=[word_embedding_model, pooling_model])

# Check if dataset exists. If not, download and extract  it
nli_dataset_path = "data/AllNLI.tsv.gz"

if not os.path.exists(nli_dataset_path):
    util.http_get("https://sbert.net/datasets/AllNLI.tsv.gz", nli_dataset_path)

# Read the AllNLI.tsv.gz file and create the training dataset
logging.info("Read AllNLI train dataset")


def add_to_samples(sent1, sent2, label):
    if sent1 not in train_data:
        train_data[sent1] = {"contradiction": set(), "entailment": set(), "neutral": set()}
    train_data[sent1][label].add(sent2)


train_data = {}
with gzip.open(nli_dataset_path, "rt", encoding="utf8") as fIn:
    reader = csv.DictReader(fIn, delimiter="\t", quoting=csv.QUOTE_NONE)
    for row in reader:
        if row["split"] == "train":
            sent1 = row["sentence1"].strip()
            sent2 = row["sentence2"].strip()

            add_to_samples(sent1, sent2, row["label"])
            add_to_samples(sent2, sent1, row["label"])  # Also add the opposite


train_samples = []
for sent1, others in train_data.items():
    if len(others["entailment"]) > 0 and len(others["contradiction"]) > 0:
        train_samples.append(
            InputExample(
                texts=[sent1, random.choice(list(others["entailment"])), random.choice(list(others["contradiction"]))]
            )
        )
        train_samples.append(
            InputExample(
                texts=[random.choice(list(others["entailment"])), sent1, random.choice(list(others["contradiction"]))]
            )
        )

logging.info("Train samples: {}".format(len(train_samples)))


# Special data loader that avoid duplicates within a batch
train_dataloader = datasets.NoDuplicatesDataLoader(train_samples, batch_size=train_batch_size)


# Our training loss
train_loss = losses.MultipleNegativesRankingLoss(model)
train_loss = losses.AdaptiveLayerLoss(model, train_loss)

stsb_dev = load_dataset("mteb/stsbenchmark-sts", split="validation")
dev_evaluator = EmbeddingSimilarityEvaluator(
    stsb_dev["sentence1"],
    stsb_dev["sentence2"],
    [score / 5 for score in stsb_dev["score"]],
    main_similarity=SimilarityFunction.COSINE,
    name="sts-dev",
)

# Configure the training
warmup_steps = math.ceil(len(train_dataloader) * num_epochs * 0.1)  # 10% of train data for warm-up
logging.info("Warmup-steps: {}".format(warmup_steps))


# Train the model
model.fit(
    train_objectives=[(train_dataloader, train_loss)],
    evaluator=dev_evaluator,
    epochs=num_epochs,
    evaluation_steps=int(len(train_dataloader) * 0.1),
    warmup_steps=warmup_steps,
    output_path=model_save_path,
    use_amp=False,  # Set to True, if your GPU supports FP16 operations
)


##############################################################################
#
# Load the stored model and evaluate its performance on STS benchmark dataset
#
##############################################################################


model = SentenceTransformer(model_save_path)
stsb_test = load_dataset("mteb/stsbenchmark-sts", split="test")
test_evaluator = EmbeddingSimilarityEvaluator(
    stsb_test["sentence1"],
    stsb_test["sentence2"],
    [score / 5 for score in stsb_test["score"]],
    main_similarity=SimilarityFunction.COSINE,
    name="sts-test",
)
test_evaluator(model, output_path=model_save_path)
```
</details>

<details><summary>Evaluation</summary>

```python
import time
import datasets
    SentenceTransformer,
    evaluation,
)

stsb = datasets.load_dataset("mteb/stsbenchmark-sts", split="test")

# This is the model without KV Div
# model = SentenceTransformer(r"output\matryoshka_nli_distilroberta-base-2024-02-26_10-46-35", device="cpu")
# With KV Div, weighted
# model = SentenceTransformer(r"output/matryoshka_nli_distilroberta-base-2024-02-26_12-03-42", device="cpu")
# Sampled, no weighting, with KV Div
# model = SentenceTransformer(r"output/matryoshka_nli_distilroberta-base-2024-02-26_12-41-19", device="cpu")
# Weighted, without KV div:
# model = SentenceTransformer(r"output/matryoshka_nli_distilroberta-base-2024-02-26_13-13-13")


# model = SentenceTransformer(r"tomaarsen/distilroberta-base-nli")
model = SentenceTransformer("tomaarsen/distilroberta-base-nli-matryoshka")
num_layers = model[0].auto_model.config.num_hidden_layers
for layer_idx in range(num_layers, 0, -1):
    model[0].auto_model.encoder.layer = model[0].auto_model.encoder.layer[:layer_idx]

    evaluator = evaluation.EmbeddingSimilarityEvaluator(
        stsb["sentence1"],
        stsb["sentence2"],
        [score / 5 for score in stsb["score"]],
        main_similarity=evaluation.SimilarityFunction.COSINE,
        name="sts-test",
    )
    start_t = time.time()
    results = evaluator(model)
    print(layer_idx, "layers:", results, time.time() - start_t)
```

</details>

<details><summary>Plotting</summary>

```python
from matplotlib import pyplot as plt
import numpy as np


distilroberta_base_adaptive_layer_no_kl = [
0.8159721234001157,
0.8004217347737033,
0.7665517448440389,
0.7469348481953944,
0.7339609963379268,
0.7014780116387153,
][::-1]

distilroberta_base_adaptive_layer_kl_weighted = [
0.7785996943617404,
0.7540883628779422,
0.7232510516990992,
0.7073519925123614,
0.6972483521396592,
0.6564191142763336,
][::-1]

distilroberta_base_adaptive_layer_kl_sampled = [
0.7758190475492528,
0.747703404414628,
0.7157785866758525,
0.7027284876713799,
0.6942918906213894,
0.656001043025149,
][::-1]

distilroberta_base_adaptive_layer_no_kl_weighted = [
0.8251635096595564,
0.7981810872476284,
0.7612060742431745,
0.7384533253019896,
0.7260377437096901,
0.6900959027048263,
][::-1]

distilroberta_base = [
0.8306162415022439,
0.7458841350790978,
0.609244366558845,
0.593080478268359,
0.5655809184315158,
0.49214436288263186,
][::-1]

distilroberta_base_matryoshka = [
0.8437115837599188,
0.7649454136341505,
0.6395590513571396,
0.6057172017006697,
0.5734948744031261,
0.49558068577101144,
][::-1]

speeds = [
7.901154041290283,
6.610083341598511,
5.288370847702026,
4.027814626693726,
2.788794994354248,
1.5712966918945312,
][::-1]
speedups = [speeds[-1] / speed for speed in speeds]

X = np.arange(1, len(distilroberta_base_adaptive_layer_no_kl) + 1)

# fig, (ax_line, ax_hist, ax_speed) = plt.subplots(3, figsize=(6.4, 8.6))
fig, ax_line = plt.subplots()#figsize=(6.4, 8.6))
ax_line.plot(X, distilroberta_base_adaptive_layer_no_kl, label="distilroberta-base-nli-adaptive-layer-no-kl")
ax_line.plot(X, distilroberta_base_adaptive_layer_kl_weighted, label="distilroberta-base-nli-adaptive-layer-kl-weighted")
ax_line.plot(X, distilroberta_base_adaptive_layer_kl_sampled, label="distilroberta-base-nli-adaptive-layer-kl-sampled")
ax_line.plot(X, distilroberta_base_adaptive_layer_no_kl_weighted, label="distilroberta-base-nli-adaptive-layer-no-kl-weighted")
ax_line.plot(X, distilroberta_base_matryoshka, label="distilroberta-base-nli-matryoshka")
ax_line.plot(X, distilroberta_base, label="distilroberta-base-nli")

# ax.set_axisbelow(True)
ax_line.grid(linestyle="--")
ax_line.set_xticks(X)
# ax.yaxis.grid(color='gray', linestyle='dashed')
# ax_line.set_xlabel("Embedding dimension")
ax_line.set_ylabel("Spearman correlation")# (higher is better)")
ax_line.legend(loc="lower right")

"""
ax_hist.bar(X + 0.2, [score / distilroberta_base_adaptive_layer_no_kl[-1] for score in distilroberta_base_adaptive_layer_no_kl], label="tomaarsen/distilroberta-base-nli-adaptive-layer", width=0.4)
ax_hist.bar(X - 0.2, [score / distilroberta_base[-1] for score in distilroberta_base], label="tomaarsen/distilroberta-base-nli", width=0.4)

# ax_hist.set_ylim(0.95, 1)
ax_hist.grid(linestyle="--")
ax_hist.set_xticks(X)
# ax_hist.set_xlabel("# Layers")
ax_hist.set_ylabel("Ratio of maximum performance")
# ax_hist.legend(loc="lower right")

ax_speed.grid(linestyle="--")
bar = ax_speed.bar(X, speedups, width=0.4)
ax_speed.bar_label(bar)
ax_speed.set_xlabel("# Layers")
ax_speed.set_ylabel("Speedup ratio")
ax_speed.set_ylim(0, 6)
"""

fig.suptitle(
    "STSB test score for various number of layers,\nwith and without Adaptive Layer loss"
)
fig.tight_layout()

plt.show()
```

</details>

Here are my findings so far:
image

1. `distilroberta-base-nli-adaptive-layer-no-kl`: Sum of MNRL loss for each layer
2. `distilroberta-base-nli-adaptive-layer-kl-weighted`: MNRL loss for the final embedding + (sum of MNRL loss for each non-final layer / `num_layers` - 1) + (sum of KL-div for each non-final layer / (`num_layers` - 1)).
    This is kind of like "normal" loss at weight 1, loss at non-final layers at weight 1 and KL-div loss at non-final layers at weight 1.
3. `distilroberta-base-nli-adaptive-layer-kl-sampled`: MNRL loss for the final embedding + MNRL loss at a randomly sampled layer + KL-div at a randomly sampled layer.
4. `distilroberta-base-nli-adaptive-layer-no-kl-weighted`: MNRL loss for the final embedding + (sum of MNRL loss for each non-final layer / `num_layers` - 1)
5. `distilroberta-base-nli-matryoshka`: MatryoshkaLoss on top of normal MNRL, no Adaptive Layer loss
6. `distilroberta-base-nli`: Normal MNRL, no Adaptive Layer loss, no Matryoshka loss

The `AdaptiveLayerLoss` seems to work correctly in combination with `MatryoshkaLoss`, which is great. At a later stage I will also introduce a shorthand for the 2d Matryoshka.

cc: @SeanLee97 @juanluisdb @ir2718 

- Tom Aarsen


sentence_transformers/losses/AdaptiveLayerLoss.py
```diff
@@ -0,0 +1,108 @@
...
+            decorated_forward.set_layer_idx(layer_idx)
+            layer_loss = self.loss(sentence_features, labels) / (decorated_forward.num_layers - 1)
+            loss = loss + layer_loss
+
+            """
+            # and KL-divergence loss between the current layer and the final layer
+            # Note: we use "batchmean" reduction as that aligns with the mathematical definition
+            embeddings = decorated_forward.get_layer_embeddings()
+            kl_div_loss = F.kl_div(
+                F.log_softmax(embeddings, dim=-1), F.softmax(final_embeddings, dim=-1), reduction="batchmean"
```
tomaarsen, line 101:
```suggestion
                F.log_softmax(embeddings, dim=-1), F.softmax(final_embeddings, dim=-1), reduction="batchmean"
```
I think the issue is that I'm running KL-div on the embeddings directly as they occur in the model, rather than after pooling etc. has been applied. 
Normally when you do knowledge distillation, you do the KL-div on the embeddings after pooling, and this approach is essentially knowledge distillation with the last embedding layer being your teacher model.

tomaarsen: I think this was it! 
image
KL-div but after pooling is slightly better than no KL-div, just like in the paper! 🎉

sentence_transformers/losses/AdaptiveLayerLoss.py
```diff
@@ -0,0 +1,108 @@
...
+            ) / (decorated_forward.num_layers - 1)
+            loss = loss + kl_div_loss
+            """
```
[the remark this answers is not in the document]
tomaarsen: (Edit: This is now updated, see https://github.com/UKPLab/sentence-transformers/pull/2506#discussion_r1502719014)

sentence_transformers/losses/AdaptiveLayerLoss.py
SeanLee97: @tomaarsen Thanks for your excellent implementation! We used the pooling embeddings for KL, sorry for not pointing out this earlier.

However, I found that 2DMSE didn’t outperform MRL at the last layer. 
In my experiments, it consistently outperforms MRL at the last layer. I will try  the code tomorrow to see what caused this difference.

tomaarsen: > We used the pooling embeddings for KL, sorry for not pointing out this earlier.

No worries! I'm glad I figured it out quickly :)

> However, I found that 2DMSE didn’t outperform MRL at the last layer.
> In my experiments, it consistently outperforms MRL at the last layer.

Could you clarify this? This seems contradictory, unless I am understanding it wrong.

SeanLee97: In the first figure, I found MRL(the purple line) performs better than 2DMSE at the last layer. In my experiment, 2DMSE performed better than MRL at the last layer.

tomaarsen: Ahh, you're right. I think that might be because the purple line is the only model that uses the original MRL, the other models only use the "Adaptive Layer" part, not the MRL part. I just pushed some code to combine the two to create roughly 2DMSE (the only difference I believe is that my MatryoshkaLoss implementation uses all dimensions at once instead of sampling 1): 211eec32df0b06c49ec4bf850e4dcd78b0ff25c1

But I haven't tested with that one yet. I will do so soon & report back. Then hopefully the `Matryoshka2dLoss` will outperform `MatryoshkaLoss` at the last layer just like in your experiments.

tomaarsen: > Then hopefully the `Matryoshka2dLoss` will outperform `MatryoshkaLoss` at the last layer just like in your experiments.

No luck so far :/ 

`Matryoshka2dLoss` + `MNRL` at the final layer: 82.59
`Matryoshka2dLoss` + `MNRL` but with all layers at once (instead of sampling 1) at the final layer: 82.60
`AdaptiveLayerLoss` + `MNRL` at the final layer: 82.92
`MatryoshkaLoss` + `MNRL` at the final layer: 84.37
Just `MNRL` at the final layer: 83.06

So, MRL > 2DMSE at the final layer for me. `AdaptiveLayerLoss` + `MNRL` is about equivalent to just MNRL for me, which is nice. But `AdaptiveLayerLoss` + `MatryoshkaLoss` + `MNRL` (i.e. `Matryoshka2dLoss` + `MNRL`) is not equivalent to just `MatryoshkaLoss` + `MNRL` sadly.

I'll try more tests, this time with AnglELoss instead of MNRL. Perhaps that makes the difference

tomaarsen:
Here's another training script, this one trains a bit quicker:

<details><summary>STS Training script</summary>

```python
import math
from sentence_transformers import SentenceTransformer, LoggingHandler, losses, models, util
from sentence_transformers.evaluation import EmbeddingSimilarityEvaluator
from sentence_transformers.readers import InputExample
import logging
from datetime import datetime
import sys
import os
import gzip
import csv

#### Just some code to print debug information to stdout
logging.basicConfig(
    format="%(asctime)s - %(message)s", datefmt="%Y-%m-%d %H:%M:%S", level=logging.INFO, handlers=[LoggingHandler()]
)
#### /print debug information to stdout


# Check if dataset exists. If not, download and extract  it
sts_dataset_path = "datasets/stsbenchmark.tsv.gz"

if not os.path.exists(sts_dataset_path):
    util.http_get("https://sbert.net/datasets/stsbenchmark.tsv.gz", sts_dataset_path)


# You can specify any huggingface/transformers pre-trained model here, for example, bert-base-uncased, roberta-base, xlm-roberta-base
model_name = sys.argv[1] if len(sys.argv) > 1 else "distilbert-base-uncased"

# Read the dataset
train_batch_size = 16
num_epochs = 4
model_save_path = (
    "output/matryoshka_sts_" + model_name.replace("/", "-") + "-" + datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
)

# Use Huggingface/transformers model (like BERT, RoBERTa, XLNet, XLM-R) for mapping tokens to embeddings
word_embedding_model = models.Transformer(model_name)

# Apply mean pooling to get one fixed sized sentence vector
pooling_model = models.Pooling(
    word_embedding_model.get_word_embedding_dimension(),
    pooling_mode_mean_tokens=True,
    pooling_mode_cls_token=False,
    pooling_mode_max_tokens=False,
)

model = SentenceTransformer(modules=[word_embedding_model, pooling_model])

# Convert the dataset to a DataLoader ready for training
logging.info("Read STSbenchmark train dataset")

train_samples = []
dev_samples = []
test_samples = []
with gzip.open(sts_dataset_path, "rt", encoding="utf8") as fIn:
    reader = csv.DictReader(fIn, delimiter="\t", quoting=csv.QUOTE_NONE)
    for row in reader:
        score = float(row["score"]) / 5.0  # Normalize score to range 0 ... 1
        inp_example = InputExample(texts=[row["sentence1"], row["sentence2"]], label=score)

        if row["split"] == "dev":
            dev_samples.append(inp_example)
        elif row["split"] == "test":
            test_samples.append(inp_example)
        else:
            train_samples.append(inp_example)


train_dataloader = DataLoader(train_samples, shuffle=True, batch_size=train_batch_size)
# train_loss = losses.CoSENTLoss(model=model)
train_loss = losses.AnglELoss(model=model)
train_loss = losses.Matryoshka2dLoss(model, train_loss, [768, 512, 256, 128, 64])


logging.info("Read STSbenchmark dev dataset")
evaluator = EmbeddingSimilarityEvaluator.from_input_examples(dev_samples, name="sts-dev")


# Configure the training. We skip evaluation in this example
warmup_steps = math.ceil(len(train_dataloader) * num_epochs * 0.1)  # 10% of train data for warm-up
logging.info("Warmup-steps: {}".format(warmup_steps))


# Train the model
model.fit(
    train_objectives=[(train_dataloader, train_loss)],
    evaluator=evaluator,
    epochs=num_epochs,
    evaluation_steps=1000,
    warmup_steps=warmup_steps,
    output_path=model_save_path,
)


##############################################################################
#
# Load the stored model and evaluate its performance on STS benchmark dataset
#
##############################################################################

model = SentenceTransformer(model_save_path)
test_evaluator = EmbeddingSimilarityEvaluator.from_input_examples(test_samples, name="sts-test")
test_evaluator(model, output_path=model_save_path)
```

</details>

- Tom Aarsen

sentence_transformers/losses/AdaptiveLayerLoss.py
tomaarsen: image

This is with a version where I sample 1 dimension and 1 layer at a time, just like the real 2DMSE implementation (yellow). Already seems a bit better.

SeanLee97: Hi, I follow AnglE's implementation to add `temperature` to KL. Additionally, I set a random seed `42` for all models and used `CLS` pooling following the paper.

Here are the results:

- MatryoshkaLoss + CoSENT: 84.62
- AdaptiveLayerLoss (n_layers_per_step=1, temperature=1.0) + CoSENT: **84.71** 
- AdaptiveLayerLoss (n_layers_per_step=-1, temperature=1.0) + CoSENT: 84.58
- Matryoshka2dLoss (n_layers_per_step=1, temperature=0.3) + CoSENT: 84.53
- Matryoshka2dLoss (n_layers_per_step=-1, temperature=50.0) + CoSENT: 84.36

AdaptiveLayerLoss outperforms others, but Matryoshka2dLoss is still not as good as expected. I am still checking the differences between the two implementations.

 
<details>
<summary>train_sts.py</summary>

```python
"""
This examples trains BERT (or any other transformer model like RoBERTa, DistilBERT etc.) for the STSbenchmark from scratch. It generates sentence embeddings
that can be compared using cosine-similarity to measure the similarity.

Usage:
python training_nli.py

OR
python training_nli.py pretrained_transformer_model_name
"""
import math
from sentence_transformers import SentenceTransformer, LoggingHandler, losses, models, util
from sentence_transformers.evaluation import EmbeddingSimilarityEvaluator
from sentence_transformers.readers import InputExample
import logging
from datetime import datetime
import sys
import os
import gzip
import csv
import torch
import numpy as np
import random

random_seed = 42
if random_seed is not None:
    os.environ['PYTHONHASHSEED'] = str(random_seed)
    random.seed(random_seed)
    np.random.seed(random_seed)
    torch.manual_seed(random_seed)

#### Just some code to print debug information to stdout
logging.basicConfig(
    format="%(asctime)s - %(message)s", datefmt="%Y-%m-%d %H:%M:%S", level=logging.INFO, handlers=[LoggingHandler()]
)
#### /print debug information to stdout


# Check if dataset exists. If not, download and extract  it
sts_dataset_path = "datasets/stsbenchmark.tsv.gz"

if not os.path.exists(sts_dataset_path):
    util.http_get("https://sbert.net/datasets/stsbenchmark.tsv.gz", sts_dataset_path)


# You can specify any huggingface/transformers pre-trained model here, for example, bert-base-uncased, roberta-base, xlm-roberta-base
model_name = sys.argv[1] if len(sys.argv) > 1 else "distilbert-base-uncased"

# Read the dataset
train_batch_size = 16
num_epochs = 4
model_save_path = (
    "output/training_stsbenchmark_" + model_name.replace("/", "-") + "-" + datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
)

# Use Huggingface/transformers model (like BERT, RoBERTa, XLNet, XLM-R) for mapping tokens to embeddings
word_embedding_model = models.Transformer(model_name)

# Apply mean pooling to get one fixed sized sentence vector
pooling_model = models.Pooling(
    word_embedding_model.get_word_embedding_dimension(),
    pooling_mode_mean_tokens=False,
    pooling_mode_cls_token=True,
    pooling_mode_max_tokens=False,
)

model = SentenceTransformer(modules=[word_embedding_model, pooling_model])

# Convert the dataset to a DataLoader ready for training
logging.info("Read STSbenchmark train dataset")

train_samples = []
dev_samples = []
test_samples = []
with gzip.open(sts_dataset_path, "rt", encoding="utf8") as fIn:
    reader = csv.DictReader(fIn, delimiter="\t", quoting=csv.QUOTE_NONE)
    for row in reader:
        score = float(row["score"]) / 5.0  # Normalize score to range 0 ... 1
        inp_example = InputExample(texts=[row["sentence1"], row["sentence2"]], label=score)

        if row["split"] == "dev":
            dev_samples.append(inp_example)
        elif row["split"] == "test":
            test_samples.append(inp_example)
        else:
            train_samples.append(inp_example)


train_dataloader = DataLoader(train_samples, shuffle=True, batch_size=train_batch_size)
train_loss = losses.CoSENTLoss(model)
# MatryoshkaLoss + CoSENT: 84.62
# AdaptiveLayerLoss (n_layers_per_step=1, temperature=1.0) + CoSENT: **84.71** 
# AdaptiveLayerLoss (n_layers_per_step=-1, temperature=1.0) + CoSENT: 84.58
# Matryoshka2dLoss (n_layers_per_step=1, temperature=0.3) + CoSENT: 84.53
# Matryoshka2dLoss (n_layers_per_step=-1, temperature=50.0) + CoSENT: 84.36
hidden_sizes = [768, 512, 256, 128, 64, 32, 16, 8]
# train_loss = losses.MatryoshkaLoss(model, train_loss, hidden_sizes)
# train_loss = losses.AdaptiveLayerLoss(model, train_loss, n_layers_per_step=1)
train_loss = losses.Matryoshka2dLoss(model, train_loss, hidden_sizes, n_layers_per_step=-1, n_dims_per_step=1, temperature=0.3)
print('train loss=', train_loss)

logging.info("Read STSbenchmark dev dataset")
evaluator = EmbeddingSimilarityEvaluator.from_input_examples(dev_samples, name="sts-dev") # dev_samples


# Configure the training. We skip evaluation in this example
warmup_steps = math.ceil(len(train_dataloader) * num_epochs * 0.1)  # 10% of train data for warm-up
logging.info("Warmup-steps: {}".format(warmup_steps))


# Train the model
model.fit(
    train_objectives=[(train_dataloader, train_loss)],
    evaluator=evaluator,
    epochs=num_epochs,
    evaluation_steps=1000,
    warmup_steps=warmup_steps,
    output_path=model_save_path,
)


##############################################################################
#
# Load the stored model and evaluate its performance on STS benchmark dataset
#
##############################################################################

model = SentenceTransformer(model_save_path)
test_evaluator = EmbeddingSimilarityEvaluator.from_input_examples(test_samples, name="sts-test")
test_evaluator(model, output_path=model_save_path)
```
</details>

<details>
<summary>Matryoshka2dLoss.py</summary>

```python
from typing import List, Optional, Union
from torch.nn import Module
from sentence_transformers.losses import AdaptiveLayerLoss, MatryoshkaLoss


class Matryoshka2dLoss(AdaptiveLayerLoss):
    def __init__(
        self,
        model: SentenceTransformer,
        loss: Module,
        matryoshka_dims: List[int],
        matryoshka_weights: Optional[List[Union[float, int]]] = None,
        n_layers_per_step: int = 1,
        n_dims_per_step: int = 1,
        temperature: float = 1.0,
    ) -> None:
        # Note, this uses n_layers_per_step=1 & n_dims_per_step=1 as default, following the 2DMSE implementation
        matryoshka_loss = MatryoshkaLoss(
            model, loss, matryoshka_dims, matryoshka_weights=matryoshka_weights, n_dims_per_step=n_dims_per_step
        )
        super().__init__(model, matryoshka_loss, n_layers_per_step, temperature=temperature)
```
</details>

<details>
<summary>AdaptiveLayerLoss.py</summary>

```python
import random
from typing import Dict, Iterable, List, Tuple
import warnings
from torch import Tensor, nn
from torch.nn import functional as F
import torch
from sentence_transformers.losses.CachedMultipleNegativesRankingLoss import CachedMultipleNegativesRankingLoss
from sentence_transformers.models import Transformer


class TransformerDecorator:
    """
    Decorator that caches the embeddings of all layers of the transformer.
    When `layer_idx` is set, it returns the cached embeddings of that layer instead.

    This is meant to override the forward function of the Transformer.
    """

    def __init__(self, transformer: Transformer, original_forward):
        self.transformer = transformer
        self.original_forward = original_forward
        self.embeddings: List[Tuple[Tensor]] = []
        self.last_embeddings: List[Tensor] = []
        self.features: List[Dict[str, Tensor]] = []
        self.layer_idx = None
        self.call_idx = 0

    def set_layer_idx(self, layer_idx):
        self.layer_idx = layer_idx
        self.call_idx = 0

    def get_layer_embeddings(self):
        return torch.concat([embedding[self.layer_idx] for embedding in self.embeddings], dim=1)

    def __call__(self, features):
        if self.layer_idx is None:
            output = self.call_grow_cache(features)
        else:
            output = self.call_use_cache(features)
            self.call_idx += 1
        return output

    def call_grow_cache(self, features):
        """
        Temporarily sets the output_hidden_states to True, runs the model, and then restores the original setting.
        Use the all_layer_embeddings to get the embeddings of all layers.
        """
        original_output_hidden_states = self.transformer.auto_model.config.output_hidden_states
        self.transformer.auto_model.config.output_hidden_states = True

        output = self.original_forward(features)
        # We ignore the first layer, as it is the input embeddings
        # and the last layer, as we already computed the loss over it
        self.num_layers = len(output["all_layer_embeddings"]) - 1
        self.embeddings.append(output["all_layer_embeddings"][1:-1])
        self.last_embeddings.append(output["token_embeddings"])
        self.features.append(
            {key: value for key, value in output.items() if key not in ["all_layer_embeddings", "token_embeddings"]}
        )

        # Restore original setting
        self.transformer.auto_model.config.output_hidden_states = original_output_hidden_states

        if original_output_hidden_states:
            del output["all_layer_embeddings"]

        return output

    def call_use_cache(self, features):
        return {**self.features[self.call_idx], "token_embeddings": self.embeddings[self.call_idx][self.layer_idx]}


class ForwardDecorator:
    """
    Decorator that caches the embeddings after all modules (e.g. pooling) of the model.
    Required to get the embeddings after all modules for the KL-divergence loss.

    This is meant to override the forward function of the SentenceTransformer.
    """

    def __init__(self, fn):
        self.fn = fn
        self.embeddings = []

    def __call__(self, features):
        output = self.fn(features)
        self.embeddings.append(output["sentence_embedding"])
        return output

    def get_embeddings(self):
        embeddings = torch.concat(self.embeddings, dim=0)
        self.embeddings = []
        return embeddings


class AdaptiveLayerLoss(nn.Module):
    def __init__(
        self,
        model: SentenceTransformer,
        loss: nn.Module,
        n_layers_per_step: int = -1,
        temperature: float = 1.0,
    ) -> None:
        """
        The AdaptiveLayerLoss can be seen as a loss *modifier* that allows you to use other loss functions at non-final
        layers of the Sentence Transformer model. This is useful for when you want to train a model where users have
        the option to lower the number of layers used to improve their inference speed and memory usage.

        :param model: SentenceTransformer model
        :param loss: The loss function to be used, e.g. :class:`MultipleNegativesRankingLoss`, :class:`CoSENTLoss`, etc.
        :param n_layers_per_step: The number of layers to use per step. If -1, then all layers are used. If > 0, then
            a random sample of n_layers_per_step layers are used per step. The 2DMSE paper uses `n_layers_per_step=1`.
            The default value is -1.

        References:
            - The concept was inspired by the 2DMSE paper: https://arxiv.org/abs/2402.14776

        Requirements:
            1. The base loss cannot be :class:`CachedMultipleNegativesRankingLoss`.

        Input:
            +---------------------------------------+--------+
            | Texts                                 | Labels |
            +=======================================+========+
            | any                                   | any    |
            +---------------------------------------+--------+

        Example:
            ::

                from sentence_transformers import SentenceTransformer, losses, InputExample

                model = SentenceTransformer('microsoft/mpnet-base')
                train_examples = [
                    InputExample(texts=['Anchor 1', 'Positive 1']),
                    InputExample(texts=['Anchor 2', 'Positive 2']),
                ]
                train_dataloader = DataLoader(train_examples, shuffle=True, batch_size=32)
                train_loss = losses.MultipleNegativesRankingLoss(model=model)
                train_loss = losses.AdaptiveLayerLoss(model, train_loss)
                model.fit(
                    [(train_dataloader, train_loss)],
                    epochs=10,
                )
        """
        super().__init__()
        self.model = model
        self.loss = loss
        self.n_layers_per_step = n_layers_per_step
        self.temperature = temperature
        assert isinstance(self.model[0], Transformer)
        if isinstance(loss, CachedMultipleNegativesRankingLoss):
            warnings.warn("MatryoshkaLoss is not compatible with CachedMultipleNegativesRankingLoss.", stacklevel=2)

    def forward(self, sentence_features: Iterable[Dict[str, Tensor]], labels: Tensor) -> Tensor:
        # Decorate the forward function of the transformer to cache the embeddings of all layers
        original_transformer_forward = self.model[0].forward
        transformer_decorator = TransformerDecorator(self.model[0], original_transformer_forward)
        self.model[0].forward = transformer_decorator

        # Decorate the forward function of the model to get the embeddings after all modules (e.g. pooling)
        original_forward = self.model.forward
        forward_decorator = ForwardDecorator(original_forward)
        self.model.forward = forward_decorator

        # Run the loss normally: i.e. the final layer, but 1) use the transformers decorator to cache
        # the embeddings of all layers and 2) use the forward decorator to get the embeddings after all modules
        # for the KL-divergence loss
        loss = self.loss(sentence_features, labels)
        final_embeddings = forward_decorator.get_embeddings()
        final_embeddings = F.softmax(final_embeddings / self.temperature, dim=-1)

        num_layers = transformer_decorator.num_layers
        layer_indices = range(num_layers - 1)
        if self.n_layers_per_step > 0 and self.n_layers_per_step < num_layers - 1:
            layer_indices = random.sample(layer_indices, self.n_layers_per_step)

        # This loop is over `num_layer - 1` layers because we already computed the loss over the final layer
        for layer_idx in layer_indices:
            # Add regular loss for each layer by using the cached embeddings of that layer
            transformer_decorator.set_layer_idx(layer_idx)
            layer_loss = self.loss(sentence_features, labels)
            loss = loss + layer_loss / len(layer_indices)

            # and KL-divergence loss between the current layer and the final layer
            # Note: we use "batchmean" reduction as that aligns with the mathematical definition
            embeddings = forward_decorator.get_embeddings()
            kl_div_loss = F.kl_div(F.log_softmax(embeddings / self.temperature, dim=-1),
                                   final_embeddings, reduction="batchmean")
            loss = loss + kl_div_loss * self.temperature / len(layer_indices)

        self.model[0].forward = original_transformer_forward
        self.model.forward = original_forward

        return loss
```
</detail>

tomaarsen: Thanks a bunch for these experiments! What temperature did you use in the paper? 
Also, 
```python
loss = loss + kl_div_loss * self.temperature / len(layer_indices)
```
Should this be 
```python
loss = loss + kl_div_loss * self.temperature**2 / len(layer_indices)
```
perhaps? I tried the latter, but it performed a bit worse (in my few experiments).

Also, I realised another difference. My `MatryoshkaLoss` with `self.n_dims_per_step=1` just samples 1 dimension, rather than 1 dimension *plus* the largest dimension.

tomaarsen:
This is about as good as I can get it:
image

It ends up performing a bit worse than just Matryoshka or not using any "loss modifiers", e.g. 85 instead of 86 on mpnet-base. Beyond that it's a bit worse, this is the performance that you're giving for these speedups (in my tests):

```
Using 12 layers: 1.00x speedup, 0.00% performance loss
Using 11 layers: 1.10x speedup, 1.32% performance loss
Using 10 layers: 1.24x speedup, 4.36% performance loss
Using 9 layers: 1.41x speedup, 6.05% performance loss
Using 8 layers: 1.55x speedup, 6.17% performance loss
Using 7 layers: 1.70x speedup, 8.21% performance loss
Using 6 layers: 1.91x speedup, 8.02% performance loss
Using 5 layers: 2.13x speedup, 9.65% performance loss
Using 4 layers: 2.61x speedup, 11.23% performance loss
Using 3 layers: 3.06x speedup, 12.66% performance loss
Using 2 layers: 3.96x speedup, 14.79% performance loss
Using 1 layers: 5.50x speedup, 17.67% performance loss
```
It seems a bit hard to justify here. e.g. 8-9% performance (85.2 -> 78) to get a 2x increase in speed is pretty rough. 

cc @SeanLee97

sentence_transformers/losses/AdaptiveLayerLoss.py
SeanLee97: I've updated the implementation of AdaptiveLayerLoss.py, here are the new results:
Now, 2dmse outperforms MRL using the same hyperparameters.

**random seed 42**
- MatryoshkaLoss + CoSENT: 84.62
- AdaptiveLayerLoss (n_layers_per_step=1, temperature=1.0) + CoSENT: 85.17
- AdaptiveLayerLoss (n_layers_per_step=-1, temperature=1.0) + CoSENT: **85.36**
- Matryoshka2dLoss (n_layers_per_step=1, temperature=0.3) + CoSENT: 84.73
- Matryoshka2dLoss (n_layers_per_step=-1, temperature=0.3) + CoSENT: 84.11

<details>
<summary>AdaptiveLayerLoss.py</summary>

Main Diffs:

1)
```python
loss = loss + layer_loss / len(layer_indices)
```
-> 
```python
loss = loss + layer_loss / (1 + layer_idx) / len(layer_indices)
```

2)
```python
loss = loss + kl_div_loss * self.temperature / len(layer_indices)
```
->
```python
loss = loss + kl_div_loss * self.temperature
```

```python
import math
import random
from typing import Dict, Iterable, List, Tuple
import warnings
from torch import Tensor, nn
from torch.nn import functional as F
import torch
from sentence_transformers.losses.CachedMultipleNegativesRankingLoss import CachedMultipleNegativesRankingLoss
from sentence_transformers.models import Transformer


class TransformerDecorator:
    """
    Decorator that caches the embeddings of all layers of the transformer.
    When `layer_idx` is set, it returns the cached embeddings of that layer instead.

    This is meant to override the forward function of the Transformer.
    """

    def __init__(self, transformer: Transformer, original_forward):
        self.transformer = transformer
        self.original_forward = original_forward
        self.embeddings: List[Tuple[Tensor]] = []
        self.last_embeddings: List[Tensor] = []
        self.features: List[Dict[str, Tensor]] = []
        self.layer_idx = None
        self.call_idx = 0

    def set_layer_idx(self, layer_idx):
        self.layer_idx = layer_idx
        self.call_idx = 0

    def get_layer_embeddings(self):
        return torch.concat([embedding[self.layer_idx] for embedding in self.embeddings], dim=1)

    def __call__(self, features):
        if self.layer_idx is None:
            output = self.call_grow_cache(features)
        else:
            output = self.call_use_cache(features)
            self.call_idx += 1
        return output

    def call_grow_cache(self, features):
        """
        Temporarily sets the output_hidden_states to True, runs the model, and then restores the original setting.
        Use the all_layer_embeddings to get the embeddings of all layers.
        """
        original_output_hidden_states = self.transformer.auto_model.config.output_hidden_states
        self.transformer.auto_model.config.output_hidden_states = True

        output = self.original_forward(features)
        # We ignore the first layer, as it is the input embeddings
        # and the last layer, as we already computed the loss over it
        self.num_layers = len(output["all_layer_embeddings"]) - 1
        self.embeddings.append(output["all_layer_embeddings"][1:-1])
        self.last_embeddings.append(output["token_embeddings"])
        self.features.append(
            {key: value for key, value in output.items() if key not in ["all_layer_embeddings", "token_embeddings"]}
        )

        # Restore original setting
        self.transformer.auto_model.config.output_hidden_states = original_output_hidden_states

        if original_output_hidden_states:
            del output["all_layer_embeddings"]

        return output

    def call_use_cache(self, features):
        return {**self.features[self.call_idx], "token_embeddings": self.embeddings[self.call_idx][self.layer_idx]}


class ForwardDecorator:
    """
    Decorator that caches the embeddings after all modules (e.g. pooling) of the model.
    Required to get the embeddings after all modules for the KL-divergence loss.

    This is meant to override the forward function of the SentenceTransformer.
    """

    def __init__(self, fn):
        self.fn = fn
        self.embeddings = []

    def __call__(self, features):
        output = self.fn(features)
        self.embeddings.append(output["sentence_embedding"])
        return output

    def get_embeddings(self):
        embeddings = torch.concat(self.embeddings, dim=0)
        self.embeddings = []
        return embeddings


class AdaptiveLayerLoss(nn.Module):
    def __init__(
        self,
        model: SentenceTransformer,
        loss: nn.Module,
        n_layers_per_step: int = -1,
        temperature: float = 1.0,
    ) -> None:
        """
        The AdaptiveLayerLoss can be seen as a loss *modifier* that allows you to use other loss functions at non-final
        layers of the Sentence Transformer model. This is useful for when you want to train a model where users have
        the option to lower the number of layers used to improve their inference speed and memory usage.

        :param model: SentenceTransformer model
        :param loss: The loss function to be used, e.g. :class:`MultipleNegativesRankingLoss`, :class:`CoSENTLoss`, etc.
        :param n_layers_per_step: The number of layers to use per step. If -1, then all layers are used. If > 0, then
            a random sample of n_layers_per_step layers are used per step. The 2DMSE paper uses `n_layers_per_step=1`.
            The default value is -1.

        References:
            - The concept was inspired by the 2DMSE paper: https://arxiv.org/abs/2402.14776

        Requirements:
            1. The base loss cannot be :class:`CachedMultipleNegativesRankingLoss`.

        Input:
            +---------------------------------------+--------+
            | Texts                                 | Labels |
            +=======================================+========+
            | any                                   | any    |
            +---------------------------------------+--------+

        Example:
            ::

                from sentence_transformers import SentenceTransformer, losses, InputExample

                model = SentenceTransformer('microsoft/mpnet-base')
                train_examples = [
                    InputExample(texts=['Anchor 1', 'Positive 1']),
                    InputExample(texts=['Anchor 2', 'Positive 2']),
                ]
                train_dataloader = DataLoader(train_examples, shuffle=True, batch_size=32)
                train_loss = losses.MultipleNegativesRankingLoss(model=model)
                train_loss = losses.AdaptiveLayerLoss(model, train_loss)
                model.fit(
                    [(train_dataloader, train_loss)],
                    epochs=10,
                )
        """
        super().__init__()
        self.model = model
        self.loss = loss
        self.n_layers_per_step = n_layers_per_step
        self.temperature = temperature
        print('n_layers_per_step>>>', n_layers_per_step)
        print('temperature2>>>', temperature)
        assert isinstance(self.model[0], Transformer)
        if isinstance(loss, CachedMultipleNegativesRankingLoss):
            warnings.warn("MatryoshkaLoss is not compatible with CachedMultipleNegativesRankingLoss.", stacklevel=2)

    def forward(self, sentence_features: Iterable[Dict[str, Tensor]], labels: Tensor) -> Tensor:
        # Decorate the forward function of the transformer to cache the embeddings of all layers
        original_transformer_forward = self.model[0].forward
        transformer_decorator = TransformerDecorator(self.model[0], original_transformer_forward)
        self.model[0].forward = transformer_decorator

        # Decorate the forward function of the model to get the embeddings after all modules (e.g. pooling)
        original_forward = self.model.forward
        forward_decorator = ForwardDecorator(original_forward)
        self.model.forward = forward_decorator

        # Run the loss normally: i.e. the final layer, but 1) use the transformers decorator to cache
        # the embeddings of all layers and 2) use the forward decorator to get the embeddings after all modules
        # for the KL-divergence loss
        loss = self.loss(sentence_features, labels)
        final_embeddings = forward_decorator.get_embeddings()
        final_embeddings = F.softmax(final_embeddings / self.temperature, dim=-1)

        num_layers = transformer_decorator.num_layers
        layer_indices = range(num_layers - 1)
        if self.n_layers_per_step > 0 and self.n_layers_per_step < num_layers - 1:
            layer_indices = random.sample(layer_indices, self.n_layers_per_step)

        # This loop is over `num_layer - 1` layers because we already computed the loss over the final layer
        for layer_idx in layer_indices:
            # Add regular loss for each layer by using the cached embeddings of that layer
            transformer_decorator.set_layer_idx(layer_idx)
            layer_loss = self.loss(sentence_features, labels)
            loss = loss + layer_loss / (1 + layer_idx) / len(layer_indices)

            # and KL-divergence loss between the current layer and the final layer
            # Note: we use "batchmean" reduction as that aligns with the mathematical definition
            embeddings = forward_decorator.get_embeddings()
            kl_div_loss = F.kl_div(F.log_softmax(embeddings / self.temperature, dim=-1),
                                   final_embeddings, reduction="batchmean")
            loss = loss + kl_div_loss * self.temperature

        self.model[0].forward = original_transformer_forward
        self.model.forward = original_forward

        return loss

```
</details>

SeanLee97:
> This is about as good as I can get it: image
> 
> It ends up performing a bit worse than just Matryoshka or not using any "loss modifiers", e.g. 85 instead of 86 on mpnet-base. Beyond that it's a bit worse, this is the performance that you're giving for these speedups (in my tests):
> 
> ```
> Using 12 layers: 1.00x speedup, 0.00% performance loss
> Using 11 layers: 1.10x speedup, 1.32% performance loss
> Using 10 layers: 1.24x speedup, 4.36% performance loss
> Using 9 layers: 1.41x speedup, 6.05% performance loss
> Using 8 layers: 1.55x speedup, 6.17% performance loss
> Using 7 layers: 1.70x speedup, 8.21% performance loss
> Using 6 layers: 1.91x speedup, 8.02% performance loss
> Using 5 layers: 2.13x speedup, 9.65% performance loss
> Using 4 layers: 2.61x speedup, 11.23% performance loss
> Using 3 layers: 3.06x speedup, 12.66% performance loss
> Using 2 layers: 3.96x speedup, 14.79% performance loss
> Using 1 layers: 5.50x speedup, 17.67% performance loss
> ```
> 
> It seems a bit hard to justify here. e.g. 8-9% performance (85.2 -> 78) to get a 2x increase in speed is pretty rough.
> 
> cc @SeanLee97

What is the specific weight name of mpnet-base? I can help test 2dmse using the new implementation.

tomaarsen:
> @tomaarsen What is the specific weight name for mpnet-base? I can help test 2dmse using the new implementation.

The base model is `microsoft/mpnet-base`. That experiment was on my `nli` script which trains on AllNLI. I used a batch size of 64. But you can reproduce similar results for the `sts` script as well & that one takes muuch less time to train (~3 minutes for sts vs 1 hour for nli, on my GPU).

I appreciate you spending time on this! Your changes seem quite promising!

SeanLee97:
> > @tomaarsen What is the specific weight name for mpnet-base? I can help test 2dmse using the new implementation.
> 
> The base model is `microsoft/mpnet-base`. That experiment was on my `nli` script which trains on AllNLI. I used a batch size of 64. But you can reproduce similar results for the `sts` script as well & that one takes muuch less time to train (~3 minutes for sts vs 1 hour for nli, on my GPU).
> 
> I appreciate you spending time on this! Your changes seem quite promising!

@tomaarsen  Oh, i used a wrong weight `sentence-transformers/all-mpnet-base-v2`.

Here are the results based on `sentence-transformers/all-mpnet-base-v2`:

1) 2DMSE
```
12 layers: 0.8665234287587303 2.2667315006256104
11 layers: 0.8371866892149431 1.992100477218628
10 layers: 0.7940166764878576 2.1122915744781494
9 layers: 0.7697092807843348 2.1292121410369873
8 layers: 0.7752811280002465 1.884674310684204
7 layers: 0.7599688869043641 1.6190872192382812
6 layers: 0.7591389552408044 1.649423599243164
5 layers: 0.7480787788105059 1.6083667278289795
4 layers: 0.7388588961743835 1.2627253532409668
3 layers: 0.7265215061637512 1.035078763961792
2 layers: 0.7185180258078995 0.8012971878051758
1 layers: 0.7080982663777546 0.7450265884399414
``` 

2) MRL:

```
12 layers: 0.865970729513565 2.302177667617798
11 layers: 0.835379589789137 1.887479543685913
10 layers: 0.7807625077256249 2.0113484859466553
9 layers: 0.7345416694952731 1.8149290084838867
8 layers: 0.7195294038506093 1.6934611797332764
7 layers: 0.6784107978339916 1.7868788242340088
6 layers: 0.6784725097086001 1.5792937278747559
5 layers: 0.6624179433298056 1.264695167541504
4 layers: 0.6393914465785708 1.3678429126739502
3 layers: 0.6041716401520056 1.2712326049804688
2 layers: 0.6080091770272112 0.975938081741333
1 layers: 0.6544690667493208 0.6437351703643799
```

Now, 2dmse performs better than MRL.

I will try `microsoft/mpnet-base` later.

tomaarsen:
Which training script did you use here? STS? These results are quite good at a glance I think.

SeanLee97:
> Which training script did you use here? STS? These results are quite good at a glance I think.

<details>
<summary>train_nli.py</summary>

```python
import math
from sentence_transformers import models, losses, datasets
from sentence_transformers import LoggingHandler, SentenceTransformer, util, InputExample
from sentence_transformers.evaluation import EmbeddingSimilarityEvaluator, SimilarityFunction
import logging
from datetime import datetime
import sys
import os
import gzip
import csv
import random

#### Just some code to print debug information to stdout
logging.basicConfig(
    format="%(asctime)s - %(message)s", datefmt="%Y-%m-%d %H:%M:%S", level=logging.INFO, handlers=[LoggingHandler()]
)
#### /print debug information to stdout

model_name = sys.argv[1] if len(sys.argv) > 1 else "sentence-transformers/all-mpnet-base-v2" # "distilroberta-base"
train_batch_size = 128  # The larger you select this, the better the results (usually). But it requires more GPU memory
max_seq_length = 75
num_epochs = 1

# Save path of the model
model_save_path = (
    # "output/mrl_nli_" + model_name.replace("/", "-") + "-" + datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
    "output/2dmse_nli_" + model_name.replace("/", "-") + "-" + datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
)
print('>>> model_save_path', model_save_path)


# Here we define our SentenceTransformer model
word_embedding_model = models.Transformer(model_name, max_seq_length=max_seq_length)
pooling_model = models.Pooling(word_embedding_model.get_word_embedding_dimension(), pooling_mode="mean")  # mean
model = SentenceTransformer(modules=[word_embedding_model, pooling_model])

# Check if dataset exists. If not, download and extract  it
nli_dataset_path = "data/AllNLI.tsv.gz"

if not os.path.exists(nli_dataset_path):
    util.http_get("https://sbert.net/datasets/AllNLI.tsv.gz", nli_dataset_path)

# Read the AllNLI.tsv.gz file and create the training dataset
logging.info("Read AllNLI train dataset")


def add_to_samples(sent1, sent2, label):
    if sent1 not in train_data:
        train_data[sent1] = {"contradiction": set(), "entailment": set(), "neutral": set()}
    train_data[sent1][label].add(sent2)

train_data = {}
with gzip.open(nli_dataset_path, "rt", encoding="utf8") as fIn:
    reader = csv.DictReader(fIn, delimiter="\t", quoting=csv.QUOTE_NONE)
    for row in reader:
        if row["split"] == "train":
            sent1 = row["sentence1"].strip()
            sent2 = row["sentence2"].strip()

            add_to_samples(sent1, sent2, row["label"])
            add_to_samples(sent2, sent1, row["label"])  # Also add the opposite


train_samples = []
for sent1, others in train_data.items():
    if len(others["entailment"]) > 0 and len(others["contradiction"]) > 0:
        train_samples.append(
            InputExample(
                texts=[sent1, random.choice(list(others["entailment"])), random.choice(list(others["contradiction"]))]
            )
        )
        train_samples.append(
            InputExample(
                texts=[random.choice(list(others["entailment"])), sent1, random.choice(list(others["contradiction"]))]
            )
        )

logging.info("Train samples: {}".format(len(train_samples)))


# Special data loader that avoid duplicates within a batch
train_dataloader = datasets.NoDuplicatesDataLoader(train_samples, batch_size=train_batch_size)


# Our training loss
train_loss = losses.MultipleNegativesRankingLoss(model)
hidden_sizes = [768, 512, 256, 128, 64, 32, 16, 8]
train_loss = losses.Matryoshka2dLoss(model, train_loss, hidden_sizes, n_layers_per_step=1, n_dims_per_step=1, temperature=.3)
# train_loss = losses.MatryoshkaLoss(model, train_loss, hidden_sizes)
print('train loss=', train_loss)

stsb_dev = load_dataset("mteb/stsbenchmark-sts", split="test")  # validation
dev_evaluator = EmbeddingSimilarityEvaluator(
    stsb_dev["sentence1"],
    stsb_dev["sentence2"],
    [score / 5 for score in stsb_dev["score"]],
    main_similarity=SimilarityFunction.COSINE,
    name="sts-dev",
)

# Configure the training
warmup_steps = math.ceil(len(train_dataloader) * num_epochs * 0.1)  # 10% of train data for warm-up
logging.info("Warmup-steps: {}".format(warmup_steps))


# Train the model
model.fit(
    train_objectives=[(train_dataloader, train_loss)],
    evaluator=dev_evaluator,
    epochs=num_epochs,
    evaluation_steps=int(len(train_dataloader) * 0.1),
    warmup_steps=warmup_steps,
    output_path=model_save_path,
    save_best_model=True,
    use_amp=False,  # Set to True, if your GPU supports FP16 operations
)


##############################################################################
#
# Load the stored model and evaluate its performance on STS benchmark dataset
#
##############################################################################


model = SentenceTransformer(model_save_path)
stsb_test = load_dataset("mteb/stsbenchmark-sts", split="test")
test_evaluator = EmbeddingSimilarityEvaluator(
    stsb_test["sentence1"],
    stsb_test["sentence2"],
    [score / 5 for score in stsb_test["score"]],
    main_similarity=SimilarityFunction.COSINE,
    name="sts-test",
)
test_evaluator(model, output_path=model_save_path)
```
</details>

tomaarsen:
You can finetune `sentence-transformers/all-mpnet-base-v2` with a batch size of 128? Now that's a GPU rig that I'm jealous of 😄 
(FYI: Multi-GPU support is still coming! I think I'll make that my #1 focus again after 2dmse is done)

tomaarsen:
These are the updated figures using your modifications:
image
These are very satisfactory findings as far as I'm concerned. I would argue here that AdaptiveLayerLoss and 2DMSE do not negatively affect performance 🎉 

The performance loss at lower layers is a bit lower, but that is to be expected as I think the loss now focuses a bit more on the last layer loss. Perhaps we can expose weights? `last_layer_weight`, `prior_layers_weight`, `kl_div_weight` all set to 1.0 to allow users to tune this a bit?

Also, I am wondering if -1 is a good default for `AdaptiveLayerLoss` or not. It seems to sometimes outperform 1 and sometimes not. I suspect that 1 might be a bit easier to train (i.e. the backwards propagation seems easier as there's "less losses").

Once that's all sorted, then I can write a bit of documentation on this & release it in a few working days :)

- Tom Aarsen

Issue #2505, opened 2024-02-26
Inconistent pooling mode options


david-waterworth:
The validation requires `pooling mode in ["mean", "max", "cls", "weightedmean", "lasttoken"]`

https://github.com/UKPLab/sentence-transformers/blob/e0d93bb377eca9e0c94da436f3f35c336e9c109d/sentence_transformers/models/Pooling.py#L52

This means `self.pooling_mode_weightedmean_tokens` cannot be set via `pooling_mode`

https://github.com/UKPLab/sentence-transformers/blob/e0d93bb377eca9e0c94da436f3f35c336e9c109d/sentence_transformers/models/Pooling.py#L63

tomaarsen:
Hello!

> This means `self.pooling_mode_weightedmean_tokens` cannot be set via `pooling_mode`

I think you mean `pooling_mode_mean_sqrt_len_tokens`, and I think you're right. We need to add `mean_sqrt_len_tokens` to the pooling modes. Perhaps someone would like to pick this up in a PR, and otherwise I'll take care of it in a few days.

Thanks for reporting!

- Tom Aarsen

kishore-s-15:
Hi @tomaarsen, Can I work on this?

tomaarsen:
@kishore-s-15 I'm afraid that there's already an open pull request in #2507 which should solve this issue.

- Tom Aarsen

david-waterworth:
Note also I forgot to mention the docstring also doesn’t mention all the possible values for pooling modes.

tomaarsen:
Both should be fixed now @david-waterworth!

PR #2507, opened 2024-02-26
Add support for mean sqrt pooling.

fkdosilovic:
#### Related PRs/Issues

Closes https://github.com/UKPLab/sentence-transformers/issues/2505

#### Description

- Add support for setting mean sqrt as pooling option
- Wrong pooling mode now throws a ValueError instead of invalidating an assert.
- Minor documentation improvements

### DIFF
--- a/sentence_transformers/models/Pooling.py
+++ b/sentence_transformers/models/Pooling.py
@@ -21,8 +21,7 @@ class Pooling(nn.Module):
     :param pooling_mode_mean_tokens: Perform mean-pooling
     :param pooling_mode_mean_sqrt_len_tokens: Perform mean-pooling, but divide by sqrt(input_length).
     :param pooling_mode_weightedmean_tokens: Perform (position) weighted mean pooling. See `SGPT: GPT Sentence Embeddings for Semantic Search <https://arxiv.org/abs/2202.08904>`_.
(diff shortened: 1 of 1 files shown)


sentence_transformers/models/Pooling.py
```diff
@@ -22,6 +22,15 @@ class Pooling(nn.Module):
     :param pooling_mode_lasttoken: Perform last token pooling, see https://arxiv.org/abs/2202.08904 & https://arxiv.org/abs/2201.10005
     """
 
+    POOLING_MODES = (
+        "cls",
+        "lasttoken",
+        "max",
+        "mean",
+        "meansqrt",
```
tomaarsen, line 30: I think we should use `mean_sqrt_len_tokens`. The inconsistency with `_` in `mean_sqrt_len_tokens` versus no `_` in `lasttoken` and `weightedmean` is pretty bothersome, so normally `meansqrt` would be a lot better, but we should probably stick with what we already had for backwards compatibility:

https://github.com/UKPLab/sentence-transformers/blob/e0d93bb377eca9e0c94da436f3f35c336e9c109d/sentence_transformers/models/Pooling.py#L95-L96

fkdosilovic: Good catch. I missed this.

tomaarsen (approved):
Looks great!

tomaarsen:
Thanks a bunch for the recent help with PRs & issues, I appreciate it!

fkdosilovic:
> Thanks a bunch for the recent help with PRs & issues, I appreciate it!

Happy to help :)

PR #2510, opened 2024-02-28
Fix pooling ONNX export

fxmarty:
Currently, the exported ONNX model of `Pooling` with `pooling_mode_lasttoken` can not be consumed by ORT due to an `ArgMax` operator in int64, that ORT is not able to consume.

The issue can be reproduced with: `optimum-cli export onnx --model fxmarty/tiny-dummy-mistral-sentence-transformer mistral_st_onnx`

As reported in https://github.com/huggingface/optimum/issues/1731

See also https://github.com/microsoft/onnxruntime/issues/10068

@tomaarsen I could also patch the `Pooling` in Optimum, but its forward is rather big so I rather have it directly in sentence-transformers. Are you OK with that?

### DIFF
--- a/sentence_transformers/models/Pooling.py
+++ b/sentence_transformers/models/Pooling.py
@@ -184,7 +184,13 @@ class Pooling(nn.Module):
             # attention_mask shape: (bs, seq_len)
             # Get shape [bs] indices of the last token (i.e. the last token for each batch item)
             # Use flip and max() to get the last index of 1 in the attention mask
+
+            if torch.jit.is_tracing():
+                # Avoid tracing the argmax with int64 input that can not be handled by ONNX Runtime: https://github.com/microsoft/onnxruntime/issues/10068
+                attention_mask = attention_mask.to(torch.int32)
+
             values, indices = attention_mask.flip(1).max(1)
+
             indices = torch.where(values == 0, seq_len - 1, indices)
             gather_indices = seq_len - indices - 1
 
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Yes, I'm totally fine with having this on the Sentence Transformers side instead.
Related: I'd love to support convenient ONNX exporting, loading & inference within Sentence Transformers, ideally by relying on `optimum`. Perhaps related is #2426, which overhauls how models are loaded & saved using `transformers`.

- Tom Aarsen

fxmarty:
@tomaarsen Yes, I can confirm `optimum-cli export onnx --model Salesforce/SFR-Embedding-Mistral onnx/sfr-embedding-mistral` now works with this fix.

Regarding the ONNX export, this part is already supported in Optimum and could be leveraged. It is true we don't have implemented any `ORTModelForxxx` for sentence-transformers (apart from https://huggingface.co/docs/optimum/main/en/onnxruntime/package_reference/modeling_ort#optimum.onnxruntime.ORTModelForCustomTasks that should work), not sure how it is best to expose (same for TRT)

tomaarsen:
I am indeed aware that Sentence Tranformers ONNX exporting is already supported in Optimum, but it is currently under-documented on the my side & probably not as convenient to use as possible. I have been considering e.g. implementing `ORTSentenceTransformer` in `sentence-transformers` which allows for loading, exporting & inferencing. It still feels like there might be a missing link between the "normal" SentenceTransformer and the new class, though. Perhaps e.g. `to_onnx` or `from_sentence_transformer` methods could be good solutions there.

As for this PR, I'll merge it when the CI goes green.

- Tom Aarsen

fxmarty:
Yes that could be great! I'd suggest to make use of https://huggingface.co/docs/optimum/main/en/exporters/onnx/package_reference/export for the export part to avoid duplicate work.

tomaarsen:
That would be ideal indeed. I'll have to see where that fits on my long-term TODOs, that list is getting mighty long.

PR #2524, opened 2024-03-01
Adding `push_to_hub` to CrossEncoder

imvladikon:
Hi, 
I've noticed that many new cool features have recently been added, so first of all, thank you for your work!

#### Description
The great CrossEncoder class currently lacks support for auto-uploading to huggingface through the `push_to_hub` function(although it's achievable using the transformers models API directly)
This PR adds inheritance from `PushToHubMixin`, which is the standard mechanism for performing `push_to_hub` in the Transformers library. 

As a result, it's possible to do it:
```python

model = CrossEncoder("./mpnet-base-nli-matryoshka")
repo_id = "imvladikon/mpnet-base-nli-matryoshka"
model.push_to_hub(repo_id)
```

The resulting repo is here: https://huggingface.co/imvladikon/mpnet-base-nli-matryoshka where the README.md is generated automatically (which is the behavior that we have in the transformers library)

### DIFF
--- a/sentence_transformers/cross_encoder/CrossEncoder.py
+++ b/sentence_transformers/cross_encoder/CrossEncoder.py
@@ -1,3 +1,5 @@
+from functools import wraps
+
 from transformers import AutoModelForSequenceClassification, AutoTokenizer, AutoConfig
 import numpy as np
 import logging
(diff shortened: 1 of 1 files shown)


tomaarsen:
Thanks a bunch for setting this up! I expanded it a bit, e.g. making some of the kwargs explicit to help with IDE auto-suggestions and by always setting the cross-encoder tag on the generated README (although the README is still quite empty).

- Tom Aarsen

Issue #2538, opened 2024-03-12
Feature-Request: Passing arbitrary keyword arguments to submodules


Markus28:
### Problem
Increasingly, embedding models are becoming dynamically configurable during inference, for instance via:
- prompts (e.g. INSTRUCTOR)
- task-specific embeddings (e.g. Cohere embeddings)
- variable embedding dimension (e.g. Matryoshka embeddings)
- various other keyword arguments (e.g. `embed_instruction` in https://github.com/ContextualAI/gritlm)

Hence, it would be convenient if we could pass keyword arguments through `SentenceTransformer.encode` to the underlying embedding model to adjust its behaviour during inference. 
This interface should allow us to pass arbitrary keyword arguments to the model, s.t. something like this would become possible:
```python
model = SentenceTransformer('<MODEL>', trust_remote_code=True)
model.encode(['Hello world'], task='sts', embedding_dim=32, foobar=0)
```
### Proposed Solution
I propose that sentence transformer models can optionally specify in their `modules.json` file which modules expect which keyword arguments. I.e., a valid configuration could look like this:
```json
[
  {
    "idx": 0,
    "name": "0",
    "path": "",
    "type": "sentence_transformers.models.Transformer",
    "kwargs": ["task", "embedding_dim", "foobar"]
  },
  {
    "idx": 1,
    "name": "1",
    "path": "1_Pooling",
    "type": "sentence_transformers.models.Pooling"
  }
]
```
In the `SentenceTransformer.forward` method, we then distribute the keyword arguments to the respective submodules from `modules.json`. In particular, we also need to modify the `Transformer.forward` method to also accept arbitrary keyword arguments which are forwarded to `Transformer.auto_model`.

tomaarsen:
Closing, see https://github.com/huggingface/sentence-transformers/pull/2539#issuecomment-4418843764

PR #2539, opened 2024-03-12
Passing arbitrary keyword arguments to submodules

Markus28:
This PR implements a prototype of the proposal #2538. I.e., it allows arbitrary keyword arguments to be passed through `SentenceTransformer.encode` to the model.

Users may now specify in `modules.json` what keyword arguments are expected by which module. The `SentenceTransformer` class no longer inherits from `nn.Sequential`, but instead from `nn.ModuleDict`. We implement the `forward` method ourselves and distribute keyword arguments to the modules.

A model may then provide a `modules.json` file like this:

```json
[
  {
    "idx": 0,
    "name": "0",
    "path": "",
    "type": "sentence_transformers.models.Transformer",
    "kwargs": ["task", "embedding_dim", "foobar"]
  },
  {
    "idx": 1,
    "name": "1",
    "path": "1_Pooling",
    "type": "sentence_transformers.models.Pooling"
  }
]
```

and users may use the sentence-transformer model like this:

```python
model = SentenceTransformer('<MODEL>', trust_remote_code=True)
model.encode(['Hello world'], task='sts', embedding_dim=32, foobar=0)
```

### Todo
- These changes will be breaking if there is someone using the `.append` method, which was previously inherited from `nn.Sequential`. However, we could also implement it here
- It would make sense to also implement arbitrary kwargs for the tokenizer

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -268,9 +268,9 @@ class SentenceTransformer(nn.ModuleDict):
         required_keys = set()
         for keys in self._module_kwargs.values():
             required_keys = required_keys.union(set(keys))
-        if provided_keys != required_keys:
+        if not provided_keys.issubset(required_keys):
             raise ValueError(
-                f"Expected keyword arguments {required_keys if len(required_keys)>0 else '{}'} "
+                f"Expected keyword arguments to be subset of {required_keys if len(required_keys)>0 else '{}'} "
                 f"but got {provided_keys if len(provided_keys)>0 else '{}'}"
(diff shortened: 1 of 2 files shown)


bwanglzu:
@tomaarsen please let us know how do you think about this PR :) thanks

Bo

tomaarsen:
Hello!

Apologies for the delay. This has been integrated similarly, with `kwargs` added to `modules.json`, which can be passed to the model's `preprocess` (formerly `tokenize`) and each module's `forward`. Closing this as already implemented!

- Tom Aarsen

PR #2541, opened 2024-03-13
Add support for M2M100EncoderModel (aka NLLB, aka SONAR text encoder)

avidale:
This is a draft pull request for supporting SONAR text encoder model.
SONAR encoder is architecturally an encoder from NLLB, which is, in turn, is implemented as M2M100 in HF Transformers.
Thus, I support the encoder-only part of it, in the same way as T5-based encoders are currently supported.

This PR depends on a pull request to Transformers https://github.com/huggingface/transformers/pull/29646, and should be reviewed only after that one.


tomaarsen:
Hello!

I'll be glad to merge this if Transformers accepts your work. Feel free to ping me when that happens.

- Tom Aarsen

Issue #2533, opened 2024-03-10
if save can support ,safe_serialization=False


NLPJCL:
The current save function cannot be set safe_serialization。but defalut safe_serialization=True。thank！

tomaarsen:
Hello!

You're right, it would be nice if users could set `safe_serialization` so that they can save a `pytorch_model.bin` if they prefer. 

- Tom Aarsen

CKeibel:
Hi @tomaarsen ! 
If no one is working on this issue yet, could I work on this to get started with contributing?

tomaarsen:
That would be great!
I think we would need to wrap [this line](https://github.com/UKPLab/sentence-transformers/blob/66e0ee30843dd411c64f37f65447bb38c7bf857a/sentence_transformers/SentenceTransformer.py#L619) with `if isinstance(module, Transformer):` and then pass the `safe_serialization` down to the Transformer, while all other modules still stay as `module.save(model_path)`. 

- Tom Aarsen

CKeibel:
Great, thanks for the tip. I'll have a look at it.

llinauer:
This issue was apparently solved in https://github.com/UKPLab/sentence-transformers/pull/2542

tomaarsen:
You're right, thanks for mentioning. I forgot to link the issue to the pull request! I'll close this.

- Tom Aarsen

PR #2542, opened 2024-03-14
Enable saving modules as pytorch_model.bin

CKeibel:
Enables switching between `model.safetensors` and `pytorch_model.bin` when saving a model by setting the `safe_serialization` parameter to `True` (safetensors - default) or `False` (pytorch).

Resolves issue  #2533 .

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -619,7 +619,7 @@ class SentenceTransformer(nn.Sequential):
 
             os.makedirs(model_path, exist_ok=True)
             if isinstance(module, Transformer):
-                module.save(model_path, safe_serialization)
+                module.save(model_path, safe_serialization=safe_serialization)
             else:
                 module.save(model_path)
 
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -576,7 +576,7 @@ class SentenceTransformer(nn.Sequential):
         model_name: Optional[str] = None,
         create_model_card: bool = True,
         train_datasets: Optional[List[str]] = None,
-        safe_serialization: bool = True
+        safe_serialization: bool = True,
     ):
         """
         Saves all elements for this seq. sentence embedder into different sub-folders
@@ -622,7 +622,7 @@ class SentenceTransformer(nn.Sequential):
                 module.save(model_path, safe_serialization)
             else:
                 module.save(model_path)
-            
+
             modules_config.append(
                 {"idx": idx, "name": name, "path": os.path.basename(model_path), "type": type(module).__module__}
             )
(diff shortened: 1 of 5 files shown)


sentence_transformers/models/Transformer.py
```diff
@@ -156,8 +156,8 @@ def tokenize(self, texts: Union[List[str], List[Dict], List[Tuple[str, str]]]):
     def get_config_dict(self):
         return {key: self.__dict__[key] for key in self.config_keys}
 
-    def save(self, output_path: str):
-        self.auto_model.save_pretrained(output_path)
+    def save(self, output_path: str, safe_serialization: bool):
```
tomaarsen, line 159:
```suggestion
    def save(self, output_path: str, safe_serialization: bool = True):
```
This prevents people from suddenly experiencing a crash if they some reason call this method directly with just `save("my/path")`.

tests/test_sentence_transformer.py
```diff
@@ -175,6 +175,17 @@ def mock_list_repo_refs(self, repo_id=None, **kwargs):
             == 'Providing an `organization` to `save_to_hub` is deprecated, please use `repo_id="sentence-transformers-testing/stsb-bert-tiny-safetensors"` instead.'
         )
 
+@pytest.mark.parametrize("safe_serialization", [True, False])
```
tomaarsen, line 178: Nice!

tomaarsen (commented):
I think this is looking good at a glance, I'll review it more thoroughly tomorrow. Thanks for this already!

sentence_transformers/models/Transformer.py
CKeibel: Yeah, you're right.

CKeibel:
What about the `CrossEncoder`, that also seems to need the adjustment if I see it right?

tomaarsen:
Indeed! It would be excellent to get the support there, too. It should be fairly easy to implement given that a CrossEncoder always used `transformers`, unlike a Sentence Transformer.

- Tom Aarsen

CKeibel:
I'll have a quick look inside.

CKeibel:
I have added a test to test the default behaviour without the parameter. But I wasn't sure which would be the best way to that.

tomaarsen:
I like the approach that you've got now 👍

sentence_transformers/SentenceTransformer.py
```diff
@@ -616,7 +618,11 @@ def save(
                 model_path = os.path.join(path, str(idx) + "_" + type(module).__name__)
 
             os.makedirs(model_path, exist_ok=True)
-            module.save(model_path)
+            if isinstance(module, Transformer):
+                module.save(model_path, safe_serialization)
```
tomaarsen, line 622:
```suggestion
                module.save(model_path, safe_serialization=safe_serialization)
```

PR #2546, opened 2024-03-16
Fix SentenceTransformer encode documentation return type default (numpy vectors)

### LINKED ISSUE #1922: Error in encode method documentation of SentenceTransformer
In the `encode` method of `SentenceTransformer` class the signature of the method states that:

> By default, a list of tensors is returned. If convert_to_tensor, a stacked tensor is returned. If convert_to_numpy, a numpy matrix is returned.

However, I think this is not correct because the default value of `convert_to_numpy` is `True` and the default value of `convert_to_tensor` is `False`, therefore, the part of the code where the output type is defined will select the branch where a `ndarray` is returned.

```python
if convert_to_tensor:
    all_embeddings = torch.stack(all_embeddings)
elif convert_to_numpy:
    all_embeddings = np.asarray([emb.numpy() for emb in all_embeddings])
```

This behaviour can be verified by loading any model (e.g. all-mpnet-base-v2) and call the `encode` method on a sentence. The return type of the method is `numpy.ndarray`.

CKeibel:
Resolves #1922

I have checked the documentation and the code again and have come to the conclusion that this has not yet been fixed. If I am missing something, the PR can be closed again.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -281,8 +281,8 @@ class SentenceTransformer(nn.Sequential):
         :param normalize_embeddings: Whether to normalize returned vectors to have length 1. In that case,
             the faster dot-product (util.dot_score) instead of cosine similarity can be used.
 
-        :return: By default, a list of numpy vectors is returned. If convert_to_tensor, a stacked tensor is returned.
-            If convert_to_numpy, a list of numpy vectors is returned.
+        :return: By default, a numpy array is returned. If convert_to_tensor, a stacked tensor is returned.
+            If convert_to_numpy, a numpy array is returned.
         """
         self.eval()
         if show_progress_bar is None:
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -281,8 +281,9 @@ class SentenceTransformer(nn.Sequential):
         :param normalize_embeddings: Whether to normalize returned vectors to have length 1. In that case,
             the faster dot-product (util.dot_score) instead of cosine similarity can be used.
 
-        :return: By default, a numpy array is returned. If convert_to_tensor, a stacked tensor is returned.
-            If convert_to_numpy, a numpy array is returned.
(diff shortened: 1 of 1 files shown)


tomaarsen:
Oh, that was indeed very wrong. But also, isn't the default a numpy array, rather than a list of numpy vectors?

- Tom Aarsen

CKeibel:
> Oh, that was indeed very wrong. But also, isn't the default a numpy array, rather than a list of numpy vectors?
> 
>     * Tom Aarsen

That's right, I missed the second part and adjusted it.

CKeibel:
> Oh, that was indeed very wrong. But also, isn't the default a numpy array, rather than a list of numpy vectors?
> 
>     * Tom Aarsen

Right, I've missed that. I will fix it!

sentence_transformers/SentenceTransformer.py
```diff
@@ -281,8 +281,8 @@ def encode(
         :param normalize_embeddings: Whether to normalize returned vectors to have length 1. In that case,
             the faster dot-product (util.dot_score) instead of cosine similarity can be used.
 
-        :return: By default, a list of numpy vectors is returned. If convert_to_tensor, a stacked tensor is returned.
-            If convert_to_numpy, a list of numpy vectors is returned.
+        :return: By default, a numpy array is returned. If convert_to_tensor, a stacked tensor is returned.
+            If convert_to_numpy, a numpy array is returned.
```
tomaarsen, line 285:
```suggestion
        :return: By default, a 2d numpy array with shape [num_inputs, output_dimension] is returned. If only one string
            input is provided, then the output is a 1d array with shape [output_dimension]. If `convert_to_tensor`, a
            torch Tensor is returned instead.
```
Perhaps we can be more explicit and also list the output shape, like this?

CKeibel: I'm also more in favor of what you're suggesting. I just didn't want to make too big changes as it's only my second contribution.

tomaarsen: All good! I appreciate you spotting this & your assistance.

tomaarsen:
Thanks for this!

- Tom Aarsen

PR #2556, opened 2024-03-24
Fix smart_batching_collate Inefficiency

PrithivirajDamodaran:
SentenceTransformer.py:846 throws a Inefficiency warning:

>.....Creating a tensor from a list of numpy.ndarrays is extremely slow. Please consider converting the list to a single numpy.ndarray with numpy.array() before converting to a tensor. (Triggered internally at ../torch/csrc/utils/tensor_new.cpp:275.) labels = torch.tensor([example.label for example in batch])

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -843,7 +843,9 @@ class SentenceTransformer(nn.Sequential):
         """
         texts = [example.texts for example in batch]
         sentence_features = [self.tokenize(sentence) for sentence in zip(*texts)]
-        labels = torch.tensor([example.label for example in batch])
+        # Handle Inefficiency warning thrown: "Creating a tensor from a list of numpy.ndarrays is extremely slow..."
+        # labels = torch.tensor([example.label for example in batch])
+        labels = torch.tensor(np.array([example.label for example in batch]))
         return sentence_features, labels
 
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Interesting, `label` of `InputExample` is usually just an int or a float, not a numpy array. But you're right, if it is a numpy array (e.g. [model distillation](https://sbert.net/examples/training/distillation/README.html), [MSELoss](https://sbert.net/docs/training/loss_overview.html#distillation)), then we should not convert a list of numpy arrays to a tensor.
Perhaps it's better if we first verify whether the labels correspond to numpy arrays, and only then use `torch.from_numpy(array)`, and otherwise keep it as it is.

- Tom Aarsen

PrithivirajDamodaran:
Makes sense, will review and update.

KTibow:
@PrithivirajDamodaran update?

PrithivirajDamodaran:
@tomaarsen - Thanks for the feedback, incorporated. How does it look now ?

tomaarsen:
Apologies for the delay, I've been working on Sentence Transformers v3.0. It looks good! I added an `if labels ...` to ensure that `labels[0]` never causes any issues. Other than that, I simplified the comments a bit (ST is not that comment-heavy usually), and I think this should be ready to go. 

I'll merge this when the tests are green and then incorporate it into the v3.0 pre-release branch!

- Tom Aarsen

PR #2571, opened 2024-04-03
Update SentenceTransformer.py to use token length for sorting

### LINKED ISSUE #2570: encoding in batches with token length leads to different results?
Hi! I was just wondering:

https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L350

You sort by token length which is really great for performance, however I was wondering if it still could be that sentences with different token length get parsed and pad? For some models it was observed that pad token changes the mean slightly. I was wondering if this is the case that sentence transformers batches also different token lengths?

Thank you for the great repo!
Aaron

achibb:


tomaarsen:
Hello!

I prefer to sort by character length as it's too expensive (in my opinion) to tokenize twice just for sorting purposes. It's meant to be just a small optimization that should never hurt. Closing this.

- Tom Aarsen

PR #2573, opened 2024-04-03
[`feat`] Add truncation support

kddubey:
Hello,

This PR follows up on the discussion in #2564.

The implementation adds an optional `SentenceTransformer` instance attribute, `output_dim`, so that:

1. the dimension of `model.encode(texts)` agrees with `model.get_sentence_embedding_dimension()`
2. the user doesn't need to change every `model.encode(texts)` call to achieve truncation
3. no changes need to be made to third-party code which internally does `model.encode(texts)` (e.g., the `EmbeddingSimilarityEvaluator`) to achieve truncation.

In case a user wants to change the truncation dimension on the fly, they can use these new utilities:

* `model.truncate_sentence_embeddings`: context manager that sets and resets the truncation dimension, `output_dim`, of the `model`. This has the 3 benefits above
* `util.truncate_embeddings`: just a simple truncation function that works even if the input `embeddings` is 1-D or 2-D. May be useful because `...`-slicing is slightly niche knowledge.

## How has this code been tested?

New unit test:

```bash
pytest tests/test_sentence_transformer.py -k test_encode_truncate -x
```

It adds ~30 sec. to the testing workflow.

### DIFF
--- a/examples/training/matryoshka/README.md
+++ b/examples/training/matryoshka/README.md
@@ -65,7 +65,7 @@ matryoshka_dim = 64
 model = SentenceTransformer(
     "nomic-ai/nomic-embed-text-v1.5",
     trust_remote_code=True,
-    output_dim=matryoshka_dim,
+    truncate_dim=matryoshka_dim,
 )
 
 embeddings = model.encode(
--- a/examples/training/matryoshka/README.md
+++ b/examples/training/matryoshka/README.md
@@ -54,16 +54,20 @@ loss = Matryoshka2dLoss(model=model, loss=base_loss, matryoshka_dims=[768, 512, 
 
 ## Inference
 
-After a model has been trained using a Matryoshka loss, you can then run inference with it using <a href="../../../docs/package_reference/SentenceTransformer.html#sentence_transformers.SentenceTransformer.encode"><code>SentenceTransformers.encode</code></a>. You must then truncate the resulting embeddings, and it is recommended to renormalize the embeddings.
+After a model has been trained using a Matryoshka loss, you can then run inference with it using <a href="../../../docs/package_reference/SentenceTransformer.html#sentence_transformers.SentenceTransformer.encode"><code>SentenceTransformers.encode</code></a>.
 
 ```python
 from sentence_transformers import SentenceTransformer
 from sentence_transformers.util import cos_sim
 import torch.nn.functional as F
 
-model = SentenceTransformer("nomic-ai/nomic-embed-text-v1.5", trust_remote_code=True)
-
 matryoshka_dim = 64
+model = SentenceTransformer(
+    "nomic-ai/nomic-embed-text-v1.5",
+    trust_remote_code=True,
+    output_dim=matryoshka_dim,
+)
+
 embeddings = model.encode(
     [
         "search_query: What is TSNE?",
@@ -71,7 +75,7 @@ embeddings = model.encode(
         "search_document: Amelia Mary Earhart was an American aviation pioneer and writer.",
     ]
 )
-embeddings = embeddings[..., :matryoshka_dim]  # Shrink the embedding dimensions
+assert embeddings.shape[-1] == matryoshka_dim
 
 similarities = cos_sim(embeddings[0], embeddings[1:])
 # => tensor([[0.7839, 0.4933]])
@@ -86,6 +90,7 @@ See the following scripts as examples of how to apply the <a href="../../../docs
 
 * **[matryoshka_nli.py](matryoshka_nli.py)**: This example uses the MultipleNegativesRankingLoss with MatryoshkaLoss to train a strong embedding model using Natural Language Inference (NLI) data. It is an adaptation of the [NLI](../nli/README) documentation.
 * **[matryoshka_nli_reduced_dim.py](matryoshka_nli_reduced_dim.py)**: This example uses the MultipleNegativesRankingLoss with MatryoshkaLoss to train a strong embedding model with a small maximum output dimension of 256. It trains using Natural Language Inference (NLI) data, and is an adaptation of the [NLI](../nli/README) documentation.
+* **[matryoshka_eval_stsb.py](matryoshka_eval_stsb.py)**: This example evaluates the embedding model trained with MatryoshkaLoss in [matryoshka_nli.py](matryoshka_nli.py) on the test set of the STSBenchmark dataset, and compares it to a non-Matryoshka trained model.
 * **[matryoshka_sts.py](matryoshka_sts.py)**: This example uses the CoSENTLoss with MatryoshkaLoss to train an embedding model on the training set of the STSBenchmark dataset. It is an adaptation of the [STS](../sts/README) documentation.
 
 And the following scripts to see how to apply <a href="../../../docs/package_reference/losses.html#matryoshka2dloss"><code>Matryoshka2dLoss</code></a>:
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -594,7 +594,7 @@ class SentenceTransformer(nn.Sequential):
         return output_dim
 
     @contextmanager
-    def truncate_sentence_embeddings(self, output_dim: Optional[int]):
+    def truncate_dim(self, output_dim: Optional[int]):
         """
         In this context, `model.encode` outputs sentence embeddings truncated at dimension `output_dim`.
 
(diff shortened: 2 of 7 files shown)


sentence_transformers/SentenceTransformer.py
```diff
@@ -572,11 +578,48 @@ def get_sentence_features(self, *features):
...
     def get_sentence_embedding_dimension(self):
+        """
+        :return: The number of dimensions in the output of `encode`. If it's not known, it's `None`.
+        """
+        output_dim = None
         for mod in reversed(self._modules.values()):
             sent_embedding_dim_method = getattr(mod, "get_sentence_embedding_dimension", None)
             if callable(sent_embedding_dim_method):
-                return sent_embedding_dim_method()
-        return None
```
kddubey, line 579: I'm not sure what causes `return None`, i.e., when there's never a `get_sentence_embedding_dimension` attribute in the modules. In this case, the

```python
return min(output_dim or np.inf, self.output_dim)
```

returns `self.output_dim`, which could be greater than whatever dimension actually ends up being outputted.

examples/training/matryoshka/README.md
```diff
@@ -54,24 +54,28 @@ loss = Matryoshka2dLoss(model=model, loss=base_loss, matryoshka_dims=[768, 512,
...
 import torch.nn.functional as F
 
-model = SentenceTransformer("nomic-ai/nomic-embed-text-v1.5", trust_remote_code=True)
-
 matryoshka_dim = 64
+model = SentenceTransformer(
+    "nomic-ai/nomic-embed-text-v1.5",
+    trust_remote_code=True,
+    output_dim=matryoshka_dim,
+)
```
tomaarsen, line 69: Looks clean. I think `output_dim` is clearer than e.g. `matryoshka_dim` indeed.

sentence_transformers/SentenceTransformer.py
```diff
@@ -253,7 +258,7 @@ def encode(
         prompt: Optional[str] = None,
         batch_size: int = 32,
         show_progress_bar: bool = None,
-        output_value: str = "sentence_embedding",
+        output_value: Optional[Literal["sentence_embedding", "token_embeddings"]] = "sentence_embedding",
```
tomaarsen, line 261: Nice!

sentence_transformers/SentenceTransformer.py
```diff
@@ -370,8 +375,9 @@ def encode(
                         row = {name: out_features[name][sent_idx] for name in out_features}
                         embeddings.append(row)
                 else:  # Sentence embeddings
-                    embeddings = out_features[output_value]
+                    embeddings: torch.Tensor = out_features[output_value]
                     embeddings = embeddings.detach()
+                    embeddings = truncate_embeddings(embeddings, self.output_dim)
```
tomaarsen, line 380: This is the best spot to perform the truncation 👍

sentence_transformers/SentenceTransformer.py
tomaarsen: The `return None` is kind of moot, in practice this will essentially never happen: only if the `SentenceTransformer` is created with incorrect Modules. I think your current setup should be totally fine, at a glance.

sentence_transformers/SentenceTransformer.py
```diff
@@ -572,11 +578,48 @@ def get_sentence_features(self, *features):
...
+                output_dim = sent_embedding_dim_method()
+                break
+        if self.output_dim is not None:
+            # The user requested truncation. If they set it to a dim greater than output_dim,
+            # no truncation will actually happen. So return output_dim insead of self.output_dim
+            return min(output_dim or np.inf, self.output_dim)
+        return output_dim
+
+    @contextmanager
+    def truncate_sentence_embeddings(self, output_dim: Optional[int]):
```
tomaarsen, line 597: This is quite clean, though I'm not sure if people will immediately draw the connection between `truncate_sentence_embeddings` and `output_dim`. Perhaps this informs us that e.g. `truncate_dim` is a stronger name than `output_dim`?

tomaarsen (commented):
Great work! This is a really strong PR. I've made some comments, though I believe only 1 warrants a discussion (regarding `truncate_dim`).

sentence_transformers/SentenceTransformer.py
kddubey: Ah right. Let's go with `truncate_dim`

tomaarsen: Apologies, I should have been clearer, I think `truncate_sentence_embeddings` is fairly clear already, but perhaps we should rename `output_dim` to `truncate_dim`.

So,
```python
model = SentenceTransformer(
    "nomic-ai/nomic-embed-text-v1.5",
    trust_remote_code=True,
    truncate_dim=matryoshka_dim,
)
```

I'm curious about your thoughts here, too!

tomaarsen (commented):
I think this is looking solid now. I'll experiment with these changes locally myself to verify that it all works as expected, and perhaps I'll see if it also works correctly with some of our third party applications (langchain, llamaindex, etc.). I'll also double-check that the new methods are correctly included in the built docs.

sentence_transformers/SentenceTransformer.py
```diff
@@ -370,8 +376,9 @@ def encode(
                         row = {name: out_features[name][sent_idx] for name in out_features}
                         embeddings.append(row)
```
kddubey, line 377: Should truncation also be done when `output_value=None`? We'd only truncate the `out_features["sentence_embedding"]`

tomaarsen: That does seem like the behaviour that people would expect, yeah.

kddubey: Makes sense. Though this makes me wonder if it's more consistent to truncate `out_features["sentence_embedding"]` even earlier—after the forward pass of the pooler module. Maybe we'd have `Truncate` be its own torch module which gets appended to `self.modules`. That'd make truncation apply to training as well. Though that seems like a non-existent use case since Matryoshka training gets you more w/o any real computational cost

I'm not really for a `Truncate` module b/c it feels like too much to connect truncation to everything about a model, including how it gets trained and saved. I see truncation more as a little inference trick. Curious if you have thoughts on this

kddubey:
Interesting, first time I've seen `SequentialEvaluator`. Given that, a `truncate_dim` parameter would definitely be  convenient. I'll add it to the `EmbeddingSimilarityEvaluator` and others where I can see `model.encode` getting called

sentence_transformers/SentenceTransformer.py
kddubey: Fixed and tested in 8125182206bcde523da519d09e92f70ec9813312

tomaarsen: Personally, I think the module-based design can be a bit complicated, so I prefer this solution I think.

tomaarsen:
It just hasn't been type annotated, though I wouldn't be surprised if it results in a circular import loop. Then I tend to use
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:

...

    def __call__(self, model: "SentenceTransformer", ...
```

Feel free to add the type annotations, but you don't have to :) I can always do them later.

- Tom Aarsen

kddubey:
> Feel free to add the type annotations, but you don't have to :) I can always do them later.

Sounds good. I tested that there isn't a circular import. But I'll keep the changes in this PR limited to truncation.

kddubey:
I see that `precision`/quantization hasn't yet been added to other evaluators. Maybe in a future PR, support for `precision` and `truncate_dim` can be added. Lmk if you'd instead prefer to add support for `truncate_dim` in all evaluators in this PR.

tomaarsen:
> I see that `precision`/quantization hasn't yet been added to other evaluators. Maybe in a future PR, support for `precision` and `truncate_dim` can be added. Lmk if you'd instead prefer that this PR adds support for `truncate_dim` in all evaluators.

That is true, also because quantized embeddings (e.g. binary or int8) can't always be compared in the same way as regular float32 embeddings. E.g. for comparing binary embeddings, you normally do Hamming distance rather than cosine similarity. That complicates things somewhat for the evaluators.
Matryoshka-style truncation is a bit simpler, luckily. 

I think this PR should be kept as-is, and then we can extend the truncate_dim to other evaluators in future PRs.

- Tom Aarsen

tomaarsen:
Big thanks for this excellent community PR!

kddubey:
> E.g. for comparing binary embeddings, you normally do Hamming distance rather than cosine similarity.

Ohh interesting, good to know

> Big thanks for this excellent community PR!

You're welcome! And thank you for maintaining sentence-transformers :-)

kddubey:
> I think this PR should be kept as-is, and then we can extend the truncate_dim to other evaluators in future PRs.

I'll work on adding `truncate_dim` to the evaluators. And add `SentenceTransformer` type annotations in a separate PR

PR #2576, opened 2024-04-04
Document dev reqs, add ruff pre-commit

kddubey:
Hello,

This PR does 3 things:

1. Document local/development requirements so that it's slightly easier to know how to get started and test changes
2. Document the paths to apply ruff in `ruff.toml` instead of `Makefile` so that it applies to any `ruff` command (see [ruff docs](https://docs.astral.sh/ruff/configuration/#default-inclusions))
3.  Add [pre-commit ruff](https://docs.astral.sh/ruff/integrations/#pre-commit) hooks which are [equivalent](https://github.com/astral-sh/ruff-pre-commit/blob/main/.pre-commit-hooks.yaml) to `make quality`. Since there's a `quality.yml` check in the workflow, I figured it'd be nice to know whether or not a PR will pass it before pushing

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,3 +1,5 @@
+include = ["examples/**/*.py", "sentence_transformers/**/*.py", "tests/**/*.py", "docs/**/*.py", "setup.py"]
+
 lint.ignore-init-module-imports = true
 
 line-length = 119
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,5 +1,3 @@
-include = ["examples/**/*.py", "sentence_transformers/**/*.py", "tests/**/*.py", "docs/**/*.py", "setup.py"]
-
 lint.ignore-init-module-imports = true
 
 line-length = 119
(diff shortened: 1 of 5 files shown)


ruff.toml
```diff
@@ -1,3 +1,5 @@
+include = ["examples/**/*.py", "sentence_transformers/**/*.py", "tests/**/*.py", "docs/**/*.py", "setup.py"]
```
kddubey, line 1: Which ruff-checkable/formattable files are excluded from these? I'm wondering if this explicit setting can be dropped to keep things simple

tomaarsen: I think None, it seems plausible to drop this completely. I think this setup is just leftover from another project of mine, where some files were purposefully excluded.

tomaarsen (commented):
This seems like a wise addition indeed. I'll go over this in more detail tomorrow to see if there's some other places where we should describe the (new) developer guide. Thanks a bunch for this!

tomaarsen (approved):

PR #2578, opened 2024-04-06
Allow passing model_args to ST

satyamk7054:
## Summary
Allow passing model_args to ST

## Details
This fixes https://github.com/UKPLab/sentence-transformers/issues/2579.
New models like [e5-mistral-7b-instruct](https://huggingface.co/intfloat/e5-mistral-7b-instruct) use FP16 as the dtype.
However, when loaded using sentence_transformers, they are loaded with FP32. This is because
HF transformers uses FP32 as the default unless torch_dtype='auto' is passed, as mentioned [here](https://huggingface.co/docs/transformers/en/main_classes/model#transformers.PreTrainedModel.from_pretrained.torch_dtype).

Passing "auto" to model_args does not work because of the below error. Moreover, the SentenceTransformers class does not currently expose a model_args param.
```
cls = <class 'transformers.models.bert.modeling_bert.BertModel'>, dtype = 'auto'

    @classmethod
    def _set_default_torch_dtype(cls, dtype: torch.dtype) -> torch.dtype:
        """
        Change the default dtype and return the previous one. This is needed when wanting to instantiate the model
        under specific dtype.
    
        Args:
            dtype (`torch.dtype`):
                a floating dtype to set to.
    
        Returns:
            `torch.dtype`: the original `dtype` that can be used to restore `torch.set_default_dtype(dtype)` if it was
            modified. If it wasn't, returns `None`.
    
        Note `set_default_dtype` currently only works with floating-point types and asserts if for example,
        `torch.int64` is passed. So if a non-float `dtype` is passed this functions will throw an exception.
        """
>       if not dtype.is_floating_point:
E       AttributeError: 'str' object has no attribute 'is_floating_point'

../venv/lib/python3.10/site-packages/transformers/modeling_utils.py:1412: AttributeError
```

## Testing Done
Added a new unit tests that validates the dtype of the loaded model using the embedding tensor
created with it.


sentence_transformers/SentenceTransformer.py
```diff
@@ -87,6 +88,7 @@ def __init__(
         token: Optional[Union[bool, str]] = None,
         use_auth_token: Optional[Union[bool, str]] = None,
         truncate_dim: Optional[int] = None,
+        model_args: Optional[Dict[str, Any]] = None
     ):
```
tomaarsen, line 92: I'm not a fan of passing a dictionary of arguments. I think this can be rather confusing, as it's unclear what can be passed here. Instead, I think we should allow passing of specific, useful parameters, starting with:
* `torch_dtype`
* `attn_implementation`

Additionally, we can add `**model_kwargs` as the final parameter. So:
```suggestion
        torch_dtype: Optional[Union[str, torch.dtype]] = None,
        attn_implementation: Optional[str] = None,
        **model_kwargs,
    ):
        model_kwargs = {
            "torch_dtype": torch_dtype,
            "attn_implementation": attn_implementation,
            **model_kwargs,
        }
```

([docs](https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.from_pretrained))

Then the user can use:
```python
model = SentenceTransformer("all-mpnet-base-v2", torch_dtype=torch.float16, low_cpu_mem_usage=True)
```
etc.

The explicit parameters (torch_dtype, attn_implementation) will then be clear to the user. The big question that remains is which parameters should be used, e.g. there are a lot of options, but many are not commonly used for most: https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.from_pretrained.torch_dtype

Additionally, we might want to do some magic around `device_map` ([docs](https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.from_pretrained.device_map)) and `device` that currently exists. Not really sure about the best approach for that yet.

sentence_transformers/models/Transformer.py
```diff
@@ -32,8 +32,7 @@ def __init__(
         self.config_keys = ["max_seq_length", "do_lower_case"]
         self.do_lower_case = do_lower_case
 
-        config = AutoConfig.from_pretrained(model_name_or_path, **model_args, cache_dir=cache_dir)
-        self._load_model(model_name_or_path, config, cache_dir, **model_args)
```
tomaarsen, line 36: I like the change here; removing `config` should help & simplify without losing any behaviour.

tomaarsen (changes requested):
As mentioned in #2579, I think this PR will be preferable to #2426 due the hacky nature of the latter. I do think that some slight changes in the API would be beneficial, however.

Thanks for taking the time to make your issue & PR, and for dealing with my slow responses.

Also, please do share your opinion on these suggestions.

sentence_transformers/SentenceTransformer.py
satyamk7054: Thank you for taking the time to review this PR and look at the issue.


I've added torch_dtype and attn_implementation params to the signature

- I'm concerned if making all the other kwargs as model_kwargs will put us in a difficult situation if we ever need pass-through tokenizer_kwargs. Probably it is best to add params on need basis?
- For the documentation, I'm still directing to Transformers. Would you prefer directly phrasing the HF documentation? I think it would become an overhead to keep the two aligned as more features are added to HF (especially for attn_implementation)
- I'm also unclear on how to handle device_map. Probably it is best to defer model_kwargs for now to not end up in in a situation where both device_map and device are passed in.

Additionally, could you please suggest how would you want me to test for attn_implementation? Would a test that mocks the underlying call and validates correct arguments are passed to it suffice?

tomaarsen: > I'm concerned if making all the other kwargs as model_kwargs will put us in a difficult situation if we ever need pass-through tokenizer_kwargs. Probably it is best to add params on need basis?

This is a concern of mine too, also as we add more features, we could run into some unexpected changes. However, I think it might still be the best solution. However, I think the tokenizer kwargs are not that commonly used: most of the time any changes there result in degraded model performance.

> For the documentation, I'm still directing to Transformers. Would you prefer directly phrasing the HF documentation? I think it would become an overhead to keep the two aligned as more features are added to HF (especially for attn_implementation)

I think it makes sense to copy the latest version of HF for now, and I can change it when necessary. This will always be a bit tricky as someone can use Sentence Tranformers with different versions of Transformers and have different behaviour. There's not really any getting around that. 

I'm not sure about device_map either. I think it makes sense to just "leave it" for now.

As for the attn_implementation - yes, that would be more than sufficient. Actual tests would require additional complex installations, and currently only a few architectures support anything other than "eager". [I'm actively working on that](https://github.com/huggingface/transformers/pull/30450), though.

tomaarsen: @muellerzr I'm quite curious about your opinion on this topic: do you think we should introduce `**model_kwargs` to pass arguments to `AutoModel`? It could cause some unexpected behaviour in the future if we add new explicit keyword arguments (as they might not be directly passed down to `AutoModel` anymore), but it has the upside of a much cleaner API.

muellerzr: Post talking with Amy & Arthur (as they handle more of that side):

It depends a bit on how likely `sentence-transformers` is to either 
1) want to modify the values 
2) ever take these values themselves. 

I can imagine sentence-transformers might want to take e.g. torch_dtype in the future. This means it can no longer be in `model_kwargs` , so you’d need to e.g. pop it out. What happens if the torch_dtype wanted by sentence transformers is now different e.g. acceptable values than what we can pass to `AutoModel`? The problem with passing all the kwargs blindly, is you don’t even know if there’s going to be argument clashes for a new kwarg!

Your version here is adaptable and flexible, very nice! You'll still have that future kwarg clashing, but here it's more pragmatic and explicit.

In general: kwargs are a pain, and if it were in raw transformers we'd say no probably, unless we were in your shoes and calling another lib we don't control, in which being more flexible and doing exactly this could be good.

(Paraphrasing Amy's response to me)

satyamk7054: Thank you for the suggestions and explanation, I've updated the PR.

satyamk7054:
@tomaarsen could you please take a look at the updated PR?

sentence_transformers/SentenceTransformer.py
```diff
@@ -1213,7 +1245,7 @@ def _load_auto_model(
                 "trust_remote_code": trust_remote_code,
                 "revision": revision,
                 "local_files_only": local_files_only,
-            },
+            } | (model_args or {}),
```
tomaarsen, line 1248: The `|` operator isn't supported in Python 3.8, and I'd like to keep supporting that version until [~6 months from now](https://endoflife.date/python).

sentence_transformers/SentenceTransformer.py
```diff
@@ -1332,6 +1365,8 @@ def _load_sbert_model(
                     kwargs["model_args"].update(hub_kwargs)
                 else:
                     kwargs["model_args"] = hub_kwargs
+                kwargs["model_args"].update(model_args or {})
```
tomaarsen, line 1368:
```suggestion
                if model_args:
                    kwargs["model_args"].update(model_args)
```
nit: I think this is slightly cleaner than calling `update` with an empty dict

tomaarsen (commented):
Two small nitpicks, but otherwise this is looking strong!

sentence_transformers/SentenceTransformer.py
satyamk7054: 👍 Updated to use Python 3.8 compatible code

satyamk7054:
@tomaarsen could you please look at this PR again?

tomaarsen (approved):
I will do some more local tests before I merge this, but I imagine that it's ready to go. I'll merge this into `master` and then propagate those changes into [v3.0-pre-release](https://github.com/UKPLab/sentence-transformers/tree/v3.0-pre-release). 
For context, v3 should release quite soon. I am working on refactoring the sbert.net documentation, but it'll be released once that's done.

- Tom Aarsen

tomaarsen:
Hello @satyamk7054 

I've done some more thinking, and it seems that there is some (albeit rather niche) needs to be able to specify parameters for the tokenizer and config ([my comment on that](https://github.com/UKPLab/sentence-transformers/pull/2630#issuecomment-2103095087)), so I think it might be best to go with your original idea of letting users specify a `model_args` dictionary (but then called `model_kwargs`, that's a more accurate name I think) and passing it directly to the transformer. Idem with tokenizers and configs. What do you think?

- Tom Aarsen

sentence_transformers/SentenceTransformer.py
```diff
@@ -89,6 +122,9 @@ def __init__(
         token: Optional[Union[bool, str]] = None,
         use_auth_token: Optional[Union[bool, str]] = None,
         truncate_dim: Optional[int] = None,
+        model_kwargs: Optional[Dict[str, Any]] = {},
+        tokenizer_kwargs: Optional[Dict[str, Any]] = {},
+        config_kwargs: Optional[Dict[str, Any]] = {},
```
satyamk7054, line 127: Should these be defaulted to None?

satyamk7054:
@tomaarsen Thank you for more testing and updating the PR! I read through the discussion in the linked issue and I think adding these additional kwargs makes sense (thank you for letting me know).

sentence_transformers/SentenceTransformer.py
tomaarsen: They should indeed! Thanks for spotting

tomaarsen:
@muellerzr @osanseviero I was wondering if you could have another quick look. I've changed my mind after a user required `use_fast=False` in their Tokenizer and another user wanted to load a model with a higher dropout for training (i.e., this can be passed to the config). We now have three `..._kwargs` parameters to allow the users perfect control over the underlying `transformers` objects, while allowing me to introduce new parameters as well. I think this is much safer, as much as I dislike having users need to provide dictionaries with values.

This is still a fairly major change, so I wanted your opinions on this, if you have a bit of time.

- Tom Aarsen

muellerzr (approved):
Very nice! I like that **much** more, and am thankful you document **where** to find those easily! Keeps your side a bit easier too as things change/etc 🤗

ZhengHongming888:
This PR is good that giving more flexible control over the transformers from 3 .. _kwargs because later on the user will need more /more freedom to add something on top of transformers. Just carefully take a look at this PR this morning .. :-) Good work! @satyamk7054  @tomaarsen

PR #2582, opened 2024-04-09
Add truncation support in evaluators

kddubey:
Hello,

This PR is a follow-up to #2573. As you suggested [in this comment](https://github.com/UKPLab/sentence-transformers/pull/2573#issuecomment-2038176178), a `truncate_dim` parameter makes it easy to construct a sequential evaluator.

## How has this been tested?

It hasn't 🥴 
I tested `EmbeddingSimilarityEvaluator` in #2573 by running a notebook offline. I'll think about how to test this change soon. Lmk what you think would make for a sufficient test, e.g., offline runs or actual tests in `tests/test_evaluator.py`.

### DIFF
--- a/sentence_transformers/evaluation/BinaryClassificationEvaluator.py
+++ b/sentence_transformers/evaluation/BinaryClassificationEvaluator.py
@@ -114,13 +114,15 @@ class BinaryClassificationEvaluator(SentenceEvaluator):
     def __call__(self, model, output_path: str = None, epoch: int = -1, steps: int = -1) -> float:
         if epoch != -1:
             if steps == -1:
-                out_txt = f" after epoch {epoch}:"
+                out_txt = f" after epoch {epoch}"
             else:
-                out_txt = f" in epoch {epoch} after {steps} steps:"
+                out_txt = f" in epoch {epoch} after {steps} steps"
         else:
-            out_txt = ":"
+            out_txt = ""
+        if self.truncate_dim is not None:
+            out_txt += f" (truncated to {self.truncate_dim})"
 
-        logger.info("Binary Accuracy Evaluation of the model on " + self.name + " dataset" + out_txt)
+        logger.info(f"Binary Accuracy Evaluation of the model on the {self.name} dataset{out_txt}:")
 
         scores = self.compute_metrices(model)
 
--- a/sentence_transformers/evaluation/BinaryClassificationEvaluator.py
+++ b/sentence_transformers/evaluation/BinaryClassificationEvaluator.py
@@ -1,3 +1,4 @@
+from contextlib import nullcontext
 from . import SentenceEvaluator
 import logging
 import os
@@ -5,7 +6,7 @@ import csv
 from sklearn.metrics.pairwise import paired_cosine_distances, paired_euclidean_distances, paired_manhattan_distances
 from sklearn.metrics import average_precision_score
 import numpy as np
-from typing import List
+from typing import List, Optional
 from ..readers import InputExample
 
 
@@ -30,6 +31,8 @@ class BinaryClassificationEvaluator(SentenceEvaluator):
     :param batch_size: Batch size used to compute embeddings
     :param show_progress_bar: If true, prints a progress bar
     :param write_csv: Write results to a CSV file
+    :param truncate_dim: The dimension to truncate sentence embeddings to. `None` uses the model's current truncation
+        dimension. Defaults to None.
     """
 
     def __init__(
@@ -41,10 +44,12 @@ class BinaryClassificationEvaluator(SentenceEvaluator):
         batch_size: int = 32,
         show_progress_bar: bool = False,
         write_csv: bool = True,
+        truncate_dim: Optional[int] = None,
     ):
         self.sentences1 = sentences1
         self.sentences2 = sentences2
         self.labels = labels
+        self.truncate_dim = truncate_dim
 
         assert len(self.sentences1) == len(self.sentences2)
         assert len(self.sentences1) == len(self.labels)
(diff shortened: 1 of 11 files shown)


sentence_transformers/evaluation/BinaryClassificationEvaluator.py
```diff
@@ -144,25 +149,31 @@ def __call__(self, model, output_path: str = None, epoch: int = -1, steps: int =
         return main_score
 
     def compute_metrices(self, model):
-        try:
-            # If the sentences are hashable, then we can use a set to avoid embedding the same sentences multiple times
-            sentences = list(set(self.sentences1 + self.sentences2))
```
kddubey, line 149: I think the intention here was to catch the `TypeError` *only* out of the set(...)` conversion. So move the rest of the `try` block here to the `else` block

sentence_transformers/evaluation/BinaryClassificationEvaluator.py
```diff
@@ -144,25 +149,31 @@ def __call__(self, model, output_path: str = None, epoch: int = -1, steps: int =
...
-            # Otherwise we just embed everything, e.g. if the sentences are images for evaluating a CLIP model
-            embeddings = model.encode(
-                self.sentences1 + self.sentences2,
-                batch_size=self.batch_size,
-                show_progress_bar=self.show_progress_bar,
-                convert_to_numpy=True,
-            )
-            embeddings1 = embeddings[: len(self.sentences1)]
-            embeddings2 = embeddings[len(self.sentences1) :]
+        with nullcontext() if self.truncate_dim is None else model.truncate_sentence_embeddings(self.truncate_dim):
```
kddubey, line 152: Could instead do:

```python
with model.truncate_sentence_embeddings(self.truncate_dim or model.truncate_dim):
    ...
```

I made the current choice b/c then the `model` doesn't have to have a `truncate_sentence_embeddings` method for the evaluator to work when `self.truncate_dim is None` (b/c the last part of the ternary is never evaluated by Python). It is a little more to read though, so I'm open to changing it to the style above

tomaarsen: I like the `nullcontext` approach, that's how I would have done it too.

sentence_transformers/evaluation/BinaryClassificationEvaluator.py
tomaarsen: Sounds good!

tomaarsen:
Hello!

Thanks a bunch for this PR! I've extended it in the following two ways:
1. Add `truncate_dim` to `ParaphraseMiningEvaluator`. Although that evaluator doesn't directly use model.encode itself, it does call `paraphrase_mining` which uses `model.encode` under the hood. That way we can still support it.
2. Add `(truncated to ...)` to the logging for the Evaluators. This way you can easily create 2 identical evaluators, but one truncated to a specific length, and it'll be clear which results are which in the logs. In practice, however, I would recommend using the `name` optional argument to specify the length, as that will eventually be included in the logs via #2449

- Tom Aarsen

tomaarsen:
Additionally, I've extended the existing matryoshka NLI training scripts to show how to use the EmbeddingSimilarityEvaluator with `truncate_dim` to evaluate it on different embedding dimensionalities. The training logs will now include e.g.:
```
2024-04-10 12:37:25 - EmbeddingSimilarityEvaluator: Evaluating the model on the sts-dev-768 dataset in epoch 0 after 440 steps (truncated to 768):
2024-04-10 12:37:27 - Cosine-Similarity :       Pearson: 0.8425 Spearman: 0.8500
2024-04-10 12:37:27 - Manhattan-Distance:       Pearson: 0.8302 Spearman: 0.8286
2024-04-10 12:37:27 - Euclidean-Distance:       Pearson: 0.8317 Spearman: 0.8301
2024-04-10 12:37:27 - Dot-Product-Similarity:   Pearson: 0.7044 Spearman: 0.7021
2024-04-10 12:37:27 - EmbeddingSimilarityEvaluator: Evaluating the model on the sts-dev-512 dataset in epoch 0 after 440 steps (truncated to 512):
2024-04-10 12:37:29 - Cosine-Similarity :       Pearson: 0.8437 Spearman: 0.8511
2024-04-10 12:37:29 - Manhattan-Distance:       Pearson: 0.8302 Spearman: 0.8285
2024-04-10 12:37:29 - Euclidean-Distance:       Pearson: 0.8318 Spearman: 0.8299
2024-04-10 12:37:29 - Dot-Product-Similarity:   Pearson: 0.7209 Spearman: 0.7172
2024-04-10 12:37:29 - EmbeddingSimilarityEvaluator: Evaluating the model on the sts-dev-256 dataset in epoch 0 after 440 steps (truncated to 256):
2024-04-10 12:37:30 - Cosine-Similarity :       Pearson: 0.8393 Spearman: 0.8496
2024-04-10 12:37:30 - Manhattan-Distance:       Pearson: 0.8276 Spearman: 0.8263
2024-04-10 12:37:30 - Euclidean-Distance:       Pearson: 0.8287 Spearman: 0.8278
2024-04-10 12:37:30 - Dot-Product-Similarity:   Pearson: 0.7090 Spearman: 0.7086
2024-04-10 12:37:30 - EmbeddingSimilarityEvaluator: Evaluating the model on the sts-dev-128 dataset in epoch 0 after 440 steps (truncated to 128):
2024-04-10 12:37:32 - Cosine-Similarity :       Pearson: 0.8268 Spearman: 0.8398
2024-04-10 12:37:32 - Manhattan-Distance:       Pearson: 0.8227 Spearman: 0.8229
2024-04-10 12:37:32 - Euclidean-Distance:       Pearson: 0.8226 Spearman: 0.8232
2024-04-10 12:37:32 - Dot-Product-Similarity:   Pearson: 0.6628 Spearman: 0.6718
2024-04-10 12:37:32 - EmbeddingSimilarityEvaluator: Evaluating the model on the sts-dev-64 dataset in epoch 0 after 440 steps (truncated to 64):
2024-04-10 12:37:34 - Cosine-Similarity :       Pearson: 0.8168 Spearman: 0.8338
2024-04-10 12:37:34 - Manhattan-Distance:       Pearson: 0.8085 Spearman: 0.8110                                                                                     
2024-04-10 12:37:34 - Euclidean-Distance:       Pearson: 0.8101 Spearman: 0.8126                                                                                     
2024-04-10 12:37:34 - Dot-Product-Similarity:   Pearson: 0.6047 Spearman: 0.6165
```

(I considered using a separate PR for that, but I think this fits as well)

I'm planning on merging this today or early tomorrow, with the intention of using your Matryoshka-focused improvements as the headliner for a new v2.7.0 release. I've manually tested each of the evaluators by running an example training script for each of them, but in the long term we definitely want to add proper tests for them. For this PR, I think we're okay without.
I also want to note that #2449 will also update the evaluators: they will start returning dictionaries with metric names as the keys. 

- Tom Aarsen

tomaarsen:
Awesome! Also, as for the v2.7.0 release that I mentioned: I think I will postpone that for now.

I think this PR should be good to go now though! Thanks a bunch for setting this up :)

- Tom Aarsen

PR #2586, opened 2024-04-10
Add ST annotation to evaluators

kddubey:
Hello,

Small follow-up. There isn't a circular import (in fact there's [already an import in main](https://github.com/UKPLab/sentence-transformers/blob/53a247048433b56ad93d83224ac27fae5f981c8d/sentence_transformers/evaluation/MSEEvaluatorFromDataFrame.py#L2)).

Another small thing is that the imports aren't in isort-like order. In case you want to apply that order, run—

```
ruff check sentence_transformers/evaluation --fix --select I
```

—and add `extend-select = ["I"]` to `ruff.toml`.

But I'm not sure if you want this style enforced. I'll leave that up to you to commit :-)


kddubey:
Resolving the conflicts now

tomaarsen:
Thank you!

kddubey:
Resolved

tomaarsen:
Thanks a bunch! PR looks good :)

- Tom Aarsen

tomaarsen:
As for `ruff check sentence_transformers/evaluation --fix --select I` - yes, I think in the long run we do want to  do proper import sorting. However, I think it's likely to result in a lot of merge conflicts, so I'd like to hold off on until most PRs for the v3 release are merged.

- Tom Aarsen

PR #2590, opened 2024-04-12
corrected comment from kmeans to agglomerative

DhruvMakwana:


tomaarsen:
Well spotted! Thanks for this fix!

- Tom Aarsen

Issue #2583, opened 2024-04-09
Combining GradCache with GISTEmbedLoss


JacksonCakes:
Hey! I've been using CachedMultipleNegativesRankingLoss to fine-tune my embedding model because I'm working with limited GPU resources and can't go for a big batch size, which is what MNR Loss needs to really shine. But I've hit a bit of a wall in terms of performance with my current dataset. So, I'm thinking about giving GISTEmbedLoss with GradCache a shot. It seems like it could offer the best of both worlds: saving memory and enhancing what MNR Loss does. What do you think? I could try and work on this if needed

tomaarsen:
Hello!

I agree 100%. I have considered this myself, but have been too busy with #2449 to already work on it. A `CachedGISTEmbedLoss` would be excellent. I think a good starting point would be to begin with a copy of `CachedMultipleNegativesRankingLoss` and try and adapt it to include the "guided filtering" that GISTEmbedLoss introduces. I hope that this should be a fairly simple addition, e.g. `embed_minibatch` and `embed_minibatch_iter` should also get the embeddings for the guiding model. Then, in the `forward` method [here](https://github.com/UKPLab/sentence-transformers/blob/4357018e664b05964dcf3a8e553bab6f17aa17ec/sentence_transformers/losses/CachedMultipleNegativesRankingLoss.py#L220) we'll have two lists: embeddings (i.e. `reps` for representations here) for the main model and for the guide model. These can both be passed to `calculate_loss_and_cache_gradients` where the [scores](https://github.com/UKPLab/sentence-transformers/blob/4357018e664b05964dcf3a8e553bab6f17aa17ec/sentence_transformers/losses/CachedMultipleNegativesRankingLoss.py#L207) are set to `-inf` at certain positions.

I would certainly welcome your (or the community's) help on this :)

cc @kwang2049 as he created Cached MNRL 
cc @avsolatorio as he created GISTEmbedLoss

- Tom Aarsen

JacksonCakes:
Thanks for the suggestions! I will try it! :)

tomaarsen:
Awesome! I'll assign the issue to you :) Good luck.

tomaarsen:
This has been implemented and released, great work!

PR #2592, opened 2024-04-13
Add CachedGISTEmbedLoss

JacksonCakes:
As per discussed in #2583, this is the implementation of GradCache version of GISTEmbedLoss to reduce memory usage while maintaining performance levels comparable to those of GISTEmbedLoss.


tomaarsen:
Hello!

Thanks a bunch for this! I've been testing this alongside MNRL, CMNRL and normal GIST yesterday and today. It seems to roughly match their performance, though I'm using some simple training & testing data.
I'll assist with finishing up this PR today, which mostly involves making sure that CachedGISTEmbedLoss is mentioned in the documentation.

- Tom Aarsen

JacksonCakes:
Hi! Sorry, I just realized that calculating similarity for the entire batch in the guide, instead of using mini-batches, also adds extra memory usage. That's not ideal since I can handle up to a batch size of 4096 with CMNR loss, but only 1024 for this. I've made some adjustments to the guided part to address this.

tomaarsen:
I think there's a small issue with 3208e611247f920ffa18262e1d1839a8f85af340, the loss is always 0 it seems. These are some of my logs: 
image

(Green is 3208e611247f920ffa18262e1d1839a8f85af340, Salmon is 5c054da665ec0e82d834f37edbb1767eb11ef093)

Additionally, the memory usage is actually a tad higher:
image

But that might also be somehow related to the 0 loss.

- Tom Aarsen

JacksonCakes:
Ah my bad, i think it's because I did not properly offset the diagonal part resulting in the guide mask always select diagonally from the first element, which easily causing -inf in the scores and potentially leading to weird loss.

tomaarsen:
After further experimentation, I can confirm that 3215c06 matches the performance (losses, evaluations) of 5c054da665ec0e82d834f37edbb1767eb11ef093 exactly, but the former allows for much higher batch sizes. E.g. I was able to set my batch size to an absurd 50k.

Great job!

I'll work on the documentation things that I had mentioned.

- Tom Aarsen

JacksonCakes:
Good to know! Thank you for your effort! Happy to help :)

tomaarsen:
I think this is ready! I'll merge it now, so it can be included in tomorrow/Thursday's release. Do feel free to let me know if there were things that you think are missing/suboptimal.

Thanks for your time/work on this!

cc @avsolatorio, @kwang2049

- Tom Aarsen

PR #2603, opened 2024-04-20
[feat] Add local_files_only argument to load model from cache

debanjum:
This PR adds the `local_files_only` keyword argument to
`SentenceTransformers`, `CrossEncoder` classes, and propagates it down to
all huggingface_hub and transformer functions that load from the Hub.

Setting `local_files_only=True` when instantiating sentence
transformer models should speed up cached model load time when offline.

Previously passing `local_files_only` kwarg to huggingface_hub, transformer
function wasn't possible via the sentence transformer library.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -69,6 +69,7 @@ class SentenceTransformer(nn.Sequential):
         will execute code present on the Hub on your local machine.
     :param revision: The specific model version to use. It can be a branch name, a tag name, or a commit id,
         for a stored model on Hugging Face.
+    :param local_files_only: If `True`, avoid downloading the model. Returns path to the local cached model if it exists.
     :param token: Hugging Face authentication token to download private models.
     :param truncate_dim: The dimension to truncate sentence embeddings to. `None` does no truncation. Truncation is
         only applicable during inference when `.encode` is called.
@@ -84,6 +85,7 @@ class SentenceTransformer(nn.Sequential):
         cache_folder: Optional[str] = None,
         trust_remote_code: bool = False,
         revision: Optional[str] = None,
+        local_files_only: bool = False,
         token: Optional[Union[bool, str]] = None,
         use_auth_token: Optional[Union[bool, str]] = None,
         truncate_dim: Optional[int] = None,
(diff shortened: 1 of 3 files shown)


tomaarsen:
Hello!

I followed along slightly in the issue on `huggingface_hub`, so I understand the reasoning for this change.

The implementation looks solid at a glance. I'll have a deeper look on Monday, but I think this makes a lot of sense.

- Tom Aarsen

debanjum:
Thanks for the quick response! Take your time, didn't mean to ping you on your weekend (lost track of the day myself 😅 )

tomaarsen:
All good, maintenance is as much of a hobby as a job for me, I don't mind having a quick look at things on the weekend 🤗

- Tom Aarsen

sentence_transformers/SentenceTransformer.py
```diff
@@ -69,6 +69,7 @@ class SentenceTransformer(nn.Sequential):
         will execute code present on the Hub on your local machine.
     :param revision: The specific model version to use. It can be a branch name, a tag name, or a commit id,
         for a stored model on Hugging Face.
+    :param local_files_only: If `True`, avoid downloading the model. Returns path to the local cached model if it exists.
```
tomaarsen, line 72:
```suggestion
    :param local_files_only: If `True`, avoid downloading the model.
```

sentence_transformers/cross_encoder/CrossEncoder.py
```diff
@@ -44,6 +44,7 @@ class CrossEncoder(PushToHubMixin):
         will execute code present on the Hub on your local machine.
     :param revision: The specific model version to use. It can be a branch name, a tag name, or a commit id,
         for a stored model on Hugging Face.
+    :param local_files_only: If `True`, avoid downloading the model. Returns path to the local cached model if it exists.
```
tomaarsen, line 47:
```suggestion
    :param local_files_only: If `True`, avoid downloading the model.
```

tomaarsen (changes requested):
Some small docstring changes are necessary, as they're a bit confusing now.
This is actually perfect timing for this, as the HF Hub seems to be slow/inacessible right now: https://status.huggingface.co/ 😅

debanjum:
Fixed the doc strings! Let me know if any other updates are required?

tomaarsen:
That was all, thanks a bunch!

- Tom Aarsen

PR #2557, opened 2024-03-24
Enable Sentence Transformer Inference with Intel Gaudi2 GPU Supported ( 'hpu' )

ZhengHongming888:
This PR belongs to one of enabling Intel's Gaudi2 GPU supported tasks for Sentence Transformer's inference/training. 

This is the first PR including the items as below -

1. Add 'hpu' device name in get_device_name() under sentence_transformers/util.py
2. Add padding strategy argument in tokenize() for Gaudi2 device in sentence_transformers/SentenceTransformer.py and sentence_transformers/models/Transformer.py. Keep the original padding strategy for cuda/cpu as the first priority. 
3. Enable graph mode for Gaudi2 device with better performance in sentence_transformers/SentenceTransformer.py.

There is no modification for any inference examples which will seamlessly choose the default device like 'cuda' in cuda system, 'hpu' in Gaudi2 system, neither of above two will choose 'cpu', etc. 

Welcome for any questions/comments!

Thanks.

PR #2608, opened 2024-04-23
Enable Sentence Transformer Inference with Intel Gaudi2 GPU Supported ( 'hpu' ) - Follow up for #2557

ZhengHongming888:
This PR belongs to one of enabling Intel's Gaudi2 GPU supported tasks for Sentence Transformer's inference/training.

This is the follow up PR to [2557](https://github.com/UKPLab/sentence-transformers/pull/2557). There are few following considerations/modifications in this new PR on hpu device  -

1.  Padding strategy for hpu device

    We have tested the performance over all 37 models listed in [pretrained models page ](https://www.sbert.net/docs/pretrained_models.html) and compared the performance between padding strategy of 'max_length' (str) and 'longest_length' (default value = True).  Found over 2/3 of all 37 models the padding strategy from 'longest_length' is better than 'max_length' option for hpu device based on the [eval](https://github.com/UKPLab/sentence-transformers/blob/master/examples/evaluation/evaluation_inference_speed.py) over 100000 sentences. So we propose to also change hpu's padding setting into 'longest_length' (value = True) and same as previous default setting. 
   But also we want to expose this padding setting as one argument in SentenceTransformer class definition which the users can easily/freely choose what they want for padding strategy and in some cases the 'max_length' padding strategy will show better performance. 

2. hpu support for PR ( [2573](https://github.com/UKPLab/sentence-transformers/pull/2573)

    Test this PR on hpu device/ failed. So we proposed the revision version to support this PR with hpu device. So now the test case of this PR (# test_encode_truncate) under tests/test_sentence_transformer.py also successfully passed. 

3. hpu support for PR [2599](https://github.com/UKPLab/sentence-transformers/pull/2599)

    Test this PR on hpu device/ failed. So we proposed the revision version to support this PR with hpu device. So now the test case of this PR (# test_simple_encode ) under tests/test_image_embeddings.py also successfully passed

Welcome for any questions/comments!

Thanks.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -20,6 +20,7 @@ from tqdm.autonotebook import trange
 import math
 import queue
 import tempfile
+import copy
 
 from . import __MODEL_HUB_ORGANIZATION__
 from .evaluation import SentenceEvaluator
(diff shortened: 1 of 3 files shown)


ZhengHongming888:
Will submit one new PR with complete performance results!

PR #2609, opened 2024-04-24
Updated urls for msmarco dataset

j-dominguez9:
previous URL isn't accessible, updated to new one.


tomaarsen (commented):
Much appreciated! I can find the new URLs in https://microsoft.github.io/msmarco/Datasets.html and they're accessible for me. 

I can find 2 more broken links in `translate_queries`, which I'll update. I'll also take care of the CI complaint about `pre-commit`/`make style`.

- Tom Aarsen

PR #2614, opened 2024-04-25
Fall back to CPU device in case there are no PyTorch parameters

### LINKED ISSUE #2607: Models without PyTorch parameters don't work since v2.3.0
Hi, models that don't have any PyTorch parameters seem broken since sentence-transformers v2.3.0, e.g.

```python
from sentence_transformers import SentenceTransformer, models

st = SentenceTransformer(modules=[models.BoW(vocab=["a", "b", "c"])])
st.encode(["a"])
```

works as expected in v2.2.2 but crashes since v2.3.0 inside `self.device` because `next(...)` in the PyTorch 1.5 fallback branch raises `StopIteration`  (https://github.com/UKPLab/sentence-transformers/blob/v2.3-release/sentence_transformers/SentenceTransformer.py#L1122).

(More context: we sometimes use `SentenceTransformer` to wrap calls an embedding API so we can use them in the same way as local trainable models, now we have to add a dummy parameter such as `self.dummy = nn.Linear(1, 1)` to make it work)

I haven't tried the v3 branches yet, feel free to close in case this is fixed soon since it's only a small inconvenience. Thanks!

maxfriedrich:
Fixes #2607

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -1388,6 +1388,7 @@ class SentenceTransformer(nn.Sequential):
     def device(self) -> device:
         """
         Get torch.device from module, assuming that the whole module has one device.
+        In case there are no PyTorch parameters, fall back to CPU.
         """
         try:
             return next(self.parameters()).device
@@ -1399,8 +1400,11 @@ class SentenceTransformer(nn.Sequential):
                 return tuples
 
             gen = self._named_members(get_members_fn=find_tensor_attributes)
-            first_tuple = next(gen)
-            return first_tuple[1].device
+            try:
+                first_tuple = next(gen)
+                return first_tuple[1].device
+            except StopIteration:
+                return torch.device("cpu")
 
     @property
     def tokenizer(self):


tomaarsen:
cc @michaelfeil your thoughts would be much appreciated, if you have the time!

Otherwise I'll merge this as-is, I think it looks very reasonable. Thanks for the PR!

- Tom Aarsen

michaelfeil:
I think this behavior is a good idea.

nit  - I think you can also fallback without a `try:.. except: ..` as `next` provides a default.  https://docs.python.org/3/library/functions.html#next  `next(iterator, default)`

maxfriedrich:
@michaelfeil The iterator doesn't return the device directly so we can't just write `next(gen, torch.device("cpu"))`, this is why I think it's nicer to keep the explicit `try: ... except: ...`, what do you think?

I only have the issue that models without any PyTorch parameters crash with `StopIteration` there, no other problems at the moment.

tomaarsen:
I'm okay with the nested try-except, as I understand that the clean `next(iterator, default)` approach does not conveniently work here. Thanks for this PR! I think it'll become especially useful when I try to add ONNX support for Sentence Transformers.

- Tom Aarsen

PR #2619, opened 2024-04-27
Bug in DenoisingAutoEncoderLoss.py

arun477:
transformers [**PreTrainedModel._tie_encoder_decoder_weight**](https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py) in [DenoisingAutoEncoderLoss](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/losses/DenoisingAutoEncoderLoss.py#L116) requires positional argument base_encoder_name.

image
image
image

### DIFF
--- a/sentence_transformers/losses/DenoisingAutoEncoderLoss.py
+++ b/sentence_transformers/losses/DenoisingAutoEncoderLoss.py
@@ -116,14 +116,19 @@ class DenoisingAutoEncoderLoss(nn.Module):
             try:
                 # Compatibility with transformers <4.40.0
                 PreTrainedModel._tie_encoder_decoder_weights(
-                    model[0].auto_model, self.decoder._modules[decoder_base_model_prefix], self.decoder.base_model_prefix
+                    model[0].auto_model,
+                    self.decoder._modules[decoder_base_model_prefix],
+                    self.decoder.base_model_prefix,
                 )
             except TypeError:
                 # Compatibility with transformers >=4.40.0
                 PreTrainedModel._tie_encoder_decoder_weights(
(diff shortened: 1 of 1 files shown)


sentence_transformers/losses/DenoisingAutoEncoderLoss.py
```diff
@@ -114,7 +114,7 @@ def __init__(self, model: SentenceTransformer, decoder_name_or_path: str = None,
                 )
             decoder_base_model_prefix = self.decoder.base_model_prefix
             PreTrainedModel._tie_encoder_decoder_weights(
```
tomaarsen, line 116: Hello!

Well spotted; I see this has been introduced in transformers v4.40.0 [here](https://github.com/huggingface/transformers/pull/29906). Because I'd rather not increment the minimum `transformers` all the way to v4.44.0 yet (I imagine that might cause problems), I propose an alternative solution:

```python
try:
    # Compatibility with transformers <4.40.0
    PreTrainedModel._tie_encoder_decoder_weights(
        model[0].auto_model, self.decoder._modules[decoder_base_model_prefix], self.decoder.base_model_prefix
    )
except TypeError:
    # Compatibility with transformers >=4.40.0
    PreTrainedModel._tie_encoder_decoder_weights(
        model[0].auto_model, self.decoder._modules[decoder_base_model_prefix], self.decoder.base_model_prefix, encoder_name_or_path
    )
```

arun477: Thanks! @tomaarsen makes sense. It has been updated and now works fine!

tomaarsen:
Thanks a bunch!

Issue #2624, opened 2024-04-29
Few Questions about sbert cross encoder & CEBinaryClassificationEvaluator


duarteharris:
First of all, let me share my appreciation for this library. It's been great to use so far and I really appreciate it.

I've been looking at the [Cross Encoders section](https://www.sbert.net/examples/applications/cross-encoder/README.html) and used the CrossEncoder example for the [Quora Duplicate Questions Detection task](https://github.com/UKPLab/sentence-transformers/blob/master/examples/training/cross-encoder/training_quora_duplicate_questions.py) to build my own model (which shows great eval metrics, thanks!).

A few questions emerged and I haven't been able to find answers for them:

1. During training, using the CEBinaryClassificationEvaluator we see something like:

CEBinaryClassificationEvaluator: Evaluating the model on dev dataset in epoch 1 after 60000 steps:
Accuracy:           97.15	(Threshold: 0.8388)
F1:                 84.52	(Threshold: 0.7532)
Precision:          86.67
Recall:             82.48
Average Precision:  91.55

What are these Thresholds we see for the Accuracy and the F1 Score? What do they mean? In my case, for the F1 Score they seem to go all over the place (in one it can show the 0.7532 you see there and the next show 0.1102 and just keep jumping around). What information should I glean from this?

2. What is the training optimising for? Accuracy? Loss? Something else? Is there a way to choose?

3. Can we choose other metrics, such as Fbeta Score, Matthews Correlation Coefficient  and/or Error Rate and add  them to the eval?

4. On the training of the CrossEncoders it reads: "For binary tasks and tasks with continuous scores (like STS), we set num_labels=1. For classification tasks, we set it to the number of labels we have.". So: if I'm understanding correctly, if I want the model.predict to return not a value between 0 and 1, but rather the integer 0 or 1 (in case of a binary problem) I need to set up num_labels=2, correct? 

Thank you for your attention in advance and thank you once again for the library,

ir2718:
Hi,

I'm going to try to answer your questions as best as I can.

> What are these Thresholds we see for the Accuracy and the F1 Score? What do they mean? In my case, for the F1 Score they seem to go all over the place (in one it can show the 0.7532 you see there and the next show 0.1102 and just keep jumping around). What information should I glean from this?

[...]

PR #2628, opened 2024-05-05
Choose which metric to optimize for `CrossEncoder`

ir2718:
Inspired by issue https://github.com/UKPLab/sentence-transformers/issues/2624 and Huggingface Trainer's `metric_for_best_model`, I thought it would be a nice addition to choose which metric an evaluator should optimize.

## PR Overview

The PR adds functionality for choosing the optimized metric for all `CrossEncoder.evaluation` classes which calculate multiple metrics.

## Details

*  Addition of an enum for all metrics
*  Addition of `return_metric` parameter to `CEBinaryClassificationEvaluator`
*  Addition of `return_metric` parameter to `CECorrelationEvaluator`
*  Addition of `return_metric` parameter to `CERerankingEvaluator`

### DIFF
--- a/sentence_transformers/cross_encoder/Metrics.py
+++ b/sentence_transformers/cross_encoder/Metrics.py
@@ -1,0 +1,13 @@
+from enum import Enum
+
+
+class Metric(Enum):
+    AVERAGE_PRECISION = "average_precision"
+    ACCURACY = "accuracy"
+    F1_SCORE = "f1_score"
+    PRECISION = "precision"
+    RECALL = "recall"
+    PEARSON = "pearson"
+    SPEARMAN = "spearman"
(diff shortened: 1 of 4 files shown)


l4b4r4b4b4:
related to https://github.com/UKPLab/sentence-transformers/issues/2737 ?

ir2718:
I'm pretty sure this PR won't be relevant anymore, since the CrossEncoder will also be refactored

tomaarsen:
Hello!

That is right, I think I will not move in this direction, but instead go for a similar approach as with the non-CE evaluators: have them return dictionaries instead. I'll close this now, and I hope to find time to refactor the CE training soon.
Thank you for taking the time to make this, @ir2718 

- Tom Aarsen

PR #2630, opened 2024-05-06
Enable Sentence Transformer Inference with Intel Gaudi2 GPU Supported ( 'hpu' ) - Follow up for #2557

ZhengHongming888:
This PR belongs to one of enabling Intel's Gaudi2 GPU supported tasks for Sentence Transformer's inference/training.

This is the follow up PR to https://github.com/UKPLab/sentence-transformers/pull/2557. There are few following considerations/modifications in this new PR on hpu device -

1. new padding algorithm added for hpu device after tokenizer (still padding=True) to give better performance

Here we give one new algorithm to add more padding to align the maximum batch length requirement for hpu graph mode. 
The algorithm is to add a little padding based on original features output from tokenizer based on the following padding calculation. 
     additional_pad_len = 2 ** math.ceil(math.log2(curr_tokenize_len[1])) - curr_tokenize_len[1]

The maximum batch length will be align into few values like [ 2, 4, 8, 16, 32, 64, 128, 256 .. ] which will greatly reduce the tokenizer/forward/cpu_conversion time under hpu graph mode. 

We have tested the performance over all 37 models listed in [pretrained models page ](https://www.sbert.net/docs/pretrained_models.html)and compared the performance between cuda A100 and hpu device.  Based on the [eval](https://github.com/UKPLab/sentence-transformers/blob/master/examples/evaluation/evaluation_inference_speed.py) code over 100000 sentences over the batch size of [32, 128, 512] the new padding for hpu device shows great performance. 

2. hpu support for PR https://github.com/UKPLab/sentence-transformers/pull/2573

Test this PR on hpu device/ failed. So we proposed the revision version to support this PR with hpu device. So now the test case of this PR (# test_encode_truncate) under tests/test_sentence_transformer.py also successfully passed.

3. hpu support for PR https://github.com/UKPLab/sentence-transformers/pull/2599

Test this PR on hpu device/ failed. So we proposed the revision version to support this PR with hpu device. So now the test case of this PR (# test_simple_encode ) under tests/test_image_embeddings.py also successfully passed

Welcome for any questions/comments!

Thanks.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -116,8 +116,9 @@ class SentenceTransformer(nn.Sequential):
             device = get_device_name()
             logger.info("Use pytorch device_name: {}".format(device))
 
-        if device=="hpu": 
+        if device == "hpu":
             from optimum.habana.transformers.modeling_utils import adapt_transformers_to_gaudi
+
             adapt_transformers_to_gaudi()
 
         if model_name_or_path is not None and model_name_or_path != "":
@@ -320,7 +321,6 @@ class SentenceTransformer(nn.Sequential):
 
             ht.hpu.wrap_in_hpu_graph(self, disable_tensor_cache=True)
             self.is_hpu_graph_enabled = True
-
 
         self.eval()
         if show_progress_bar is None:
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -119,6 +119,7 @@ class SentenceTransformer(nn.Sequential):
 
         if device == "hpu" and importlib.util.find_spec("optimum") is not None:
             from optimum.habana.transformers.modeling_utils import adapt_transformers_to_gaudi
+
             adapt_transformers_to_gaudi()
 
         if model_name_or_path is not None and model_name_or_path != "":
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -319,6 +319,7 @@ class SentenceTransformer(nn.Sequential):
             self.is_hpu_graph_enabled = True
 
             from optimum.habana.transformers.modeling_utils import adapt_transformers_to_gaudi
+
             adapt_transformers_to_gaudi()
 
         self.eval()
@@ -381,13 +382,31 @@ class SentenceTransformer(nn.Sequential):
             sentences_batch = sentences_sorted[start_index : start_index + batch_size]
             features = self.tokenize(sentences_batch)
             if self.device.type == "hpu":
-                if 'input_ids' in features:
-                    curr_tokenize_len = features['input_ids'].shape
+                if "input_ids" in features:
+                    curr_tokenize_len = features["input_ids"].shape
                     additional_pad_len = 2 ** math.ceil(math.log2(curr_tokenize_len[1])) - curr_tokenize_len[1]
-                    features['input_ids'] = torch.cat((features['input_ids'], torch.ones((curr_tokenize_len[0],additional_pad_len), dtype=torch.int8)), -1)
-                    features['attention_mask'] = torch.cat((features['attention_mask'],  torch.zeros((curr_tokenize_len[0],additional_pad_len), dtype=torch.int8)), -1)
-                    if 'token_type_ids' in features:
-                        features['token_type_ids'] = torch.cat((features['token_type_ids'],  torch.zeros((curr_tokenize_len[0],additional_pad_len), dtype=torch.int8)), -1)
+                    features["input_ids"] = torch.cat(
+                        (
+                            features["input_ids"],
+                            torch.ones((curr_tokenize_len[0], additional_pad_len), dtype=torch.int8),
+                        ),
+                        -1,
+                    )
+                    features["attention_mask"] = torch.cat(
+                        (
+                            features["attention_mask"],
+                            torch.zeros((curr_tokenize_len[0], additional_pad_len), dtype=torch.int8),
+                        ),
+                        -1,
+                    )
+                    if "token_type_ids" in features:
+                        features["token_type_ids"] = torch.cat(
+                            (
+                                features["token_type_ids"],
+                                torch.zeros((curr_tokenize_len[0], additional_pad_len), dtype=torch.int8),
+                            ),
+                            -1,
+                        )
 
             features = batch_to_device(features, device)
             features.update(extra_features)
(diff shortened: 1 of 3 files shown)


ZhengHongming888:
@tomaarsen Hi Tom,  could you take time to review this PR/merge it? Thanks much!

tomaarsen:
@ZhengHongming888 Yes, I will review it today. Perhaps I can also merge it today, but I don't think I can bring out a release with it today. Is that okay?

ZhengHongming888:
@tomaarsen  That's already very good! haha. :-) Thanks so much for your great support!

sentence_transformers/models/CLIPModel.py
```diff
@@ -72,7 +73,11 @@ def tokenize(self, texts, padding: Union[str, bool] = True):
             encoding["pixel_values"] = image_features.pixel_values
 
         encoding["image_text_info"] = image_text_info
-        return encoding
+        device = get_device_name()
+        if device == "hpu":
+            return dict(encoding)
+        else:
+            return encoding
```
tomaarsen, line 80:
```suggestion
        return dict(encoding)
```
I think it should be fine to just return a dictionary here regardless, rather than the BatchEncoding instance.

sentence_transformers/models/CLIPModel.py
```diff
@@ -3,6 +3,7 @@
 import transformers
 import torch
 from PIL import Image
+from sentence_transformers.util import get_device_name
```
tomaarsen, line 6:
```suggestion
```
No longer necessary since simplifying the dict() call below

ZhengHongming888:
@tomaarsen  everything you updated seems good for me! .. Thanks for your review!

ZhengHongming888:
@tomaarsen if everything is ok could you merge it by today? :-) Thanks!

tomaarsen:
@ZhengHongming888 I won't merge it today, I'm afraid. I've realised that 
* #2578 requires users to pass options to the model,
* #2630 requires users to pass options to the tokenizer,
* #2634 requires users to pass options to the config.

I think the best solution is sadly to create three `__init__` arguments: `model_kwargs`, `tokenizer_kwargs`, and `config_kwargs`. I won't merge this PR as-is because it adds the `padding` parameter which I'd like to avoid. Most likely I'll add the `kwargs` parameters in #2578 and discuss it with @satyamk7054 as he might have some good insight as well. 
There are also a lot of public holidays this time of year, so I'm slightly less available too.

- Tom Aarsen

ZhengHongming888:
@tomaarsen  but if i remove the request for padding argument in the __init__ part i can do the specific code for hpu side could it be merged for my PR tomorrow? Thanks.

tomaarsen:
@ZhengHongming888 I think so, yes. Feel free to make those changes.

- Tom Aarsen

ZhengHongming888:
@tomaarsen  I removed the request for adding padding as one argument in init and also passed all checks. :-) Now it should be ok. So please check and if any further please feel free to let me know. Thanks !

sentence_transformers/SentenceTransformer.py
```diff
@@ -111,6 +112,15 @@ def __init__(
         if cache_folder is None:
             cache_folder = os.getenv("SENTENCE_TRANSFORMERS_HOME")
 
+        if device is None:
+            device = get_device_name()
+            logger.info("Use pytorch device_name: {}".format(device))
+
+        if device == "hpu":
```
tomaarsen, line 119: Can we be sure that `optimum-habana` is installed here?

sentence_transformers/SentenceTransformer.py
```diff
@@ -374,11 +381,42 @@ def encode(
...
             features = batch_to_device(features, device)
             features.update(extra_features)
 
             with torch.no_grad():
-                out_features = self.forward(features)
+                if self.device.type == "hpu":
+                    hpu_graph_out = self.forward(features)
+                    out_features = copy.deepcopy(hpu_graph_out)
+                else:
+                    out_features = self.forward(features)
```
tomaarsen, line 419:
```suggestion
                out_features = self.forward(features)
                if self.device.type == "hpu":
                    out_features = copy.deepcopy(out_features)
```
I would prefer this, could you verify if that works? Admittedly, I'm not sure why you have to copy these.

sentence_transformers/SentenceTransformer.py
```diff
@@ -594,9 +632,6 @@ def tokenize(self, texts: Union[List[str], List[Dict], List[Tuple[str, str]]]):
         Tokenizes the texts
         """
         kwargs = {}
-        # HPU models reach optimal performance if the padding is not dynamic
-        if self.device.type == "hpu":
-            kwargs["padding"] = "max_length"
```
tomaarsen, line 599: Maybe it's best to keep the existing behaviour for now? Unless "max_length" is worse than `padding=True`, but I doubt that.

ZhengHongming888: No need for this now because we are proposing to use new additional padding to align the batch length into few power2 levels like [4, 8, 16, 32, 128 ..] still under the 'padding=True'  setting. It's kind of 'max_length' strategy used here to tradeoff between padding overhead and hpu alignment requirement.

sentence_transformers/SentenceTransformer.py
ZhengHongming888: In order to support the PR (test_encode_truncate() under tests/test_sentence_transformer) https://github.com/UKPLab/sentence-transformers/pull/2573,  we need the code here. The reason is in original implementation of out_features = self.forward(features) the variables of features/out_features are using one single memory and out_features just adding additional key ("sentence_embedding"). When later on done with Dim_truncate there will be some error in hpu's graph mode because graph mode will take the original output as one reference. 

I think what you changed here is good for simplicity.

sentence_transformers/SentenceTransformer.py
ZhengHongming888: good suggestion. i will add the check. thanks.

ZhengHongming888:
@tomaarsen thanks much for your suggestions! All are good for me and I have done to add the 'optimum-habana' check and also deepcopy part after forward output for hpu. All checks passed. Thanks for your some holiday time. :-) Sorry for that.

sentence_transformers/SentenceTransformer.py
tomaarsen: Makes total sense!

tomaarsen:
I have made one final modification to the code in https://github.com/UKPLab/sentence-transformers/pull/2630/commits/cc8b7e9c6647d2c65d06c7eae910de2c8949c617, because kwargs is now always empty. I think this is ready now.

Thanks for implementating my suggestions.

- Tom Aarsen

ZhengHongming888:
@tomaarsen  yes you are right for kwargs. Thanks for your great support on the PRs! Also thanks for your time! :-)

PR #2654, opened 2024-05-17
[`v3`] Resolve inplace modification error in DDP

tomaarsen:
See https://github.com/UKPLab/sentence-transformers/pull/2449#issuecomment-2114787400

Hello!

## Pull Request overview
* Override the loss model on every call, rather than just once. This seems necessary for DDP.

## Details

@Jakobhenningjensen would you be able to rerun your experiment with this branch instead?
i.e.
```
pip install git+https://github.com/tomaarsen/sentence-transformers.git@v3/fix_ddp_inplace_op
```
or 
```
git clone https://github.com/tomaarsen/sentence-transformers
cd sentence-transformers
git checkout v3/fix_ddp_inplace_op
```
or related commands.

I'll run this on my multi-GPU setup as well.

- Tom Aarsen

### DIFF
--- a/sentence_transformers/models/Pooling.py
+++ b/sentence_transformers/models/Pooling.py
@@ -209,7 +209,7 @@ class Pooling(nn.Module):
             output_vectors.append(embedding)
 
         output_vector = torch.cat(output_vectors, 1)
-        features.update({"sentence_embedding": output_vector})
+        features["sentence_embedding"] = output_vector
         return features
 
     def get_sentence_embedding_dimension(self):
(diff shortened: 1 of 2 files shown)


tomaarsen:
For me this works on single GPU, DP and DDP, so I'm merging it into v3.0-pre-release, feel free to try with that branch @Jakobhenningjensen.

- Tom Aarsen

Jakobhenningjensen:
@tomaarsen 
Unfortunately I still get the same error.

I think the "sentence-transformer" part which fails is the call `reps = [self.model(sentence_feature)["sentence_embedding"] for sentence_feature in sentence_features]` in `/usr/local/lib/python3.11/dist-packages/sentence_transformers/losses/MultipleNegativesRankingLoss.py`

and the actual call which fails might be `torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)` in `usr/local/lib/python3.11/dist-packages/torch/nn/functional.py` .

The stacktrace is a bit badly formatted (and difficult to extract), but that would be my guess. What it means tho, I have no idea of, unfortunately.

tomaarsen:
I'll look into it further, thanks!

PR #2663, opened 2024-05-22
[`v3`] Set `broadcast_buffers = False` when training with DDP

tomaarsen:
@Jakobhenningjensen

Hello!

## Pull Request overview
* Set `broadcast_buffers = False` when training with DDP to avoid `RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation.`

## Details
This PR *should* resolve the issue from https://github.com/UKPLab/sentence-transformers/pull/2654#issuecomment-2123849962. First of all, I was able to reproduce @Jakobhenningjensen his issue by also finetuning a bert-based model, rather than the MPNet-based model that I was using in my previous tests. After a full morning of experimentation, I found https://github.com/huggingface/transformers/issues/24996 which showed that the `buffers` as they are used in `transformers` do not play nicely with the default `broadcast_buffers` option in `DistributedDataParallel`. Following a recommendedation by https://github.com/huggingface/transformers/issues/24996#issuecomment-1956907209 (Thanks @nguyentanthong!), I was able to get training working with bert-based models.

Apologies for continuously asking this; but could you experiment with these changes @Jakobhenningjensen?

## Reproduction
I used this script to perform the training. Before this PR, it would fail:
```python
from sentence_transformers import SentenceTransformer, SentenceTransformerTrainingArguments, losses, SentenceTransformerTrainer

def train_model():
    model = SentenceTransformer("intfloat/multilingual-e5-small", device="cuda")
    loss = losses.MultipleNegativesRankingLoss(model)
        # Required parameter:
        output_dir="./sbert_fitted/",
        num_train_epochs=1,
        per_device_eval_batch_size=32,
        warmup_ratio=0.1,
        bf16=False,  # Set to True if you have a GPU that supports BF16
        eval_steps=100,
        save_strategy="steps",
        save_steps=100,
        save_total_limit=2,
        logging_steps=100,
        run_name="sts",  # Will be used in W&B if `wandb` is installed,
        report_to="none",
        dataloader_drop_last=True, # <- This seems necessary to prevent halting at the end of the training
    )
    train_dataset = load_dataset("sentence-transformers/all-nli", "triplet", split="train")

        model=model,
        args=training_args,
        train_dataset=train_dataset,
        loss=loss,
    )
    print("Training ...")
    trainer.train()


if __name__=="__main__":
    train_model()
```

- Tom Aarsen

### DIFF
--- a/sentence_transformers/training_args.py
+++ b/sentence_transformers/training_args.py
@@ -41,3 +41,7 @@ class SentenceTransformerTrainingArguments(TransformersTrainingArguments):
         # The `compute_loss` method in `SentenceTransformerTrainer` is overridden to only compute the prediction loss,
         # so we set `prediction_loss_only` to `True` here to avoid
         self.prediction_loss_only = True
+
+        # Disable broadcasting of buffers to avoid `RuntimeError: one of the variables needed for gradient computation
+        # has been modified by an inplace operation.` when training with DDP & a BertModel-based model.
+        self.ddp_broadcast_buffers = False


Jakobhenningjensen:
Fantastic! That seems to do the trick - the training is going on now :-)

I do get a funny progress-bar though (computing widget example) - I assume it might be something for the plotting of the loss/evaluation score.
image


Thanks a bunch @tomaarsen !

tomaarsen:
Are you running the training loop with
```
torchrun --nproc_per_node=2 train_script.py
```
(assuming 2 GPUs)

A preview of some of the upcoming Distributed Training docs:
image

As for the widget example: that is preparing some (hopefully) useful inference examples to put in the model card. You can disable it with:
```python
from sentence_transformers import SentenceTransformer, SentenceTransformerModelCardData

model = SentenceTransformer("intfloat/multilingual-e5-small", device="cuda", model_card_data=SentenceTransformerModelCardData(generate_widget_examples=False))
```

- Tom Aarsen

Jakobhenningjensen:
>As for the widget example: that is preparing some (hopefully) useful inference examples to put in the model card. You can >disable it with:

>from sentence_transformers import SentenceTransformer, SentenceTransformerModelCardData

>model = SentenceTransformer("intfloat/multilingual-e5-small", device="cuda", >model_card_data=SentenceTransformerModelCardData(generate_widget_examples=False))

@tomaarsen thanks, that works!

An idea might be to check, if "private=True" and then not automatically upload prediction samples. I just figured out that it happend - luckily it was on some dummy data but say you train on some confidential data, then it might lead to some issues that some of the data is automatically uploaded in the model-card

(btw, how do you disabling the upload of examples? I would assume it is `predict_example=False` in `SentenceTransformerModelCardData` but i'm not quite sure

tomaarsen:
Hello!

Right now the samples are always included, no option to disable them. If you're training on sensitive data, then my recommendation is to save the model locally with `model.save(...)` or `model.save_pretrained(...)`, read the model card & remove all data that can't be uploaded, then load the saved model again and call `push_to_hub` directly after loading the model.
If you don't pass the model to a SentenceTransformerTrainer, the original README.md is used as the model card, so it will upload the model card with your manual fixes.

That said, I'll mark this on my TODO list for a 3.1 release, it does seem rather useful.

- Tom Aarsen

Jakobhenningjensen:
Thanks for the suggestion! The automated card is fantastic though!

Jakobhenningjensen:
I was wondering (please let me know if you prefer opening such questions as issues); when a model uses BatchNorm, is that  automatically transformed (no pun intended) to `SyncBatchNorm` or do we need to do such things manually before training the model, when training using DDP?

tomaarsen:
I think this would need to be done manually, like so: https://github.com/huggingface/accelerate/issues/791. The `accelerator.prepare` is done internally on the SentenceTransformer model, but the `nn.SyncBatchNorm.convert_sync_batchnorm` call is never done.
(I wasn't aware that `SyncBatchNorm` even existed 😅)

- Tom Aarsen

PR #2676, opened 2024-05-27
[`v3`] Move training dependencies into a "train" extra

tomaarsen:
Hello!

## Pull Request overview
* Move training dependencies into a "train" extra

## Details
Training/finetuning now requires you to install
```
pip install sentence_transformers[train]
```
rather than just `sentence_transformers`. This is to avoid 2 additional dependencies (`datasets`, `accelerate`) from being required when you're only interested in inference.

The changes are primarily moving `datasets` imports behind `is_datasets_available()`, converting e.g. `Dataset` type hints into `"Dataset"`, and updating the docs.

cc @osanseviero 

- Tom Aarsen

### DIFF
--- a/sentence_transformers/fit_mixin.py
+++ b/sentence_transformers/fit_mixin.py
@@ -14,8 +14,8 @@ from torch.optim import Optimizer
 from torch.utils.data import DataLoader
 from tqdm.autonotebook import trange
 from transformers import TrainerCallback, TrainerControl, TrainerState
+from transformers.utils import is_datasets_available
 
-from datasets import Dataset, DatasetDict
 from sentence_transformers.datasets.NoDuplicatesDataLoader import NoDuplicatesDataLoader
 from sentence_transformers.datasets.SentenceLabelDataset import SentenceLabelDataset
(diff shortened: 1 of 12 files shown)


sentence_transformers/fit_mixin.py
```diff
@@ -14,8 +14,8 @@
 from torch.utils.data import DataLoader
 from tqdm.autonotebook import trange
 from transformers import TrainerCallback, TrainerControl, TrainerState
+from transformers.utils import is_datasets_available
```
osanseviero, line 17: (no strong opinion)
I would likely reimplement this here rather than import a utility from an external library. Given you're not using the version, you can have a simple implementation

```py
import importlib.util

spec = importlib.util.find_spec("datasets")
return spec is not None

tomaarsen: Sounds good. I was already thinking about combining `is_datasets_available` and `is_accelerate_available` into an `is_training_available` as I have to use both for some checks in several locations, so it makes sense to implement it all locally.

PR #2690, opened 2024-05-29
fix loss link

Samoed:
Currently, this links leads to losses and loss_overview, but they should to losses.html and loss_overview.html
image


tomaarsen (approved):
Thanks a bunch for spotting this!

tomaarsen:
I've refreshed the documentation, it should be live in an hour or two max (due to caches).

- Tom Aarsen

PR #2700, opened 2024-05-31
Fix a bug in SequentialEvaluator to use primary_metric if defined in evaluator.

hetulvp:
Fixed `SequentialEvaluator` to use `primary_metric` if defined. Before this change, it was using `evaluation` instead of `evaluator`.


tomaarsen:
Hello!

Very well spotted, once again. I appreciate these fixes.

- Tom Aarsen

PR #2701, opened 2024-05-31
Fix bug that restricts users from specifying custom primary_function in InformationRetrievalEvaluator

hetulvp:
This change allows users to specify `primary_metric` without being overwritten during the evaluation call.

Currently the `primary_function` is always overwritten by the `__call__` method.


tomaarsen:
Hello!

This is much better indeed. Thanks for this PR! I'll merge this once the tests go green.

- Tom Aarsen

PR #2709, opened 2024-06-04
[`fix`] Always override the originally saved __version__ in the ST config

tomaarsen:
Hello!

## Pull Request overview
* Always override the originally saved __version__ in the ST config

## Details
Currently, the `__version__` data is only determines once, and then all future models finetuned on that model will just adopt the original versions. This can become a problem as e.g. ST v3.0 will save models that have e.g. "2.2.2" in their configuration, which is rather confusing.

## Example
E.g. https://huggingface.co/ammumadhu/indic-bert-nli-matryoshka/blob/main/config_sentence_transformers.json
which shows 2.2.2 whereas the model itself was trained with v3.0.0.

cc @osanseviero 

- Tom Aarsen

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -1004,12 +1004,11 @@ class SentenceTransformer(nn.Sequential, FitMixin):
         modules_config = []
 
         # Save some model info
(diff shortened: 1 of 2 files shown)


osanseviero (approved):
Nice find!

PR #2734, opened 2024-06-09
[Spelling] update examples

michaelfeil:


tomaarsen:
Hello!

Thanks for this. I'm not sure why the images show up as both deleted & added, but it seems fine to me. 

- Tom Aarsen

PR #2735, opened 2024-06-09
spelling: code comment updates

michaelfeil:
This should only be comments, not code.


tomaarsen:
Hello!

Thank you for these!

- Tom Aarsen

PR #2744, opened 2024-06-11
Make Cleaning Of Strings In NoDuplicatesDataLoader Optional

rovermicrover:
With very large datasets with a high percentage of duplicates you will some times need to sys.intern the strings to save memory by only having each uniq string in memory once.

Currently the NoDuplicatesDataLoader calls strip and lower on all the strings as part of the dedupe process which then means a new string is added to memory.

To this end we propose adding an option to NoDuplicatesDataLoader that will allow you to skip the text cleaning portion of the DataLoader to ensure that multiples of the same string are not pulled into memory.

Default behavior will be to clean the strings to ensure backwards compatibility.

### DIFF
--- a/sentence_transformers/datasets/NoDuplicatesDataLoader.py
+++ b/sentence_transformers/datasets/NoDuplicatesDataLoader.py
@@ -3,7 +3,7 @@ import random
 
 
 class NoDuplicatesDataLoader:
-    def __init__(self, train_examples, batch_size):
+    def __init__(self, train_examples, batch_size, clean=True):
         """
         A special data loader to be used with MultipleNegativesRankingLoss.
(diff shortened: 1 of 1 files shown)


rovermicrover:
We just upgraded to 3.X and realized that this class has been deprecated in the process. Should this be closed out then?

tomaarsen:
Hello!

Indeed, this class will no longer be commonly used (only for backwards compatibility), and is effectively deprecated. Your desired behavior should now be the default in the [NO_DUPLICATES Batch Sampler](https://sbert.net/docs/package_reference/sentence_transformer/training_args.html#batchsamplers), if I remember correctly.

Feel free to let me know if that batch sampler works well for you.

I'll close this, but I do want to thank you for taking the time to make this.

- Tom Aarsen

PR #2747, opened 2024-06-12
Update DenoisingAutoEncoderDataset.py

sophia8844:
from nltk.tokenize.treebank import TreebankWordDetokenizer instead of nltk


tomaarsen:
Hello!

This seems good to me, it does seem better to be more explicit.

- Tom Aarsen

PR #2761, opened 2024-06-18
[`typing`] Overload the Encode method

### LINKED ISSUE #2758: API type hinting not overloaded properly
Just wondering if this is a desirable thing, but there are a few methods that have type annotations that strictly correspond to an input. 

```
Returns:
    Union[List[Tensor], ndarray, Tensor]: By default, a 2d numpy array with shape [num_inputs, output_dimension] is 
returned. If only one string input is provided, then the output is a 1d array with shape [output_dimension]. If 
convert_to_tensor, a torch Tensor is returned instead. If self.truncate_dim <= output_dimension then output_dimension is 
self.truncate_dim.
```

This could be specified using [overload](https://docs.python.org/3/library/typing.html)

tomaarsen:
Resolves #2758

Hello!

## Pull Request overview
* Overload the Encode method

### Current
image
image

### New
340583794-6510131f-2c90-483f-8f1e-5aac432f3eb5
340583957-e31f8cbf-04e3-4d67-b445-b2c72de534bd

I'm not really sure if one is better or clearer than the other.

- Tom Aarsen


tomaarsen:
Closing this in favor of #2763.

PR #2778, opened 2024-06-25
Elasticsearch example: simplify setup

maxjakob:
The instructions to run locally were simplified to omit TLS:
https://www.elastic.co/guide/en/elasticsearch/reference/current/run-elasticsearch-locally.html
Hence we can simplify this example script to skip the ssl_context part.

Also, the `hosts` can now be a single URL and don't have to be a keyword arg anymore (see [docs](https://elasticsearch-py.readthedocs.io/en/v8.14.0/#example-usage)).


tomaarsen:
Hello!

Excellent - thank you for taking the time to keep this up to date with the latest from ES! I'll try to run this myself to verify that it works.

- Tom Aarsen

tomaarsen:
Works wonderfully as long as you use ES v8.0+, otherwise the authentication gets ignored (I believe because you have to use `http_auth` instead of `basic_auth` for v7.0 or older).

Thank you for this!

- Tom Aarsen

PR #2783, opened 2024-06-26
Cast predict scores to float before converting to numpy

malteos:
Torch might throw a type error (see https://github.com/pytorch/pytorch/issues/109873) when a model uses a different precision (e.g., bf16). Casting the scores to float first fixes this.

I encountered this error when evaluating the recent Jina reranker (https://huggingface.co/jinaai/jina-reranker-v2-base-multilingual).


tomaarsen:
Hello!

I got the exact same behaviour myself! You beat me to the PR 😄 Thanks a bunch.

- Tom Aarsen

Issue #2782, opened 2024-06-26
GroupByLabelBatchSampler


waileong-leong:
In the GroupByLabelBatchSampler, what is the reason of doing (num_samples := len(sample_indices) // 2) to determine the groups? 

https://github.com/UKPLab/sentence-transformers/blob/2dee8c22f618df3f34f55d406bf97aa88383728a/sentence_transformers/sampler.py#L68-L72

if a label has 4 sentence {"0" : [0,1,2,3]}, it will only contain [0,1] in the group

fpgmaas:
Coincidentally, I made a small PR to improve that class today: https://github.com/UKPLab/sentence-transformers/pull/2788

Part of the PR is to remove these lines of code since I also was not sure why they were there.

tomaarsen:
Hello!

Ai, this is indeed a bug. The goal of this snippet of code is to ensure that there is an even number of text values under `groups`. The `__iter__` code relies on this (and that `batch_size % 2 == 0`) to ensure that each batch has 2+ texts from every label that occurs in the text. 
E.g. if you have 17 texts from one label and a batch size of 16, then `__iter__` would yield the 16 and keep the leftover 1 to use in the next batch. This can cause issues with losses where there must be at least 2 texts for that label.

I believe the code should instead be:
```python
        self.groups = {
            label: sample_indices[:num_samples]
            for label, sample_indices in groups.items()
            if (num_samples := len(sample_indices) // 2 * 2)
        }
```
This way we'll only throw away 1 sample at most per label.

- Tom Aarsen

fpgmaas:
This

> [...] to ensure that each batch has 2+ texts from every label that occurs in the text.

should be

> [...] to ensure that each batch has 2+ texts from every label that occurs in the **batch**

Right? Not trying to be pedantic, just trying to confirm that my understanding of the sampler is correct :) Then I could incorporate the above in [this PR](https://github.com/UKPLab/sentence-transformers/pull/2788). Although it can of course also be solved earlier in a separate PR that aims to fix just that.

tomaarsen:
> to ensure that each batch has 2+ texts from every label that occurs in the batch

Yes. That should be correct, indeed. The dataset can have labels and corresponding texts that do not get used in the batch, that should be fine. 
I'd love to incorporate the `groups` fix that I described above in your PR if possible, as I quite like the other changes that you're proposing @fpgmaas

- Tom Aarsen

waileong-leong:
[...]

PR #2788, opened 2024-06-27
[feat] Improve GroupByLabelBatchSampler

fpgmaas:
This PR does two things:

#### Fix the `GroupByLabelBatchSampler`

I believe this part of code to be incorrect;

```py
    for column_name in valid_label_columns or []:
        if column_name in dataset.column_names:
            labels = dataset["label"]
            break
    else:
        raise ValueError(f"None of the valid_label_columns {valid_label_columns} are in the dataset.")
```

It seems to try and always use the `"label"` column, where I expect that the purpose of this code is to find the first value of `valid_label_columns` that is a valid column in the dataset. This PR fixes that, and make some other small adjustments, for example removing the lines of code mentioned here: https://github.com/UKPLab/sentence-transformers/issues/2782

#### Add unit tests

Current test coverage for `sampler.py` is 31%, this bumps that up a little bit :) In addition, I think having this unit test can also make it more clear for future developers to understand what the class is aiming to do.


waileong-leong (approved):

sentence_transformers/sampler.py
```diff
@@ -34,6 +34,23 @@ class DefaultBatchSampler(SetEpochMixin, BatchSampler):
...
+        dataset (Dataset): The dataset to sample from.
+        batch_size (int): Number of samples per batch. Must be divisible by 2.
+        drop_last (bool): If True, drop the last incomplete batch, if the dataset size
+                          is not divisible by the batch size.
+        valid_label_columns (List[str]): List of column names to check for labels.
+                                         The first column name found in the dataset will
+                                         be used as the label column.
+        generator (torch.Generator, optional): Optional random number generator for shuffling
+                                               the indices.
+        seed (int, optional): Seed for the random number generator to ensure reproducibility.
```
tomaarsen, line 51:
```suggestion
    This sampler groups samples by their labels and aims to create batches such that
    each batch contains samples where the labels are as homogeneous as possible.
    This sampler is meant to be used alongside the `Batch...TripletLoss` classes, which
    require that each batch contains at least 2 examples per label class.

    Args:
        dataset (Dataset): The dataset to sample from.
        batch_size (int): Number of samples per batch. Must be divisible by 2.
        drop_last (bool): If True, drop the last incomplete batch, if the dataset size
            is not divisible by the batch size.
        valid_label_columns (List[str]): List of column names to check for labels.
            The first column name found in the dataset will
            be used as the label column.
        generator (torch.Generator, optional): Optional random number generator for shuffling
            the indices.
        seed (int, optional): Seed for the random number generator to ensure reproducibility.
```
Added some (sorely lacking) information about when this sampler should be used, and I updated the formatting to match the Google-style docstrings that I use throughout ST.

sentence_transformers/sampler.py
```diff
@@ -53,32 +70,28 @@ def __init__(
...
-                break
-        else:
-            raise ValueError(f"None of the valid_label_columns {valid_label_columns} are in the dataset.")
-
-        del dataset
+        labels = self._determine_labels_to_use(dataset, valid_label_columns)
         groups = defaultdict(list)
         for sample_idx, label in enumerate(labels):
             groups[label].append(sample_idx)
+        self.groups = groups
```
tomaarsen, line 78:
```suggestion
        self.groups = {
            label: sample_indices[:num_samples]
            for label, sample_indices in groups.items()
            if (num_samples := len(sample_indices) // 2 * 2)
        }
```

See https://github.com/UKPLab/sentence-transformers/issues/2782#issuecomment-2196579410

Ideally, this should be tested as well. e.g.:
```python
@pytest.fixture
def dummy_uneven_dataset():
    """
    Dummy dataset for testing purposes. The dataset looks as follows:
    {
        "data": ["a"] * 51,
        "label": [0] * 17 + [1] * 17 + [2] * 17,
    }
    """
    data = {"data": ["a"] * 51, "label": [0] * 17 + [1] * 17 + [2] * 17}
    return Dataset.from_dict(data)
```
with a `GroupByLabelBatchSampler` with batch size of 16. Each batch should still always have 2+ texts for each label.

tomaarsen (changes requested):
Thanks for this! I do think I'll need changes regarding the `self.groups` as discussed in that comment itself, but otherwise this is heading in the right direction!

sentence_transformers/sampler.py
```diff
@@ -53,32 +72,31 @@ def __init__(
...
+        labels = self._determine_labels_to_use(dataset, valid_label_columns)
         groups = defaultdict(list)
         for sample_idx, label in enumerate(labels):
             groups[label].append(sample_idx)
 
         self.groups = {
-            label: sample_indices[:num_samples]
-            for label, sample_indices in groups.items()
-            if (num_samples := len(sample_indices) // 2)
+            label: sample_indices[: len(sample_indices) // 2 * 2] for label, sample_indices in groups.items()
```
tomaarsen, line 81: I believe this optimization isn't ideal, as `sample_indices[: len(sample_indices) // 2 * 2]` can now be an empty list. Though it doesn't matter much (it only costs an unnecessary iteration in the `__iter__`), it might be best to keep the `if`. I put that filter in there to avoid this edge case

fpgmaas: Valid point! I changed it back to the way you proposed.

sentence_transformers/sampler.py
```diff
@@ -53,32 +72,33 @@ def __init__(
...
+            if (num_samples := len(sample_indices) // 2 * 2)
         }
 
+    @staticmethod
+    def _determine_labels_to_use(dataset: "Dataset", valid_label_columns: List[str]) -> List[Any]:
+        for column_name in valid_label_columns or []:
+            if column_name in dataset.column_names:
+                return dataset[column_name]
+        else:
+            raise ValueError(f"None of the valid_label_columns {valid_label_columns} are in the dataset.")
```
tomaarsen, line 92: Now that you've moved this into a separate method, I think we can simplify this a bit further by removing the `else`. I've also expanded the error to include the column names, as that can be helpful.
```suggestion
        raise ValueError(
            f"None of the valid_label_columns {valid_label_columns} are in the dataset, "
            f"which only has these columns: {dataset.column_names}."
        )
```

tomaarsen (changes requested):
Small nitpick regarding the error

tomaarsen (approved):
Thanks a bunch for taking care of my review comments. After some nitpicky commit, I think this is all good to go (at least, once the tests go green)!

- Tom Aarsen

PR #2789, opened 2024-06-27
[chore] Enable ruff rules `Warning (W)`

fpgmaas:
This PR proposes to enable ruff rules [`Warning (W)`](https://docs.astral.sh/ruff/rules/#warning-w). The others in the list were already enabled, since they are listed under the default value, see [here](https://docs.astral.sh/ruff/settings/#lint_select)

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,8 +1,11 @@
 
 line-length = 119
 
+[lint]
 # Skip `E731` (do not assign a lambda expression, use a def)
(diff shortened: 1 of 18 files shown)


tomaarsen:
Hello!

This seems great; thank you! I'm a bit surprised that this wasn't already enabled to be honest.
I'm merging this quickly to avoid merge conflicts on this PR.

- Tom Aarsen

PR #2790, opened 2024-06-28
[chore] improve ruff config

fpgmaas:
Small PR to improve the ruff configuration; just enable all `E` rules except `E501`, add `I`. The code already seems to adhere to the `isort` rules though. Not sure how that happened, since `I` is not shown in the [list of defaults](https://docs.astral.sh/ruff/settings/#lint_select)?

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -2,10 +2,14 @@
 line-length = 119
 
 [lint]
-# Skip `E731` (do not assign a lambda expression, use a def)
-ignore = ["E731"]
+ignore = [
+    # LineTooLong
+    "E501",
+    # DoNotAssignLambda
+    "E731",
+]
 
-select = ["E4", "E7", "E9", "F", "W"]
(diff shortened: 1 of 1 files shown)


tomaarsen:
@fpgmaas I ran `I` manually at one point. I had some issues there as the CI would constantly be red requiring changes, but locally everything was fine. I think the cause may have been the `datasets` library, and whether `import datasets` referred to that third-party project or this folder: https://github.com/UKPLab/sentence-transformers/tree/master/sentence_transformers/datasets

In the end I just did it manually once and kept it out of the CI, but I'm happy to reintroduce it now that `datasets` is a mandatory dependency of the `train` (and thus `dev`) extra.

- Tom Aarsen

fpgmaas:
Thanks for elaborating @tomaarsen! I encountered the same problem. I closed this because I didn't want to bother you with it yet until I figured that out :) Will reopen if I manage to solve it.

PR #2798, opened 2024-06-30
[chore] Add test for RoundrobinBatchSampler

fpgmaas:
This PR

- Makes the `seed` and `generator` arguments optional
- Adds unit tests for the `RoundRobinBatchSampler`.
- Have the class return a ValueError if the length of the `ConcaDataset` does not match the length of the list of `BatchSamplers`. 

The reason for the last point above, is that I ran into an issue while creating the unit tests. Initially I tried:

```py
import pytest
from datasets import Dataset
from sentence_transformers.sampler import RoundRobinBatchSampler

from torch.utils.data import BatchSampler, SequentialSampler, ConcatDataset


@pytest.fixture
def dummy_dataset() -> ConcatDataset:
    """
    Dummy dataset for testing purposes. The dataset looks as follows:
    {
        "data": [0, 1, 2, ... , 23, 24, 100, 101, ..., 123, 124],
        "label": [0, 1, 0, 1, ..., 0, 1],
    }
    """

    values_1 = list(range(25))
    labels = [x % 2 for x in values_1]
    dataset_1 = Dataset.from_dict({"data": values_1, "label": labels})

    values_2 = [x + 100 for x in values_1]
    dataset_2 = Dataset.from_dict({"data": values_2, "label": labels})

    return ConcatDataset([dataset_1, dataset_2])


def test_round_robin_batch_sampler(dummy_dataset):
    batch_sampler_1 = BatchSampler(SequentialSampler(range(len(dummy_dataset))), batch_size=4, drop_last=True)
    batch_sampler_2 = BatchSampler(SequentialSampler(range(len(dummy_dataset))), batch_size=4, drop_last=True)
    sampler = RoundRobinBatchSampler(dataset=dummy_dataset, batch_samplers=[batch_sampler_1, batch_sampler_2])

    batches = list(iter(sampler))

    for batch in batches:
        for idx in batch:
            print(dummy_dataset[idx])
```

But that failed with a `IndexError: list index out of range`. Since the sampler takes both a `ConcatDataset`and a list of `BatchSamplers`, I am assuming the intended way to use this class is to have a sampler for each dataset in the `ConcatDataset`, so I added that as a check in the class. But maybe my assumption here is wrong; if so please let me know.

### DIFF
--- a/sentence_transformers/sampler.py
+++ b/sentence_transformers/sampler.py
@@ -149,7 +149,9 @@ class NoDuplicatesBatchSampler(SetEpochMixin, BatchSampler):
 
 class RoundRobinBatchSampler(SetEpochMixin, BatchSampler):
     """
-    Batch sampler that yields batches in a round-robin fashion from multiple batch samplers.
+    Batch sampler that yields batches in a round-robin fashion from multiple batch samplers, until one is exhausted.
+    With this sampler, it's unlikely that all samples from each dataset are used, but we do ensure that each dataset
+    is sampled from equally.
 
     Args:
         dataset (ConcatDataset): A concatenation of multiple datasets.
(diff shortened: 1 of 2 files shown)


tomaarsen:
Looks great! Thanks for this!

- Tom Aarsen

PR #2799, opened 2024-07-01
[`chore`] Clean-up `.gitignore`

fpgmaas:
Clean up `.gitignore`, borrowing some parts from [here](https://raw.githubusercontent.com/github/gitignore/main/Python.gitignore).


tomaarsen:
Seems great! I don't see any files suddenly being included, so I think we're all set.

- Tom Aarsen

PR #2808, opened 2024-07-04
[`chore`] Add test for MatryoshkaLoss

fpgmaas:
`losses/MatryoshkaLoss.py` currently has 27% test coverage, this PR raises that to 70%.

### DIFF
--- a/tests/losses/test_MatryoshkaLoss.py
+++ b/tests/losses/test_MatryoshkaLoss.py
@@ -1,0 +1,37 @@
+import pytest
+import torch
+from sentence_transformers import SentenceTransformer
+from sentence_transformers.losses import MultipleNegativesRankingLoss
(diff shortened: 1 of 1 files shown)


tomaarsen:
The overall test here is a bit suboptimal, because MatryoshkaLoss *also* overrides the forward method, so the mocking prevents that from working normally:
image

Which is why these don't get used. With other words, I think the mocking is actually somewhat "breaking" the loss, as the ForwardDecorator is supposed to result in shrunk embedding dimensions being returned, but that isn't happening due to the mocking. I tried to play around with this a bit to change up what I'm mocking (e.g. trying to mock the underlying Transformer rather than the full SentenceTransformer class), but I couldn't get that to work well.

In short, I think I'm leaning towards not accepting this as-is.

- Tom Aarsen

PR #2810, opened 2024-07-04
Allow inheriting the Transformer class

mokha:
Creating a custom Transformer class that inherits Transformer would fail as the load function would always the original class. This change would allow smooth inheritance


tomaarsen (approved):
I agree completely, this is a much better convention for loading a class instance.

PR #2818, opened 2024-07-09
updated mine_hard_negatives method to include a seperate corpus for m…

ChrisGeishauser:
…ining hard negatives.

### DIFF
--- a/sentence_transformers/util.py
+++ b/sentence_transformers/util.py
@@ -663,6 +663,12 @@ def mine_hard_negatives(
     if not separate_corpus:
         corpus = positives
 
+    # Deduplicate the corpus
+    corpus = list(set(corpus))
+    pos_to_corpus_indices = torch.tensor(
+        [corpus.index(positive) if positive in corpus else -1 for positive in positives], dtype=torch.long
+    )
+
     # Embed the corpus, queries, and positives
     corpus_embeddings = model.encode(corpus, batch_size=batch_size, convert_to_tensor=True, show_progress_bar=True)
     query_embeddings = model.encode(queries, batch_size=batch_size, convert_to_tensor=True, show_progress_bar=True)
@@ -671,7 +677,7 @@ def mine_hard_negatives(
             positives, batch_size=batch_size, convert_to_tensor=True, show_progress_bar=True
         )
     else:
-        positives_embeddings = corpus_embeddings
+        positives_embeddings = corpus_embeddings[pos_to_corpus_indices]
     batch_idx = torch.arange(len(queries)).unsqueeze(-1)
 
     if use_faiss:
@@ -714,19 +720,7 @@ def mine_hard_negatives(
 
     # Scores is a [num_queries, range_max + 1] tensor, where we set the values to -inf to disqualify the corresponding
     # text as a negative candidate. Here we disqualify the positive pair
-
-    # positive_indices = indices == torch.arange(len(queries), device=indices.device).unsqueeze(-1)
-    # scores[positive_indices] = -float("inf")
-    positive_indices = []
-    for positive in positives:
-        if positive in corpus:
-            index = corpus.index(positive)
-            positive_indices.append(index)
-        else:
-            positive_indices.append(-1)
-    positive_indices = torch.Tensor(positive_indices).to(torch.int32).unsqueeze(-1)
-    positive_indices = indices == positive_indices
-
+    positive_indices = indices == pos_to_corpus_indices.unsqueeze(-1)
     scores[positive_indices] = -float("inf")
 
     num_candidates = scores.numel()
--- a/sentence_transformers/util.py
+++ b/sentence_transformers/util.py
@@ -515,7 +515,6 @@ def semantic_search(
     return queries_result_list
 
 
-
 def mine_hard_negatives(
     dataset: "Dataset",
     model: "SentenceTransformer",
@@ -669,12 +668,8 @@ def mine_hard_negatives(
         corpus = corpus[corpus_columns[0]]
 
     # Embed the corpus, queries, and positives
-    corpus_embeddings = model.encode(
-        corpus, batch_size=batch_size, convert_to_tensor=True, show_progress_bar=True
-    )
-    query_embeddings = model.encode(
-        queries, batch_size=batch_size, convert_to_tensor=True, show_progress_bar=True
-    )
+    corpus_embeddings = model.encode(corpus, batch_size=batch_size, convert_to_tensor=True, show_progress_bar=True)
+    query_embeddings = model.encode(queries, batch_size=batch_size, convert_to_tensor=True, show_progress_bar=True)
     positives_embeddings = model.encode(
         positives, batch_size=batch_size, convert_to_tensor=True, show_progress_bar=True
     )
@@ -684,9 +679,7 @@ def mine_hard_negatives(
         import faiss
 
         # Compute the positive scores separate from FAISS
-        positive_scores = model.similarity_pairwise(
-            query_embeddings, positives_embeddings
-        ).cpu()
+        positive_scores = model.similarity_pairwise(query_embeddings, positives_embeddings).cpu()
 
         query_embeddings = query_embeddings.cpu().numpy()
         corpus_embeddings = corpus_embeddings.cpu().numpy()
@@ -703,9 +696,7 @@ def mine_hard_negatives(
         scores_list = []
         indices_list = []
         # Iterate over query embeddings in batches so we can track the progress
-        for i in trange(
-            0, len(query_embeddings), batch_size, desc="Querying FAISS index"
-        ):
+        for i in trange(0, len(query_embeddings), batch_size, desc="Querying FAISS index"):
             query_chunk = query_embeddings[i : i + batch_size]
             scores, indices = index.search(query_chunk, k=range_max + 1)
             scores_list.append(scores)
@@ -715,9 +706,7 @@ def mine_hard_negatives(
     else:
         # Compute all similarity scores
         scores = model.similarity(query_embeddings, corpus_embeddings).cpu()
-        positive_scores = model.similarity_pairwise(
-            query_embeddings, positives_embeddings
-        ).cpu()
+        positive_scores = model.similarity_pairwise(query_embeddings, positives_embeddings).cpu()
 
         # Keep only the range_max + 1 highest scores. We offset by 1 to potentially include the positive pair
         scores, indices = torch.topk(scores, k=range_max + 1, dim=1)
@@ -745,9 +734,7 @@ def mine_hard_negatives(
 
     # Rescore with cross_encoder
     if cross_encoder is not None and (margin is not None or max_score is not None):
-        for idx, candidate_neg_idx in tqdm(
-            enumerate(indices), desc="Rescoring with CrossEncoder", total=len(indices)
-        ):
+        for idx, candidate_neg_idx in tqdm(enumerate(indices), desc="Rescoring with CrossEncoder", total=len(indices)):
             query = queries[idx]
             candidate_passages = [corpus[neg_idx] for neg_idx in candidate_neg_idx]
             pred_scores = cross_encoder.predict(
@@ -811,9 +798,7 @@ def mine_hard_negatives(
         num_options = indices.size(1) - negative_scores.isinf().sum(1)
         num_options = num_options.clamp(min=num_negatives)
         # Randomly sample negatives from each row
-        sampled_idx = [
-            random.sample(range(options), k=num_negatives) for options in num_options
-        ]
+        sampled_idx = [random.sample(range(options), k=num_negatives) for options in num_options]
         indices = indices[batch_idx, sampled_idx]
         negative_scores = negative_scores[batch_idx, sampled_idx]
         # Resort the indices and scores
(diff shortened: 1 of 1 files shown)


ChrisGeishauser:
Extended mine_hard_negatives method in sentence_transformers/util.py to enable the retrieval of negative document candidates from a seperate corpus instead of using the positives as corpus.

tomaarsen:
(Related to #2818)

Thanks for opening this! Will have a look at the details shortly.

- Tom Aarsen

sentence_transformers/util.py
```diff
@@ -655,18 +658,27 @@ def mine_hard_negatives(
...
-    # Embed the corpus and queries
+    # Embed the corpus, queries, and positives
     corpus_embeddings = model.encode(corpus, batch_size=batch_size, convert_to_tensor=True, show_progress_bar=True)
     query_embeddings = model.encode(queries, batch_size=batch_size, convert_to_tensor=True, show_progress_bar=True)
+    if separate_corpus:
+        positives_embeddings = model.encode(
+            positives, batch_size=batch_size, convert_to_tensor=True, show_progress_bar=True
+        )
+    else:
+        positives_embeddings = corpus_embeddings
```
tomaarsen, line 674: If we don't have a separate corpus, then we don't have to reembed the positives :)

sentence_transformers/util.py
```diff
@@ -655,18 +658,27 @@ def mine_hard_negatives(
 
     log_counters = {}
     queries = dataset[columns[0]]
-    corpus = dataset[columns[1]]
+    positives = dataset[columns[1]]
+    separate_corpus = corpus is not None
+    if not separate_corpus:
+        corpus = positives
```
tomaarsen, line 664: I've simplified this to just a list of strings. For `dataset` it makes sense for it to be a 2-column Dataset (in my opinion), as that is a common format that is often used to train embedding models, which can be easily loaded. See for example this collection: https://huggingface.co/collections/sentence-transformers/embedding-model-datasets-6644d7a3673a511914aa7552

A lot of these datasets have 2 columns.

But if you also have a collection of negative candidates, then it might be a hastle to turn that into a Dataset?

sentence_transformers/util.py
```diff
@@ -702,7 +714,19 @@ def mine_hard_negatives(
...
+    # scores[positive_indices] = -float("inf")
+    positive_indices = []
+    for positive in positives:
+        if positive in corpus:
+            index = corpus.index(positive)
+            positive_indices.append(index)
+        else:
+            positive_indices.append(-1)
+    positive_indices = torch.Tensor(positive_indices).to(torch.int32).unsqueeze(-1)
+    positive_indices = indices == positive_indices
```
tomaarsen, line 728: I did a bit of tests with this, and I ran into a bit of an issue (that already existed, too!):
The same text can occur multiple times in `corpus`/`positives`. For example:
```python
pairs_dataset = Dataset.from_dict({
    "query": ["What is the capital of France?", "What is the capital of Germany?", "What is the capital of Italy?", "What is Paris Hilton's first name?"],
    "positive": ["Paris", "Berlin", "Rome", "Paris"],
})
```
If this is your input `dataset`, and you don't specify a separate `corpus`, then your scores before the `topk` might be:
```python
tensor([[0.6274, 0.3247, 0.3938, 0.6274],
        [0.3650, 0.6055, 0.3545, 0.3650],
        [0.3440, 0.3384, 0.5845, 0.3440],
        [0.5039, 0.1982, 0.2269, 0.5039]], dtype=torch.float16)
```
This makes sense: the 1st and 4th column are identical because the 1st and 4th corpus element is identical. After topk, we get:
```python
tensor([[0.6274, 0.6274, 0.3938, 0.3247],
        [0.6055, 0.3650, 0.3650, 0.3545],
        [0.5845, 0.3440, 0.3440, 0.3384],
        [0.5039, 0.5039, 0.2269, 0.1982]], dtype=torch.float16)
```
And then the `positive_indices` become:
```
tensor([[ True, False, False, False],
        [ True, False, False, False],
        [ True, False, False, False],
        [ True, False, False, False]])
```
because this only finds the first occurrence, not all occurrences. We have this same problem with the new
```python
    positive_indices = []
    for positive in positives:
        if positive in corpus:
            index = corpus.index(positive)
            positive_indices.append(index)
        else:
            positive_indices.append(-1)
    positive_indices = torch.Tensor(positive_indices).to(torch.int32).unsqueeze(-1)
    positive_indices = indices == positive_indices
```
as well as the old
```python
    positive_indices = indices == torch.arange(len(queries), device=indices.device).unsqueeze(-1)
```

I'll try to think of something to counteract this.

- Tom Aarsen

tomaarsen:
I made some slight adjustments that I'd like to get your thoughts on. I also have some more thoughts that you might be able to help brainstorm with.

sentence_transformers/util.py
ChrisGeishauser: I totally agree, thanks for the clarification!

sentence_transformers/util.py
ChrisGeishauser: @tomaarsen 
Oh nice, great catch! 

I agree that the easiest is to remove duplicates from the `corpus`. I think assuming/forcing the corpus to contain only non-duplicates is also a reasonable assumption.  As an alternative, we could try to find all occurences of a positive in the corpus and thus keep the corpus intact, but that might be computationally less efficient? But in general, I guess it makes sense that during retrieval you retrieve different documents and no duplicates (so advocating the remove duplicate option for corpus).

Regarding removing duplicates of the positives, in some parts we require that positives and queries have the same size (i.e. calculate similarity_pairwise), so it is good to keep them intact?

ChrisGeishauser:
> I made some slight adjustments that I'd like to get your thoughts on. I also have some more thoughts that you might be able to help brainstorm with.

Sure, anytime! How would like to communicate, just via comments here? @tomaarsen

sentence_transformers/util.py
tomaarsen: >  As an alternative, we could try to find all occurences of a positive in the corpus and thus keep the corpus intact, but that might be computationally less efficient?

Beyond being computationally less efficient, I now realise that keeping duplicates can hurt the quality of the mined negatives. In particular, the script searches for the `range_max + 1` most similar embeddings, and then does filtering based on e.g. `margin`, `max_score`, etc. Lastly, we take the `num_negatives` negatives from the `range_max + 1` that were not filtered away. If we have duplicates, then it'll occur more often that there are not `num_negatives` non-filtered negative candidates. With other words: we'll be able to find (slightly) more useful negatives if we remove duplicates.

> Regarding removing duplicates of the positives, in some parts we require that positives and queries have the same size (i.e. calculate similarity_pairwise), so it is good to keep them intact?

This is indeed a tricky part. One idea that I had was to remove duplicates in `corpus`, but not in `positives`. Of course, this is a bit tricky to do efficiently, but I think I can make this work. Should I get started on this?

tomaarsen:
Yes! I think I'll try and work on removing the duplicates from `corpus` based on our thread here: https://github.com/UKPLab/sentence-transformers/pull/2818#discussion_r1670668533

sentence_transformers/util.py
ChrisGeishauser: > With other words: we'll be able to find (slightly) more useful negatives if we remove duplicates.
Yes, I totally agree!

> This is indeed a tricky part. One idea that I had was to remove duplicates in corpus, but not in positives. Of course, this is a bit tricky to do efficiently, but I think I can make this work. Should I get started on this?
Yes, for me that also feels the way to go now! :) Great to discuss with you Tom!

tomaarsen: I believe I solved it via 46ae230b6938b74ab44391a78170022b72cf53ab. The patch is smaller than expected, but I believe it works well. In essence, I:
1. deduplicate the corpus naively (via list & set)
2. create a mapping from positive indices to corpus indices
3. use the aforementioned mapping to get the full positive embeddings from the deduplicates corpus embeddings
4. use the aforementioned mapping to recognize when a candidate negative from the corpus is equal to a positive, so I can filter those positives away

I think this works pretty nicely: less embeddings required too

tomaarsen: This also allows us to simplify the rescoring filtering for positive pairs: aba1fb77d8cc80a152d2d4a62703c47052970d83

ArthurCamara:
Hey, I was about to open a PR with something similar from here: https://github.com/ArthurCamara/sentence-transformers/tree/hard_negatives_from_corpus
I think the biggest difference is that I considered the case where there could be multiple positives as well (e.g., TREC-Covid has ~500 positives per query). 

This means including a relevance dictionary as a parameter too. 
I can try bringing some of my changes to this PR if you think it is as a meaningful use case as well, @tomaarsen @ChrisGeishauser

tomaarsen:
Hello @ArthurCamara!
I haven't worked much with datasets that have multiple positives per query, but I can imagine that they're fairly common in IR datasets indeed (please correct me if I'm wrong). In your branch, what would a normal `relevance_dataset` look like? A dataset with `["query-id", "corpus-id", "score"]` columns? Also, if I'm looking at it right, then your function requires that both the queries_dataset and corpus_dataset have id columns right?

Beyond your relevance-related changes, I like the `chunk_size` option: I can imagine that you could OOM with `model.encode` otherwise. As for `keep_title`, perhaps that `map` should only trigger if the dataset has a `title` column.

I think your variant is certainly more powerful, but `mine_hard_negatives` from this PR is also a lot simpler to use in terms of data preprocessing. Ideally, I'd like to keep the option for the simple usage, so I think it might be best to support 2 functions. To avoid duplication, it might be possible to rewrite `mine_hard_negatives` to use `mine_hard_negatives_from_corpus` (or `mine_hard_negatives_from_relevance`?) under the hood. One tricky component is that `mine_hard_negatives_from_corpus` must then support all behavior from `mine_hard_negatives`, which would require some extra work (e.g. a non-FAISS implementation, perhaps add corpus deduplication that this PR adds).

Curious about both of your stances on this @ChrisGeishauser @ArthurCamara 

- Tom Aarsen

ArthurCamara:
> Hello @ArthurCamara! I haven't worked much with datasets that have multiple positives per query, but I can imagine that they're fairly common in IR datasets indeed (please correct me if I'm wrong). 
They are somewhat common, but I'm mostly coming from the MTEB/BEIR side of things, where some of the datasets may have multiple positives with different scores (that's why the `positive_score_threshold` argument). 

> In your branch, what would a normal `relevance_dataset` look like? A dataset with `["query-id", "corpus-id", "score"]` columns? Also, if I'm looking at it right, then your function requires that both the queries_dataset and corpus_dataset have id columns right?
> 

You can look at the [trec-covid](https://huggingface.co/datasets/mteb/trec-covid) from MTEB for an example. The `default` subset is a dataset with three columns: `query-id`, `corpus-id` and `scores`. This is also why we need ids for the queries and documents, so we can map the relevance back into the texts. 


> Beyond your relevance-related changes, I like the `chunk_size` option: I can imagine that you could OOM with `model.encode` otherwise. As for `keep_title`, perhaps that `map` should only trigger if the dataset has a `title` column.

Makes sense. Easy to fix. =)

> I think your variant is certainly more powerful, but `mine_hard_negatives` from this PR is also a lot simpler to use in terms of data preprocessing. Ideally, I'd like to keep the option for the simple usage, so I think it might be best to support 2 functions. To avoid duplication, it might be possible to rewrite `mine_hard_negatives` to use `mine_hard_negatives_from_corpus` (or `mine_hard_negatives_from_relevance`?) under the hood. One tricky component is that `mine_hard_negatives_from_corpus` must then support all behavior from `mine_hard_negatives`, which would require some extra work (e.g. a non-FAISS implementation, perhaps add corpus deduplication that this PR adds).
> 
> Curious about both of your stances on this @ChrisGeishauser @ArthurCamara
> 
> * Tom Aarsen

I agree that both settings should be supported. I also don't like the idea that `mine_hard_negatives_from_corpus` needs such specific `Dataset` formats. Probably better to support passing a list of tests for corpus and a mapping from scores. Under the hood this can be transformed into whatever format `mine_hard_negatives_from_corpus` ends up using, including creating some fake document ids.

The non-faiss option is something I was planning on doing today. I was a bit sick the last couple days and finished this way too late at night.

I can do some push a commit to this PR later today with some of the changes and than we can see what it looks like.

ArthurCamara:
Hey, @tomaarsen and @ChrisGeishauser I've merged some of my changes on this PR: https://github.com/UKPLab/sentence-transformers/pull/2848. 
It allows for both a separate corpus being passed and multiple positives per query, using the same interface as before. Let me know if you prefer me to merge my changes into the branch instead.

ChrisGeishauser:
Hi @ArthurCamara  and @tomaarsen! Thanks a lot for keeping me in the loop and sorry for the late reply, I had quite a busy week :) 
@ArthurCamara thanks for suggesting these advancements, the option to have multiple positives would complete this function definitely! I didn't have time yet to have an in-detail look into the code, but can do that probably next Tuesday (if necessary).

Just one question regarding Tom's suggestion to have two functions
> I think your variant is certainly more powerful, but mine_hard_negatives from this PR is also a lot simpler to use in terms of data preprocessing. Ideally, I'd like to keep the option for the simple usage, so I think it might be best to support 2 functions.

Is this still relevant or everything is now put into a single extended method?

ChrisGeishauser:
Hi @tomaarsen and @ArthurCamara, just wanted to ask whether there is any update on your side regard the MR :)

PR #2819, opened 2024-07-09
Move `pytest` config to `pyproject.toml` and remove `pytest.ini`

fpgmaas:
Now that the project uses `pyproject.toml`, we can also consider moving the `pytest` configuration to `pyproject.toml` under [[tool.pytest.ini_options]](https://docs.pytest.org/en/stable/reference/customize.html#pyproject-toml) and remove the `pytest.ini` file.


tomaarsen:
Works great for me! Happy to remove unnecessary another file.

- Tom Aarsen

PR #2823, opened 2024-07-10
[`chore`] Add some additional commands to the Makefile

fpgmaas:
Proposal to add two commands to the `Makefile` to make developer's life a bit easier.

```     
> make help

venv                 Create virtual environment and install development dependencies
check                Run code quality tools
test                 Run unit tests
test-cov             Run unit tests and generate a coverage report
build                Build the wheel file
help                 Show help for the commands
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,18 @@
+.PHONY: venv
+venv: ## Create virtual environment and install development dependencies
+	@echo "Creating virtual environment..."
+	@python -m venv .venv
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

I think venv's are a bit personal, so I don't want to recommend an approach, and I think `make build` is not much simpler than `python -m build`, so I think we can leave this be.

- Tom Aarsen

PR #2826, opened 2024-07-10
Fix `ruff` pre-commit hook.

fpgmaas:
Currently the pre-commit hooks do not automatically fix issues that ruff can autofix, due to a recent change by me. This PR fixes that problem.


tomaarsen:
I appreciate it!

- Tom Aarsen

PR #2830, opened 2024-07-10
[`chore`] Enable ruff rules `UP006` and `UP007` to improve type hints.

fpgmaas:
This PR proposes to enable the following rules in `ruff`:

- [UP006](https://docs.astral.sh/ruff/rules/non-pep585-annotation/)
- [UP007](https://docs.astral.sh/ruff/rules/non-pep604-annotation/)

This implements changes as proposed by [PEP585](https://peps.python.org/pep-0585/) and [PEP604](https://peps.python.org/pep-0604/). For example, this changes

```py
from typing import List

def foo(
        documents: List[str],
        top_k: Optional[int] = None,
        ...
)
```

to

```py
def foo(
        documents: list[str],
        top_k: int | None = None,
        ...
)
```

Steps to create this PR were simply updating the `pyproject.toml` and then running `ruff check --unsafe-fixes`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,7 +62,7 @@ line-length = 119
 fix = true
 
 [tool.ruff.lint]
-select = ["E", "F", "W", "I"]
+select = ["E", "F", "W", "I", "UP006", "UP007"]
 # Skip `E731` (do not assign a lambda expression, use a def)
 ignore = [
     # LineTooLong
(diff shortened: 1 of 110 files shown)


pyproject.toml
```diff
@@ -72,11 +72,21 @@ ignore = [
...
+    "I002"
+    ]
+"docs/**" = [
+    # Ignore missing required imports
+    "I002"
+    ]
 
 [tool.ruff.lint.isort]
 known-third-party = ["datasets"]
+required-imports = ["from __future__ import annotations"]
```
tomaarsen, line 88: Perfect, was looking for this. Otherwise I can't move forward with the typing changes without breaking backwards compatibility.

tomaarsen:
Thank you! I think this'll be quite useful. Would you be able to resolve the conflicts in #2834?

- Tom Aarsen

PR #2834, opened 2024-07-11
[`chore`] Enable ruff's pypgrade (`UP`) ruleset

fpgmaas:
This PR builds on top of [the previous PR](https://github.com/UKPLab/sentence-transformers/pull/2830), which served as a stepping stone; now we can easily enable the full pyupgrade (`UP`) ruleset.

The most common changes are:

## [UP008](https://docs.astral.sh/ruff/rules/super-call-with-parameters/)

Checks for super calls that pass redundant arguments.

e.g. replace

```py
super(Dense, self).__init__()
```

with

```py
super().__init__()
```

## [UP032](https://docs.astral.sh/ruff/rules/f-string/)

Checks for str.format calls that can be replaced with f-strings.

e.g. replace

```py
print("\nCluster {}, #{} Elements ".format(i + 1, len(cluster)))
```

with

```py
print(f"\nCluster {i + 1}, #{len(cluster)} Elements ")
```

## [UP037](https://docs.astral.sh/ruff/rules/quoted-annotation/)

Checks for the presence of unnecessary quotes in type annotations.

e.g. replace

```py
self, model: "SentenceTransformer", output_path: str = None, epoch: int = -1, steps: int = -1

```

with

```py
self, model: SentenceTransformer, output_path: str = None, epoch: int = -1, steps: int = -1
```

### DIFF
--- a/examples/applications/clustering/fast_clustering.py
+++ b/examples/applications/clustering/fast_clustering.py
@@ -57,11 +57,11 @@ start_time = time.time()
 # threshold: Consider sentence pairs with a cosine-similarity larger than threshold as similar
 clusters = util.community_detection(corpus_embeddings, min_community_size=25, threshold=0.75)
 
-print("Clustering done after {:.2f} sec".format(time.time() - start_time))
+print(f"Clustering done after {time.time() - start_time:.2f} sec")
 
 # Print for all clusters the top 3 and bottom 3 elements
 for i, cluster in enumerate(clusters):
-    print("\nCluster {}, #{} Elements ".format(i + 1, len(cluster)))
+    print(f"\nCluster {i + 1}, #{len(cluster)} Elements ")
     for sentence_id in cluster[0:3]:
         print("\t", corpus_sentences[sentence_id])
     print("\t", "...")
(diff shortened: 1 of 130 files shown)


fpgmaas:
@tomaarsen rebased to resolve conflicts and updated the description to include the most common changes in the codebase.

tomaarsen (approved):
I had no idea there were still this many uses of the old string formatting in the ST codebase.

tomaarsen:
Thank you! I'm glad to see the codebase get modernized.

- Tom Aarsen

tomaarsen:
@fpgmaas Are these upgrades automatically applied via `make check` now? Or do they still need to be fixed via `ruff check . --unsafe-fixes`?

fpgmaas:
@tomaarsen All are automatically applied, except the ones shown in [this commit](https://github.com/UKPLab/sentence-transformers/pull/2834/commits/cbf1ccf6569177d897cf266eed8f90cee4f4aa94), which only contains fixes for UP008. These were applied as `--unsafe-fixes`, so likely that also needs to happen in the future. I was slightly confused by this, since the [docs](https://docs.astral.sh/ruff/rules/super-call-with-parameters/) for UP008 state "Fix is always available."

PR #2846, opened 2024-07-18
# Pass Image color channels information to Transformers

davychxn:
Background:
In Huggingface Transformers' image processor, e.g. CLIPImageProcessor, the constructor requires input of input_data_format, which gives the Image's color channels being in the first or the last position in its shape.

For example, if an image's shape is (512, 512, 3), it means its resolution is 512*512 pixels, and it has RBG, 3 color channels. In this case, input_data_format is ImageChannelDimension.LAST or ChannelDimension.LAST in Transformers.

Sometimes, people would use customized Image format in a shape of (3, 512, 512) for performance purpose. Transformers requires users to point it out, or it would infer to tell it from its shape.

Generally, an image would have 1 or 3 color channels representing Gray or RGB. So, the inferring algorithm in Transformers looks for 1 or 3 values in the image's shape.

If your input images are in the shape of (3, xxx, 1) or (1, xxx, 3), the inferring algorithm would get confused, and raise following exception: 'The channel dimension is ambiguous. Got image shape (1, xxx, 3). Assuming channels are the first dimension.' 'ValueError: mean must have 1 elements if it is an iterable, got 3'

Fix:
1. Add a class ImageChannelDimension to define 2 possible Image color channels position in an Image's shape
2. Input this information in model.encode method, and pass it to Tokenizer and image processor from Transformers.


davychxn:
How to reproduce the issue:
Use this image:
header-separator

Save above image as "header-separator.png".

Run the official example:
```
from PIL import Image

#Load CLIP model
model = SentenceTransformer('clip-ViT-B-32')

#Encode an image:
img_emb = model.encode(Image.open('header-separator.png'))
```

davychxn:
With the fix we can tokenize image in the shape of (xxx, xxx, 3) like:
```
...
// By default
model.encode(Image.open('image.png'))
// or
model.encode(Image.open('image.png', input_data_format=util.ImageChannelDimension.LAST))
```
And tokenize image in the shape of (3, xxx, xxx) like:
```
model.encode(Image.open('image.png', input_data_format=util.ImageChannelDimension.FIRST))
```

davychxn:
@tomaarsen Hi Tom, would you help to check my PR? Thank you.

davychxn:
@fpgmaas Hi Florian, would you take some time to review my PR? Thank you.

sentence_transformers/SentenceTransformer.py
```diff
@@ -430,6 +435,7 @@ def encode(
         convert_to_tensor: bool = False,
         device: str = None,
         normalize_embeddings: bool = False,
+        input_data_format: str = ImageChannelDimension.LAST,
```
fpgmaas, line 438: this should also be mentioned and explained in the docstring

sentence_transformers/util.py
```diff
@@ -28,6 +28,12 @@
     from sentence_transformers.cross_encoder.CrossEncoder import CrossEncoder
     from sentence_transformers.SentenceTransformer import SentenceTransformer
 
+class ImageChannelDimension():
```
fpgmaas, line 31: This should probably be an `Enum`

fpgmaas:
Maybe `input_data_format` is not an appropriate name? In my opinion the name should be more specific. Maybe just `image_channel_dimension`?

fpgmaas:
Hey @davychxn , I am not a maintainer of the project and I lack in-depth knowledge to judge your proposed changes, so I'm afraid I cannot approve the PR. However, I left a few review comments anyways to help the PR along.

sentence_transformers/SentenceTransformer.py
davychxn: Fixed.

sentence_transformers/util.py
davychxn: This class is copied from Transformers' repo. It is defined like this there. Because the string defined in the class is needed by Transformers' image-processor. If we use Enum, I think, we'll get integer values? And we need to convert to string before passing to Transformers?

davychxn:
Hi @tomaarsen , would you review my change? Any improvement we'll need here? Thank you.

davychxn:
This PR is needed while writing code to this project: https://github.com/Immortalise/SearchAnything
The issue can crash the code there, I had to use a hard-coded workaround.

I submitted the initial fix PR to Transformers, but @amyeroberts suggested fix in your repo. The closed PR is: https://github.com/huggingface/transformers/pull/31950

yaner-here:
Approved. What's worse, it can cause counterintuitive crash in some circumstances.

```python
from transformers import CLIPProcessor, CLIPModel # type: ignore
from torch import Tensor
import json
import PIL
import PIL.Image
from math import floor, ceil

if __name__ == "__main__":

    image = PIL.Image.open("../grounding_data/coco/train2017/000000229284.jpg").convert("RGB") # any large image as you wish
    bboxes = [(154, 410, 163, 413), (25, 439, 37, 445)]
    regions = [image.crop(bbox) for bbox in bboxes]

    # >>> Just pick one. <<<
    # regions = [regions[0]] # Fine.
    # regions = [regions[1]] # Fine.
    # regions = [regions[2]] # Fine.
    # regions = [regions[1], regions[2]] # Fine.
    regions = [regions[0], regions[1]] # Error!
    # regions = [regions[0], regions[2]] # Error!
    # regions = [regions[0], regions[1], regions[2]] # Error!

    device = "cuda:0"
    model = CLIPModel.from_pretrained("openai/clip-vit-base-patch32").to(device)
    processor = CLIPProcessor.from_pretrained("openai/clip-vit-base-patch32")
    inputs = processor(images=regions, return_tensors="pt", padding=True)
    inputs = inputs["pixel_values"].to(device)
    outputs = model.get_image_features(inputs)
```

`regions[0]` is in shape of `[9,3]` with RGB channels, resulting in confusing channel dimension.

```
TypeError: Cannot handle this data type: (1, 1, 6), |u1
```


It's a pity that the [#31950](https://github.com/huggingface/transformers/pull/31950#issue-2407328057) PR have not been merged.

tomaarsen:
Hello!

In v5.4 image preprocessing goes through the model's processor; channel-format overrides now flow through `processing_kwargs={"image": {"input_data_format": ...}}`, so you should now be able to do as you wish! Closing as covered.

- Tom Aarsen

PR #2847, opened 2024-07-18
Change eval dataloader to use eval_batch_size

akashd-2:
The evaluation dataloader is currently initialized using the train_batch_size which seems to be a bug. This PR fixes this.

### DIFF
--- a/sentence_transformers/trainer.py
+++ b/sentence_transformers/trainer.py
@@ -628,7 +628,7 @@ class SentenceTransformerTrainer(Trainer):
         else:
             batch_sampler = self.get_batch_sampler(
                 eval_dataset,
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Well spotted! I had another look at the other batch sizes as well, and I noticed that I'm using `train` for the `get_test_dataloader` currently: I think `eval` is better (there is no `per_device_test_batch_size`). This also matches what `transformers` does.

Additionally, I started using `..._batch_size` everywhere instead of `per_device_..._batch_size`: The former is identical for single-GPU and DDP, but has a higher batch size for DP (which is the expected behaviour). I think after this PR, everything should be as expected again!

- Tom Aarsen

PR #2859, opened 2024-07-24
Perform additional check for owner string in `is_<library>_available` functions

### LINKED ISSUE #2858: Initializing the package fails if there's unrelated `datasets` module in PYTHONPATH
This is because sentence-transformers optionally imports from a package also called `datasets`. Now in my data-related contexts, repos using the `sentence-transformers` package might *also* have a module called `datasets`. I do not want to have to change the name of my internal module.

At the moment the use of Huggingface's `datasets` functionality is optional in this repo: there's a function that checks whether the package is available:

https://github.com/UKPLab/sentence-transformers/blob/c0fc0e8238f7f48a1e92dc90f6f96c86f69f1e02/sentence_transformers/util.py#L1314-L1318

There are various other similar issues related to it, namely #2817

I believe a slightly smarter availability function would fix it. Something like

```py
def is_datasets_available() -> bool:
    """
    Returns True if the Hugging Face datasets library is available.
    """
    spec = importlib.util.find_spec("datasets")
    if spec and 'huggingface' in spec.origin:
        return True
    return False
```

leblancfg:
Currently if you have a `datasets` module in your `sys.path`, the `is_datasets_available` function will return `True`, even if it's not the one from Huggingface.

Here we add a slightly smarter logic to check the origin string of the `importlib` spec, and match it against the expected owner. This makes it possible for users who also have an unrelated module in their `sys.path` called `datasets` to do imports like

```py
```

while leaving all other functionality untouched.

Closes #2858

### DIFF
--- a/sentence_transformers/util.py
+++ b/sentence_transformers/util.py
@@ -1304,7 +1304,7 @@ def get_device_name() -> Literal["mps", "cuda", "npu", "hpu", "cpu"]:
     return "cpu"
 
 
-def check_package_availability(package_name: str, owner: str) -> None:
+def check_package_availability(package_name: str, owner: str) -> bool:
     """
     Checks if a package is available from the correct owner.
     """
--- a/sentence_transformers/util.py
+++ b/sentence_transformers/util.py
@@ -9,6 +9,7 @@ import queue
 import random
 import sys
 from contextlib import contextmanager
+from importlib.metadata import PackageNotFoundError, metadata
 from typing import TYPE_CHECKING, Any, Callable, Literal, overload
 
 import numpy as np
@@ -1308,10 +1309,11 @@ def check_package_availability(package_name: str, owner: str) -> bool:
     """
     Checks if a package is available from the correct owner.
     """
-    spec = importlib.util.find_spec(package_name)
-    if spec and owner in spec.origin:
-        return True
-    return False
+    try:
+        meta = metadata(package_name)
+        return meta["Name"] == package_name and owner in meta["Home-page"]
+    except PackageNotFoundError:
+        return False
 
 
 def is_accelerate_available() -> bool:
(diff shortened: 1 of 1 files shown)


sentence_transformers/util.py
```diff
@@ -1304,22 +1304,25 @@ def get_device_name() -> Literal["mps", "cuda", "npu", "hpu", "cpu"]:
...
+def check_package_availability(package_name: str, owner: str) -> bool:
     """
-    Returns True if the accelerate library is available.
+    Checks if a package is available from the correct owner.
     """
-    return importlib.util.find_spec("accelerate") is not None
+    spec = importlib.util.find_spec(package_name)
+    if spec and owner in spec.origin:
+        return True
+    return False
```
tomaarsen, line 1314: Hello!

I'd like to improve the `is_..._available` functions to avoid the issue with a `datasets` folder, but I don't believe this works correctly:

```python
# !pip install datasets
>>> import importlib
>>> spec = importlib.util.find_spec("datasets")
>>> spec
ModuleSpec(name='datasets', loader=<_frozen_importlib_external.SourceFileLoader object at 0x7ae5f3430e20>, origin='/usr/local/lib/python3.10/dist-packages/datasets/__init__.py', submodule_search_locations=['/usr/local/lib/python3.10/dist-packages/datasets'])
>>> spec.origin
'/usr/local/lib/python3.10/dist-packages/datasets/__init__.py'
>>> "huggingface" in spec.origin
False
```

Perhaps we can use this:
```suggestion
    try:
        metadata = importlib.metadata.metadata(package_name)
        return metadata["Name"] == package_name and owner in metadata["Home-page"]
    except PackageNotFoundError:
        return False
```

This does work better for me:
```python
>>> from importlib.metadata import metadata
>>> meta = metadata("datasets")
>>> meta["Name"]
'datasets'
>>> meta["Home-page"]
'https://github.com/huggingface/datasets'
>>> meta = metadata("tests") # Note: the tests folder exists locally for me
```

What do you think?

- Tom Aarsen

tomaarsen:
I ran some tests locally, and it seems to work just like you'd hope now! Thanks a bunch, I'm merging this now. (We can ignore the CI test failures, they're unrelated due to disk space).

- Tom Aarsen

PR #2861, opened 2024-07-25
[`docs`] Fix typo and broken links in documentation

ZiyiXia:


tomaarsen:
Hello!

The documentation is written in .rst (ReStructuredText), so we can't just use Markdown formatting. Instead, I've updated the links to use .rst, and they should work once I rebuild the documentation for the next release. Thanks for catching these issues!

- Tom Aarsen

PR #2862, opened 2024-07-25
Add triplet margin for distance functions in TripletEvaluator

zivicmilos:
### Problem:

Triplet margin exists in TripletLoss but does not exist in TripletEvaluator.

---

### Motivation:

When evaluating the model using TripletEvaluator during, for example, the training it would be a nice thing to have a piece of information not only if the negative is further away from the anchor than the positive, but also if it is further away from the anchor by some **predefined margin**. Because the triplet evaluator uses 4 different distance metrics, we would need an option for each one, so 4 different margins for each of them.

---

### Solution:

In the commit, I implemented exactly that, option for setting the margin for each distance function (or any subset of them). It is an optional argument to the class constructor, so it can be called perfectly the same as previously, with the difference that now you have more flexibility in setting how far away you want your negative and positive to be for evaluation.

---

PS. I can also write some evaluator tests, both for TripletEvaluator with and without the margin.

PPS. I ran a make check, so the code follows coding guidelines.

---


tomaarsen:
Hello!

Thanks for this PR. I made some changes as I noticed that cosine, manhattan, and Euclidean were based on "distance", but dot was based on similarity instead. I've now updated all to use similarity, so that higher is always better regardless of which similarity function is used.

I've also renamed the parameter to `margin` - the `triplet_...` part is already implied because it's on the TripletEvaluator. And I made it singular because I think most people will use it as a single float. 

I also added a test. Please let me know what you think @zivicmilos!

- Tom Aarsen

tomaarsen:
Hello!

Apologies for the delay - I still wanted to add d8f4c3d0298bb34f606d818484758801dc18767c to ensure that there's no backwards incompatibility when people could previously use `TripletEvaluator(..., main_distance_function=...)`. 

What do you think?

- Tom Aarsen

zivicmilos:
Hi @tomaarsen,

This looks awesome, great catch, thanks for the help!

tomaarsen:
Thanks a bunch for raising this & for dealing with my slowness, I appreciate it.

- Tom Aarsen

Issue #2882, opened 2024-08-07
Precision does not work when requesting token_embeddings


kacperlukawski:
Hi! I've been experimenting with token embeddings recently and found out I cannot reduce the precision if I request them.

```python
emb = model.encode(
    [
        "Hello Word, a test sentence",
        "Here comes another sentence",
        "My final sentence",
    ],
    output_value="token_embeddings",
    precision="uint8"
)
```

The `quantize_embeddings` function raises a ValueError since the shapes of token embedding arrays are unaligned and cannot be combined into a single numpy array. Here is part of the stack trace:

```
  File "/home/kacper/.cache/pypoetry/virtualenvs/beir-qdrant-wY0sLiQM-py3.10/lib/python3.10/site-packages/sentence_transformers/SentenceTransformer.py", line 553, in encode
    all_embeddings = quantize_embeddings(all_embeddings, precision=precision)
  File "/home/kacper/.cache/pypoetry/virtualenvs/beir-qdrant-wY0sLiQM-py3.10/lib/python3.10/site-packages/sentence_transformers/quantization.py", line 400, in quantize_embeddings
    embeddings = np.array(embeddings)
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (4,) + inhomogeneous part.
```

@tomaarsen I'm happy to provide a PR fixing this.

PR #2885, opened 2024-08-08
[fix] Quantization of token embeddings

kacperlukawski:
### Problem

The `encode` method raises a `ValueError` when we request `precision` different than `float32` and `output_value="token_embeddings"`, as reported in #2882.

### Solution

This PR provides a fix that combines all the token embeddings into a single array, runs the normalization, and eventually reconstructs the shape of the original array so we can distinguish token embeddings coming from each input example.

### DIFF
--- a/sentence_transformers/quantization.py
+++ b/sentence_transformers/quantization.py
@@ -394,9 +394,15 @@ def quantize_embeddings(
     Returns:
         Quantized embeddings with the specified precision
     """
+    outputs, lengths = None, None
     if isinstance(embeddings, Tensor):
         embeddings = embeddings.cpu().numpy()
+        embeddings = np.concatenate(embeddings)
     elif isinstance(embeddings, list):
+        if not isinstance(embeddings[0], list) and len(embeddings[0].shape) == 2:
+            # It will happen when we request token_embeddings
+            lengths = [embedding.shape[0] for embedding in embeddings]
+            embeddings = np.concatenate(embeddings)
         if isinstance(embeddings[0], Tensor):
             embeddings = [embedding.cpu().numpy() for embedding in embeddings]
         embeddings = np.array(embeddings)
@@ -404,7 +410,7 @@ def quantize_embeddings(
         raise Exception("Embeddings to quantize must be float rather than int8 or uint8.")
 
     if precision == "float32":
-        return embeddings.astype(np.float32)
+        outputs = embeddings.astype(np.float32)
 
     if precision.endswith("int8"):
         # Either use the 1. provided ranges, 2. the calibration dataset or 3. the provided embeddings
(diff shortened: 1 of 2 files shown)


sentence_transformers/quantization.py
```diff
@@ -394,17 +394,23 @@ def quantize_embeddings(
...
+    outputs, lengths = None, None
     if isinstance(embeddings, Tensor):
         embeddings = embeddings.cpu().numpy()
+        embeddings = np.concatenate(embeddings)
     elif isinstance(embeddings, list):
+        if not isinstance(embeddings[0], list) and len(embeddings[0].shape) == 2:
+            # It will happen when we request token_embeddings
+            lengths = [embedding.shape[0] for embedding in embeddings]
+            embeddings = np.concatenate(embeddings)
         if isinstance(embeddings[0], Tensor):
```
ir2718, line 406: Shouldn't this if statement be above the previous if, as sending in a list of Tensors is also valid?

kacperlukawski: @ir2718 You were absolutely right, thank you! Changed the order of the statements.

tomaarsen:
Hello!

Apologies, I haven't yet had time to look into this a bit deeper, but I think an edge case that might be missed is `output_value=None`. This is not very well documented, but it returns both the sentence embedding and the token embeddings. I can imagine that this might be valuable for some use cases.

- Tom Aarsen

ir2718:
Not sure if I can modify the PR, but following Tom's dict edge case, I think adding this should suffice:
```
        if isinstance(embeddings[0], dict):
            sentence_embeddings = [x["sentence_embedding"].unsqueeze(0).cpu().numpy() for x in embeddings]

            token_embeddings = []
            for emb_dict in embeddings:
                token_emb = emb_dict["token_embeddings"]
                attention = emb_dict["attention_mask"]
                last_mask_id = len(attention) - 1
                while last_mask_id > 0 and attention[last_mask_id].item() == 0:
                    last_mask_id -= 1

                token_embeddings.append(token_emb[0 : last_mask_id + 1])

            token_embeddings = [x.cpu().numpy() for x in token_embeddings]
            embeddings = token_embeddings + sentence_embeddings
            lengths = [x.shape[0] for x in embeddings]
```

with a modification in SentenceTransformer.py, line 638, right before the return statement:
```
        if output_value is None:
            return {
                "token_embeddings": all_embeddings[:len(all_embeddings)//2],
                "sentence_embedding": all_embeddings[len(all_embeddings)//2:]
            }
```

kacperlukawski:
Thanks, @ir2718! I wonder whether we should return a dictionary. That breaks the interface of the encode method. @tomaarsen Would that be the expected behaviour?

kacperlukawski:
I decided to implement the quantization for this edge case differently than suggested. The `quantize_embeddings` method remains unchanged, but I extended the `encode` method. The `all_embeddings` were already a dictionary there, so I combined token and sentence embeddings and passed them all together to quantize. The output dictionary structure remains unchanged except for a different precision.

@ir2718 I didn't use the attention mask on purpose. I thought it would be best to keep the shapes consistent, no matter if we use `float32` or any other precision.

ir2718:
>  I wonder whether we should return a dictionary. That breaks the interface of the encode method.

Agreed, I was thinking about that myself, but since transformers mostly handle things in dicts my first idea was to implement it that way. Not breaking the interface is probably a better solution, but requires adding some kind of note in the docs about the ordering of embeddings.

Issue #2891, opened 2024-08-15
Inconsistent token_embeddings shape with "output_value" parameter in model.encode()


yves03:
I am using the latest version of sentence-transformers, version 3.0.1, installed with pip in a conda environment on a windows machine. 
I use python 3.12.4.

I found that when calling the encode() function on a SentenceTransformer model, the shape (and thus content) of the "token_embeddings" depends on the value of the "output_value" argument. I would expect this to be consistent instead.

In my specific case, I am using this model running on a nvidia GPU : https://huggingface.co/manu/bge-m3-custom-fr 
Depending on how I extract the token embeddings, I get different shapes:

```
model.encode(test_chunk_list, output_value='token_embeddings')[0].shape
> torch.Size([99, 1024])
```

```
model.encode(test_chunk_list, output_value=None)[0]['token_embeddings'].shape
> torch.Size([147, 1024])
```

I think the issue is related to [this line of code](https://github.com/UKPLab/sentence-transformers/blob/c0fc0e8238f7f48a1e92dc90f6f96c86f69f1e02/sentence_transformers/SentenceTransformer.py#L596) that is done when output_value="token_embeddings" but not [when it is set to None ](https://github.com/UKPLab/sentence-transformers/blob/c0fc0e8238f7f48a1e92dc90f6f96c86f69f1e02/sentence_transformers/SentenceTransformer.py#L601).

tomaarsen:
Hello!

Apologies for the delay, I've been recovering from a surgery this last month.
I'm a bit hesitant to make this change, as with `output_value=None` I'm outputting all of the "raw" data, including inputs etc. With other words, I think it might make some sense that the token embeddings then still have padding token embeddings that would need to be removed by the user. 
Beyond that, this change (or fix) would be breaking, which I'd like to avoid.

- Tom Aarsen

yves03:
Hi Tom,

No problem, thank you for your work !

It is true that with the help attention mask, we can obtain the same shape and content of tokens_embeddings with the output_value set to None as when it is set to "token_embeddings" only.

I understand that you are cautious about the proposed change, I didn't test the rest of the code to make sure it wouldn't break anything. I'll remove the PR then.

I wish you a speedy recovery !
Yves

PR #2892, opened 2024-08-15
Add attention masking to the return of the encode() function "token_embeddings" when "output_value" is set to None

yves03:
Fix the issue described [here](https://github.com/UKPLab/sentence-transformers/issues/2891).


tomaarsen:
I'm not 100% sure that we should make this change, see https://github.com/UKPLab/sentence-transformers/issues/2891#issuecomment-2340036538

Issue #2841, opened 2024-07-15
MatryoshkaLoss as 'module' object is not callable


Anakeen1274:
Hello,
I am simply running the example program:
```
!pip install sentence_transformers
from sentence_transformers import SentenceTransformer, losses

model = SentenceTransformer("microsoft/mpnet-base")

train_loss = losses.MultipleNegativesRankingLoss(model=model)
train_loss = losses.Matryoshka2dLoss(model, train_loss, [768, 512, 256, 128, 64])
```
and receiving the following error:
```
in Matryoshka2dLoss.__init__ [...]
matryoshka_loss = MatryoshkaLoss( model, [...])
TypeError: 'module' object is not callable
```
The same error appeared both on a google collaboratory notebook and on a vscode kernel.

I shall remain available if any further information is needed on the environment i ran it on.
Best regards

Anakeen1274:
Here is a simplistic example of the error:
https://colab.research.google.com/drive/1mFC9tswL5CE-_UuYi6mG-cheCRUSgj8q?usp=sharing
Hope it helps!

Woracle:
to add weight to this, I have encountered this issue using version sentence-transformers==3.0.1

Anakeen1274:
Exactly, the same goes for AdaptiveLayerLoss(), it works on it's own, only Matryoshka2dLoss has this bugged behaviour.

Anakeen1274:
I've gotta stop closing issues on accident.

pesuchin:
The bug will likely be fixed by the following pull request.
https://github.com/UKPLab/sentence-transformers/pull/2907

tomaarsen:
Indeed! Feel free to use 
```
pip install git+https://github.com/UKPLab/sentence-transformers.git
```
if you want to use the fix before I release it in a new version on PyPI.

- Tom Aarsen

PR #2907, opened 2024-08-23
Fix 'module object is not callable' error in Matryoshka2dLoss

pesuchin:
## Issue

https://github.com/UKPLab/sentence-transformers/issues/2841

## Problem
A 'module object is not callable' error was occurring when using the Matryoshka2dLoss class.

## Solution
I changed the absolute imports to relative imports in the Matryoshka2dLoss.py file, as shown in the following code:
https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/losses/BatchSemiHardTripletLoss.py#L10

Specifically:
- Before: 
```python
from sentence_transformers.losses import AdaptiveLayerLoss, MatryoshkaLoss
```
- After:
```python
from .AdaptiveLayerLoss import AdaptiveLayerLoss
from .MatryoshkaLoss import MatryoshkaLoss
```

## Verification

Tested the changes in a local environment and confirmed that this modification resolves the aforementioned error. The following code was used for verification:

```python

train_loss = losses.MultipleNegativesRankingLoss(model=model)
train_loss = losses.Matryoshka2dLoss(model, train_loss, [768, 512, 256, 128, 64])
```

## Review Notes
Please pay attention to:
- The appropriateness of the import changes
- Any potential side effects on other parts of the codebase

Your review and feedback are greatly appreciated.

### DIFF
--- a/sentence_transformers/losses/Matryoshka2dLoss.py
+++ b/sentence_transformers/losses/Matryoshka2dLoss.py
@@ -4,9 +4,10 @@ from typing import Any
 
 from torch.nn import Module
 
+from sentence_transformers import SentenceTransformer
+
 from .AdaptiveLayerLoss import AdaptiveLayerLoss
 from .MatryoshkaLoss import MatryoshkaLoss
-from sentence_transformers import SentenceTransformer
 
 
 class Matryoshka2dLoss(AdaptiveLayerLoss):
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

I can reproduce the issue, and this PR indeed seems to resolve it!

- Tom Aarsen

Issue #2933, opened 2024-09-12
Version 3.1.0 forces a downgrade to Numpy 1.x


kozlek:
Hi,

My team and I were using `sentence-transformers` version `3.0.1` with `numpy 2.x` without any trouble.
Today, when we tried to upgrade to the recently released version `3.1.0`, `numpy` was forced to downgrade to `1.x`.

It seems to be related to the migration to `pyproject.toml` where an upper bound constraint has been added to `numpy`:
- https://github.com/UKPLab/sentence-transformers/blob/v3.0-release/setup.py#L25
- https://github.com/UKPLab/sentence-transformers/blob/v3.1-release/pyproject.toml#L36

Is this an expected behaviour?
If yes, I suggest to mention it in the release note as a `numpy` downgrade is not a minor change.

On our side, we'll stick to version `3.0.1` as we already migrated our code to `numpy 2.x`.

tomaarsen:
Hello!

This was indeed intended, though I recognize that it's suboptimal. The choice was originally made when `numpy==2.0.0` caused failures for all Windows builds: https://github.com/UKPLab/sentence-transformers/actions/runs/9547213601/job/26311747682?pr=2757

> A module that was compiled using NumPy 1.x cannot be run in NumPy 2.0.0 as it may crash. To support both 1.x and 2.x versions of NumPy, modules must be compiled with NumPy 2.0. Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.
>
> If you are a user of the module, the easiest solution will be to downgrade to 'numpy<2' or try to upgrade the affected module. We expect that some modules will need time to support NumPy 2.

I'd love to get rid of the `numpy<2` upper bound, but only if Windows users can use the project without hastle in a fresh install, and I haven't yet investigated what that would take.

- Tom Aarsen

kozlek:
Hi Tom, 

Thank you for your quick answer.

I get that it could be problematic for Windows users, however version `3.1.0` is now unusable for all the projects which depends on `numpy 2.x` (and which spent multiple days to migrate their existing codebase 🥲), and for all platforms.

We'll wait for the next version `numpy 2.x`-compatible, but I suggest to add a mention in the release note as it is an important breaking change.

Good luck with the Windows build issue, I know this is not always easy...

tomaarsen:
[...]

PR #2934, opened 2024-09-12
[chore] Allow numpy 2 for all platforms except win32

kozlek:
See #2933.


tomaarsen:
The correct `numpy` versions is being installed on each platform, and all tests are passing. I'll see if I can get `numpy` 2 support for Windows too, but otherwise I'll merge & release this soon.

Thanks!

- Tom Aarsen

tomaarsen:
Looks like I can fully remove the numpy 2 restriction! See #2937
In short, it seems that scipy has been updated.

- Tom Aarsen

PR #2947, opened 2024-09-20
[ENH] -- `CrossEncoder.rank`

it176131:
I found some spots in the :method:`CrossEncoder.rank` that could be updated to reflect best practices. Feel free to cherry-pick or reject.

Tested with Python 3.8 for minimum version coverage.

### DIFF
--- a/sentence_transformers/cross_encoder/CrossEncoder.py
+++ b/sentence_transformers/cross_encoder/CrossEncoder.py
@@ -546,7 +546,7 @@ class CrossEncoder(PushToHubMixin):
         for i, score in enumerate(scores):
             results.append({"corpus_id": i, "score": score})
             if return_documents:
-                results[-1] |= {"text": documents[i]}
+                results[-1] = {**results[-1], "text": documents[i]}
 
         results = sorted(results, key=lambda x: x["score"], reverse=True)
         return results[:top_k]
--- a/sentence_transformers/cross_encoder/CrossEncoder.py
+++ b/sentence_transformers/cross_encoder/CrossEncoder.py
@@ -532,7 +532,7 @@ class CrossEncoder(PushToHubMixin):
         """
         query_doc_pairs = [[query, doc] for doc in documents]
         scores = self.predict(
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

I'm a bit hesitant to (already) add code with the `|=` operator, because Sentence Transformers is currently still supported on Python 3.8, whereas that operator is not. 
However, in about 1 month, [Python 3.8 will become deprecated](https://endoflife.date/python), and then I'll have no issues merging this PR.

- Tom Aarsen

it176131:
> Hello!
> 
> I'm a bit hesitant to (already) add code with the `|=` operator, because Sentence Transformers is currently still supported on Python 3.8, whereas that operator is not. However, in about 1 month, [Python 3.8 will become deprecated](https://endoflife.date/python), and then I'll have no issues merging this PR.
> 
> * Tom Aarsen

Good catch! Interesting that the 3.8 tests didn't raise an exception 🤔

tomaarsen:
Huh, you're right. Perhaps `rank` isn't under test, or I'm wrong and `|=` also works on Python 3.8.

it176131:
> Huh, you're right. Perhaps `rank` isn't under test, or I'm wrong and `|=` also works on Python 3.8.

I'm not seeing a check in the [current test](https://github.com/UKPLab/sentence-transformers/blob/b6f114288f8cf1edc6eb7739d95b7adf15d49fcb/tests/test_cross_encoder.py#L136), but I'd be willing to add one for better coverage.

tomaarsen:
That would be much appreciated!

it176131:
> That would be much appreciated!

Updated the test!

it176131:
@tomaarsen is there anything else we need before merging?

tomaarsen:
I don't believe so! I think we're all set now, I hadn't merged it yet as I was wrapping up some work in another PR.
I plan to include this in the next release, which should be very soon (i.e. this week, I imagine).

- Tom Aarsen

PR #2951, opened 2024-09-23
[`feat`] Add query prompts to Information Retrieval Evaluator

ArthurCamara:
Currently, there is no way to add prompts to the Evaluators. This PR adds this function to the Information Retrieval Evaluator, as discussed in https://github.com/UKPLab/sentence-transformers/pull/2848#issuecomment-2361069029

----
Older description that need to go another PR:

Most models, specially based on Mistral, are trained with left-padding in the tokenizer, instead of the common right-padding when evaluating. (e.g., [LLM2Vec](https://github.com/McGill-NLP/llm2vec/blob/68dc1d3244cc710942a5bbbf11d9677de9f8f68a/llm2vec/llm2vec.py#L97)), but this is not covered in the current [`prompt_length`](https://github.com/UKPLab/sentence-transformers/blob/7290448809cb73f08f63c955550815775434beb4/sentence_transformers/SentenceTransformer.py#L555) strategy. 
Currently, If a batch with mismatched lengths is left-padded, the masking will mostly mask out padding tokens. This PR fixes that by adding a `mask_prompt` argument to the tokenize function. When this flag is set, the code tries to find the first non-padding token in each sentence, and will mask everything between that and `prompt_length`, adding a `prompt_mask` representation to the output dictionary.

Perhaps a more "elegant" solution would be to replace `prompt_length` entirely, but this could break Instructor models.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,6 @@ dependencies = [
     "transformers>=4.38.0,<5.0.0",
     "tqdm",
     "torch>=1.11.0",
-    "numpy<2.0.0",
     "scikit-learn",
     "scipy",
     "huggingface-hub>=0.19.3",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,12 +1,15 @@
 [project]
 name = "sentence-transformers"
 version = "3.2.0.dev0"
-description = "Multilingual text embeddings"
-license = { file = "LICENSE" }
+description = "State-of-the-Art Text Embeddings"
+license = { text = "Apache 2.0" }
 readme = "README.md"
 authors = [
     { name = "Nils Reimers", email = "info@nils-reimers.de" },
-    { name = "Tom Aarsen" },
+    { name = "Tom Aarsen", email = "tom.aarsen@huggingface.co" },
+]
+maintainers = [
+    { name = "Tom Aarsen", email = "tom.aarsen@huggingface.co" }
 ]
 requires-python = ">=3.8"
 keywords = [
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -154,6 +154,7 @@ class SentenceTransformer(nn.Sequential, FitMixin):
         prompts: dict[str, str] | None = None,
         default_prompt_name: str | None = None,
         similarity_fn_name: str | SimilarityFunction | None = None,
+        mask_prompt: bool = False,
         cache_folder: str | None = None,
         trust_remote_code: bool = False,
         revision: str | None = None,
@@ -166,10 +167,11 @@ class SentenceTransformer(nn.Sequential, FitMixin):
         config_kwargs: dict[str, Any] | None = None,
         model_card_data: SentenceTransformerModelCardData | None = None,
     ) -> None:
-        # Note: self._load_sbert_model can also update `self.prompts` and `self.default_prompt_name`
+        # Note: self._load_sbert_model can also update `self.prompts`, `self.default_prompt_name` and `self.mask_prompt`
         self.prompts = prompts or {}
         self.default_prompt_name = default_prompt_name
         self.similarity_fn_name = similarity_fn_name
+        self.mask_prompt = mask_prompt
         self.trust_remote_code = trust_remote_code
         self.truncate_dim = truncate_dim
         self.model_card_data = model_card_data or SentenceTransformerModelCardData()
@@ -354,6 +356,8 @@ class SentenceTransformer(nn.Sequential, FitMixin):
         # suspect the user is using an INSTRUCTOR model.
         if model_name_or_path in ("hkunlp/instructor-base", "hkunlp/instructor-large", "hkunlp/instructor-xl"):
             self.set_pooling_include_prompt(include_prompt=False)
+        if self.mask_prompt:
+            self.set_pooling_mask_prompt(mask_prompt=True)
         elif (
             model_name_or_path
             and "/" in model_name_or_path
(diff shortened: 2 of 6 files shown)


ShengYun-Peng:
Hi @ArthurCamara, I'm reading the description of the PR and curious whether we need left padding in the text embedding models. I was also thinking about adding a left_pad argument following the LLM training procedure. But we may not need left_pad because text embeddings are not language modeling, and it only requires one forward pass to fetch the embedding, while language modeling typically requires multiple forward passes. Thus, there shouldn't be any gap (pad tokens) between the prompt and generated output in language modeling, leading to the left_pad operation. However, text embedding can take padding either on left or right, thus I think right_pad should already satisfy current text embedding models. Lmk your thoughts on this and correct me if I am wrong.

ArthurCamara:
> Hi @ArthurCamara, I'm reading the description of the PR and curious whether we need left padding in the text embedding models. I was also thinking about adding a left_pad argument following the LLM training procedure. But we may not need left_pad because text embeddings are not language modeling, and it only requires one forward pass to fetch the embedding, while language modeling typically requires multiple forward passes. Thus, there shouldn't be any gap (pad tokens) between the prompt and generated output in language modeling, leading to the left_pad operation. However, text embedding can take padding either on left or right, thus I think right_pad should already satisfy current text embedding models. Lmk your thoughts on this and correct me if I am wrong.

Hi @ShengYun-Peng! I agree that, in theory, either left or right-padding should be fine when training a text embedding model. However, there are some cases, specially when training a decoder-only model that uses causal attention (like most LLMs today), we usually want to use the `<EOS>` token as the sentence representation, instead of the first (e.g., the `[CLS]` token in BERT`) or the average of all tokens. This is really hard (and inefficient) to do if the batch sentences are right-padded, as the last token is now in some random position. 

This is the approach taken by, for instance, E5-Mistral and the [NV-Retriever](https://huggingface.co/nvidia/NV-Retriever-v1/blob/605680d4e3e267c0aa7b8d9026be373facdd6f68/modeling_nvretriever.py#L102) models. 
If you are not using the last token as the sentence representation (say, you have modified your attention mechanism to be bi-directional, like the NV-Embed-v2 model), then you should be fine using right padding.

ArthurCamara:
> > Hi @ArthurCamara, I'm reading the description of the PR and curious whether we need left padding in the text embedding models. I was also thinking about adding a left_pad argument following the LLM training procedure. But we may not need left_pad because text embeddings are not language modeling, and it only requires one forward pass to fetch the embedding, while language modeling typically requires multiple forward passes. Thus, there shouldn't be any gap (pad tokens) between the prompt and generated output in language modeling, leading to the left_pad operation. However, text embedding can take padding either on left or right, thus I think right_pad should already satisfy current text embedding models. Lmk your thoughts on this and correct me if I am wrong.
> 
> Hi @ShengYun-Peng! I agree that, in theory, either left or right-padding should be fine when training a text embedding model. However, there are some cases, specially when training a decoder-only model that uses causal attention (like most LLMs today), we usually want to use the `<EOS>` token as the sentence representation, instead of the first (e.g., the `[CLS]` token in BERT`) or the average of all tokens. This is really hard (and inefficient) to do if the batch sentences are right-padded, as the last token is now in some random position.
> 
> This is the approach taken by, for instance, E5-Mistral and the [NV-Retriever](https://huggingface.co/nvidia/NV-Retriever-v1/blob/605680d4e3e267c0aa7b8d9026be373facdd6f68/modeling_nvretriever.py#L102) models. If you are not using the last token as the sentence representation (say, you have modified your attention mechanism to be bi-directional, like the NV-Embed-v2 model), then you should be fine using right padding.

That being said, I've just noticed an issue in the code that will be harder to fix than I expected. So, I'm reverting these changes and keeping only the prompts to the evaluators in this PR for now.

tomaarsen:
> That being said, I've just noticed an issue in the code that will be harder to fix than I expected. So, I'm reverting these changes and keeping only the prompts to the evaluators in this PR for now.

Sounds good!
I'm glad that most "LLM-converted embedding models" use last-token pooling, as it's comparatively easy to implement.

I think this PR looks good at a glance. Do you want to mark it ready for review?

- Tom Aarsen

tomaarsen:
Thanks a bunch for this! I appreciate it.

- Tom Aarsen

ShengYun-Peng:
> > Hi @ArthurCamara, I'm reading the description of the PR and curious whether we need left padding in the text embedding models. I was also thinking about adding a left_pad argument following the LLM training procedure. But we may not need left_pad because text embeddings are not language modeling, and it only requires one forward pass to fetch the embedding, while language modeling typically requires multiple forward passes. Thus, there shouldn't be any gap (pad tokens) between the prompt and generated output in language modeling, leading to the left_pad operation. However, text embedding can take padding either on left or right, thus I think right_pad should already satisfy current text embedding models. Lmk your thoughts on this and correct me if I am wrong.
> 
> Hi @ShengYun-Peng! I agree that, in theory, either left or right-padding should be fine when training a text embedding model. However, there are some cases, specially when training a decoder-only model that uses causal attention (like most LLMs today), we usually want to use the `<EOS>` token as the sentence representation, instead of the first (e.g., the `[CLS]` token in BERT`) or the average of all tokens. This is really hard (and inefficient) to do if the batch sentences are right-padded, as the last token is now in some random position.
> 
> This is the approach taken by, for instance, E5-Mistral and the [NV-Retriever](https://huggingface.co/nvidia/NV-Retriever-v1/blob/605680d4e3e267c0aa7b8d9026be373facdd6f68/modeling_nvretriever.py#L102) models. If you are not using the last token as the sentence representation (say, you have modified your attention mechanism to be bi-directional, like the NV-Embed-v2 model), then you should be fine using right padding.

Thank you so much for the explanation! I agree that if we are using the eos token for embedding, left padding will make it easier.

Issue #2945, opened 2024-09-19
Best practice to train on multiple datasets with different prompts


ShengYun-Peng:
Thanks for the great work on facilitating the text embedding community!

I plan to train the instructor and other llm-based encoder models on multiple datasets. Since all of these models rely on different prompts on different embedding tasks. I'm curious what is the best way to prepend the prompt to the training dataset. 

1) use dataset.map to prepend different prompts for different datasets
2) change model.default_prompt_name in each batch according to the task

tomaarsen:
Hello!

Good question - this isn't clearly mentioned in the documentation anywhere.

The `default_prompt_name`, `prompt_name` and `prompt` options only affect the final inference of a trained model via `model.encode`.
So, if you want to train a model that can "understand" certain prompts that you want your users to apply, then you should use `dataset.map` to add those prompts to your training dataset.

Best of luck!

- Tom Aarsen

ShengYun-Peng:
Thank you for the quick response! After tracing the code, I notice that the entire encode function is not called in the training pipeline and the forward function of nn.Sequential is actually being called in the loss function, thus I will take your suggestion and implement the prompt logic while loading the dataset.

ShengYun-Peng:
A quick follow-up question on this: How do I exclude prompts in computing the mean embedding in the above scenario?

tomaarsen:
Hmm, I hadn't considered that yet. Via `model.encode` you can exclude it by setting `include_prompt` to `False` in the Pooling module, and then this section will trigger if someone passes a prompt:
https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/models/Pooling.py#L140-L141

For training, I think the easiest will be to write a custom Pooling module, e.g.:
```python
from __future__ import annotations

import json
import os
from typing import Any

import torch
from torch import Tensor, nn


class PoolingExcludingPrompts(nn.Module):
    """
    A pooling layer that computes the mean sentence embedding from a sequence of token embeddings,
    excluding the prompt tokens.
    """
    def __init__(self, word_embedding_dimension: int) -> None:
        super().__init__()
        self.word_embedding_dimension = word_embedding_dimension

    def forward(self, features: dict[str, Tensor]) -> dict[str, Tensor]:
        token_embeddings = features["token_embeddings"]
        attention_mask = (
            features["attention_mask"]
            if "attention_mask" in features
[...]

PR #2964, opened 2024-09-27
[`feat`] Trainer with prompts and prompt masking

ArthurCamara:
## Pull Request overview
- Adds support to including prompts in the `Trainer` class
- Supports masking the prompts in the `Pooling` when training.

## Details
Currently, the `encode` method of `SentenceTransformer` supports adding prompts (or instructions) dynamically to the sentences by passing either `prompt` or `prompt_name`. However, this is not supported when training, as mentioned in #2945, as it uses the `forward` method instead. 

This PR implements a similar functionality to the `Trainer`, by adding `prompt` parameter that can be:
  - `str`: The prompt will be appended to all sentences in the dataset
  - `dict[str, str]`: If the keys are column names, it will append the prompt to the respective column. If the training dataset is a dictionary of datasets, and the dictionary keys are names of the datasets, it will add the prompt to all the columns of the respective dataset. 
  - `dict[str, dict[str, str]]`: Same as above, but assumes the first level is the dataset name and the second level are the column names.

As the prompts can be dynamic (changing for each dataset and column), they are injected in the sentences by the `get_train|test|eval|_dataloader` methods, by calling `add_prompts_to_dataset`, which solves for each dataset and column which prompt to inject. 

Finally, the `add_prompts_to_dataset` also adds `<column_name>_prompt_length` columns that, when passed to `Pooling` method with `include_prompt=False`, will mask the instructions properly as well. (currently this is only explicitly for `Instructor` models, but can be set by the user by calling `model.set_pooling_include_prompt(include_prompt=False)`


tomaarsen:
Hello!

Thanks for this PR. I rebased it to get rid of the leftover commits that aren't necessary here. 
I have a few hesitations with the current approach, although I do quite like the idea of being able to specify prompts to use during training apart from manually adding them in your dataset(s). My current hesitations:
1. Adding a column to the entire dataset before training will be incompatible with `datasets` IterableDataset (i.e., `load_dataset("...", streaming=True)`. 
2. I'm in theory okay with adding a `..._prompt_length` column: I recognize that it's crucial to get this information if `include_prompt` is False in the `Pooling`. However, I have two notes:
   * Could we e.g. only add the information if the Pooling module (if it exists) actually has `include_prompt=False`?
   * Could we perhaps not add an entire column, but instead create a nested dictionary with dataset names mapping to column names mapping to prompt lengths? Dataset names should be a column if there's multiple datasets.

Could we perhaps add the prompts (and prompt lengths) in the data collator? E.g. right here: https://github.com/ArthurCamara/sentence-transformers/blob/bf9eb803ce2dda26a8ef903c33d80cd1fcb55a3d/sentence_transformers/data_collator.py#L50-L56

The data collator knows the dataset name, the column name (see the snippet), and should then be able to use that information to "on the fly" prepend the prompts. In a perfect world we could even only tokenize the prompts once, but that gets complicated with padding and truncation, so it's better to keep it simpler.

I also like your idea that `prompt` can be multiple things: a single prompt, a prompt per column, or a prompt per column per dataset. I do think `prompts` is a bit better though, because that's what we use in the `encode` etc. 

I'm curious to hear your thoughts on this.

- Tom Aarsen

ArthurCamara:
> Hello!
> 
> Thanks for this PR. I rebased it to get rid of the leftover commits that aren't necessary here. I have a few hesitations with the current approach, although I do quite like the idea of being able to specify prompts to use during training apart from manually adding them in your dataset(s). My current hesitations:
> 
> 1. Adding a column to the entire dataset before training will be incompatible with `datasets` IterableDataset (i.e., `load_dataset("...", streaming=True)`.
> 2. I'm in theory okay with adding a `..._prompt_length` column: I recognize that it's crucial to get this information if `include_prompt` is False in the `Pooling`. However, I have two notes:
>    
>    * Could we e.g. only add the information if the Pooling module (if it exists) actually has `include_prompt=False`?
>    * Could we perhaps not add an entire column, but instead create a nested dictionary with dataset names mapping to column names mapping to prompt lengths? Dataset names should be a column if there's multiple datasets.

> Could we perhaps add the prompts (and prompt lengths) in the data collator? E.g. right here: https://github.com/ArthurCamara/sentence-transformers/blob/bf9eb803ce2dda26a8ef903c33d80cd1fcb55a3d/sentence_transformers/data_collator.py#L50-L56
> 
> The data collator knows the dataset name, the column name (see the snippet), and should then be able to use that information to "on the fly" prepend the prompts. In a perfect world we could even only tokenize the prompts once, but that gets complicated with padding and truncation, so it's better to keep it simpler.

This was one of the things I was considering, to change the Collator instead of the dataset itself. But I had issues with Accelerator and DDP before when the data was not exclusively tensors (i.e., strings), but I think we can walk around it within the collator. I will give it a shot and let you know. 

> I also like your idea that `prompt` can be multiple things: a single prompt, a prompt per column, or a prompt per column per dataset. I do think `prompts` is a bit better though, because that's what we use in the `encode` etc.

Agreed. =) 

> I'm curious to hear your thoughts on this.
> 
> * Tom Aarsen

ArthurCamara:
@tomaarsen There you go. I've changed the prompting logic to the `data_collator`. The `prompt_lenght` column will only be added if the Pooler has `include_prompt` set to `False`.

What needed to be kept in the `Trainer`, however is the added logic to decide if `we should call `add_dataset_name_column` or not. To clean up the code a bit, I've moved all these checks into the `maybe_add_dataset_name_column` method. Let me know if you think it is a good idea or not.

ShengYun-Peng:
Hi @ArthurCamara @tomaarsen , thank you both for the PR, and sorry for the late reply! I have implemented a local trainer that supports the prompt logic following the [instructor code](https://github.com/xlang-ai/instructor-embedding). It's a bit different from the logic of current PR, and I would like to hear your thoughts on it. 

The idea is only to change the dataset and the Transformer of the Sentence Transformer model. The dataset part will be similar to yours, which is loading the prompt/instruction and the user sentence separately. However, the Transformer tokenizer will compute an instruction mask based on the tokenized length of the prompt/instruction. Thus, the input dict contains three items: input_ids, attention_mask, and instruction_mask. After the Transformer finishes the forward pass, I update the attention_mask with the element-wise product of the attention_mask and the instruction_mask. Thus the instruction is excluded in the final embedding. Of course, if the user wants to include the instruction in the final embedding, we can add a flag in the argument. 

I'm sure your code will also work. From my first glance, it seems like the above logic is implemented in the Pooling rather than the Transformer layer of the Sentence Transformer. It would be great if you can share your high-level design and thought process. Thanks!

- Anthony Peng

ShengYun-Peng:
Thus, in my implementation, each sentence in a dataset is a pair of strings. Take STS for an example, one input sample was {"sentence1": "A plane is taking off.", "sentence2": "An air plane is taking off.", "score": 1} previously. Now the input is {"sentence1": ["Embed the semantic meaning of the sentence: ", "A plane is taking off."], "sentence2": ["Embed the semantic meaning of the sentence: ", "An air plane is taking off."], "score": 1}. It will break the [model card creation](https://github.com/UKPLab/sentence-transformers/blob/b6f114288f8cf1edc6eb7739d95b7adf15d49fcb/sentence_transformers/model_card.py#L434) since there is no `.values` for a sequence object in the dataset.

ArthurCamara:
> Hi @ArthurCamara @tomaarsen , thank you both for the PR, and sorry for the late reply! I have implemented a local trainer that supports the prompt logic following the [instructor code](https://github.com/xlang-ai/instructor-embedding). It's a bit different from the logic of current PR, and I would like to hear your thoughts on it.

> The idea is only to change the dataset and the Transformer of the Sentence Transformer model. The dataset part will be similar to yours, which is loading the prompt/instruction and the user sentence separately. However, the Transformer tokenizer will compute an instruction mask based on the tokenized length of the prompt/instruction. Thus, the input dict contains three items: input_ids, attention_mask, and instruction_mask. After the Transformer finishes the forward pass, I update the attention_mask with the element-wise product of the attention_mask and the instruction_mask. Thus the instruction is excluded in the final embedding. Of course, if the user wants to include the instruction in the final embedding, we can add a flag in the argument.
> 
> I'm sure your code will also work. From my first glance, it seems like the above logic is implemented in the Pooling rather than the Transformer layer of the Sentence Transformer. It would be great if you can share your high-level design and thought process. Thanks!
> 
> * Anthony Peng

Hi @ShengYun-Peng! Sorry for the late reply too. Last few weeks have been quite busy. 

Either approach would work. The only difference is where we are computing the mask for the instruction. IMO, the instruction is part of the dataset, not of the model itself. Thus, makes more sense if its is computed in the dataset (or collator) level, rather than by the Transformer model itself. This is also closer to how the `encode()` method works, where the user only needs to provide the prompt (or the name of the prompt) to train method.

The biggest downside to the approach here is the added column to the dataset, which can be somewhat wasteful. Perhaps it is possible to cache these and just include the prompt on the last step of computation.

JosephGatto:
Hi thanks for implementing this. Any guide on how to fine-tune with prompts?

tomaarsen:
Hello!

Until this is integrated, I would recommend manually adding the prompts to your training datasets. E.g.:
```python
from typing import Dict, List, Any

def prepend_prompt(batch: Dict[str, List[Any]], prompts: Dict[str, str] | None = None) -> Dict[str, List[Any]]:
    if not prompts:
        return batch

    for column_name, prompt in prompts.items():
        batch[column_name] = [prompt + value for value in batch[column_name]]
    return batch

train_dataset = load_dataset("sentence-transformers/natural-questions", split="train")
train_dataset = train_dataset.map(
    prepend_prompt,
    batched=True,
    fn_kwargs={"prompts": {"question": "Represent this sentence for searching relevant passages: "}}
)
print(train_dataset[0])
# {'query': 'Represent this sentence for searching relevant passages: when did richmond last play in a preliminary final', 'answer': "Richmond Football Club Richmond began 2017 with 5 straight wins, a feat it had not achieved since 1995. A series of close losses hampered the Tigers throughout the middle of the season, including a 5-point loss to the Western Bulldogs, 2-point loss to Fremantle, and a 3-point loss to the Giants. Richmond ended the season strongly with convincing victories over Fremantle and St Kilda in the final two rounds, elevating the club to 3rd on the ladder. Richmond's first final of the season against the Cats at the MCG attracted a record qualifying final crowd of 95,028; the Tigers won by 51 points. Having advanced to the first preliminary finals for the first time since 2001, Richmond defeated Greater Western Sydney by 36 points in front of a crowd of 94,258 to progress to the Grand Final against Adelaide, their first Grand Final appearance since 1982. The attendance was 100,021, the largest crowd to a grand final since 1986. The Crows led at quarter time and led by as many as 13, but the Tigers took over the game as it progressed and scored seven straight goals at one point. They eventually would win by 48 points – 16.12 (108) to Adelaide's 8.12 (60) – to end their 37-year flag drought.[22] Dustin Martin also became the first player to win a Premiership medal, the Brownlow Medal and the Norm Smith Medal in the same season, while Damien Hardwick was named AFL Coaches Association Coach of the Year. Richmond's jump from 13th to premiers also marked the biggest jump from one AFL season to the next."}
```
And the rest is the same as the normal training: https://sbert.net/docs/sentence_transformer/training_overview.html

- Tom Aarsen

JosephGatto:
Hey thanks so much for the quick reply. My main concern here would be if pooling is being done on just the text (and excluding the prompt). I believe in the INSTRUCTOR paper they do not include the embeddings of the prompt during mean pooling. Would this solution take care of that?

tomaarsen:
Indeed, my solution only works if you're including the prompt in the pooling. If you're not, i.e. with setting this to False: https://github.com/UKPLab/sentence-transformers/blob/b9316f90cdd8dc9e0c5c6d99353f89c4621c8828/sentence_transformers/models/Pooling.py#L60

Then you must use this PR. You can use:
```
pip install git+https://github.com/ArthurCamara/sentence-transformers@trainer-with-prompt-masking
```
and then use the [regular training](https://sbert.net/docs/sentence_transformer/training_overview.html) with one extra parameter in the `SentenceTransformerTrainer`:
```python
    model=model,
    args=args,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    loss=loss,
    prompts={
        "query": "Represent this sentence for searching relevant passages: ",
    },
    evaluator=dev_evaluator,
)
```
The `prompts` can be:
1. a prompt string
2. a column name to prompt mapping
3. a dataset to prompt mapping (if you use a dataset dict to train on multiple datasets simultaneously)
4. a dataset to column name to prompt mapping (i.e. nested dicts, only if you use a dataset dict to train on multiple datasets simultaneously)

I do want to warn you that I'm about to fully overhaul this PR, although the usage will remain the same.

- Tom Aarsen

JosephGatto:
Thanks so much. And if I was interested in training with dynamic prompts (unique prompt per sample) would that be possible with the methods you described?

tomaarsen:
Unique per sample is not possible here without subclassing the Trainer, no. You could use a unique sample per dataset, if that helps. I didn't think that a unique prompt per sample was a notable use case, so I didn't think to integrate it.

JosephGatto:
Got it. Thank you!

tomaarsen:
Heya @ArthurCamara,

I've overhauled the prompt prepending once more, as I still had some slight concerns with the previous implementations after some experimentation. You have worked on 2 implementations, and I'm now proposing a third as well:
1. 'Greedily' `.map` over each dataset to add the prompt string to each dataset.
2. 'Lazily' prepends prompts in the data collator in the Trainer.
3. Use `.set_transform` for `Dataset(Dict)` and `.map` for `IterableDataset(Dict)` to add the prompt string to each dataset.

I had concerns with the first two:
1. I'm wary that this results in large memory usage and/or cache files.
2. During model card generation, I sample from the datasets to use in the model card ([example](https://huggingface.co/minhdang/gte-base-law-matryoshka#training-dataset)), I'd also like for the prompts to be included, which isn't the case if the prompt prepending only exists in the Trainer.

After getting some valuable recommendations by the Datasets team and @lhoestq in particular, I'm now using `.set_transform` and `.map` to lazily apply 1) the prompts (if provided), 2) prompt lengths (if needed for pooling), and 3) dataset name (if needed for determining which loss to use). The implementation now lives as a 1-time "update" of the provided train/eval datasets, so the model card can easily fetch samples that include the prompts. 

I've also trained 2 near-identical models:
* https://huggingface.co/tomaarsen/mpnet-base-nq: [mpnet-base](https://huggingface.co/microsoft/mpnet-base) finetuned on [natural-questions](https://huggingface.co/datasets/sentence-transformers/natural-questions) and evaluated on [NanoBEIR](https://huggingface.co/collections/zeta-alpha-ai/nanobeir-66e1a0af21dfd93e620cd9f6).
* https://huggingface.co/tomaarsen/mpnet-base-nq-prompts: See above, except with `prompts={"query": "query: ", "answer": "document: "}` & `query_prompt` and `corpus_prompt` provided to the NanoBEIR evaluator. Identical seeds, etc.

The former consistently performs slightly worse than the model with the `prompts`:
image

Also, the prompts model shows the prompts in the model card easily: https://huggingface.co/tomaarsen/mpnet-base-nq-prompts#natural-questions

Lastly, I built an extensive training suite for this feature because there are a LOT of moving parts between training, evaluation, iterable datasets, and the various prompt formats. 

I'm curious about your thoughts on my proposal @ArthurCamara, as I know you're using this yourself too!

- Tom Aarsen

ArthurCamara:
> Heya @ArthurCamara,
> 
> I've overhauled the prompt prepending once more, as I still had some slight concerns with the previous implementations after some experimentation. You have worked on 2 implementations, and I'm now proposing a third as well:
> 
> 1. 'Greedily' `.map` over each dataset to add the prompt string to each dataset.
> 2. 'Lazily' prepends prompts in the data collator in the Trainer.
> 3. Use `.set_transform` for `Dataset(Dict)` and `.map` for `IterableDataset(Dict)` to add the prompt string to each dataset.
> 
> I had concerns with the first two:
> 
> 1. I'm wary that this results in large memory usage and/or cache files.
> 2. During model card generation, I sample from the datasets to use in the model card ([example](https://huggingface.co/minhdang/gte-base-law-matryoshka#training-dataset)), I'd also like for the prompts to be included, which isn't the case if the prompt prepending only exists in the Trainer.

Adding the prompts to the model card is something very useful that I haven't thought of. Nice.

> 
> After getting some valuable recommendations by the Datasets team and @lhoestq in particular, I'm now using `.set_transform` and `.map` to lazily apply 1) the prompts (if provided), 2) prompt lengths (if needed for pooling), and 3) dataset name (if needed for determining which loss to use). The implementation now lives as a 1-time "update" of the provided train/eval datasets, so the model card can easily fetch samples that include the prompts.
> 

Nice to learn something new. Didn't know about `set_transform` This is a cleaner solution than doing multiple passes over the datasets. 

> I've also trained 2 near-identical models:
> 
> * https://huggingface.co/tomaarsen/mpnet-base-nq: [mpnet-base](https://huggingface.co/microsoft/mpnet-base) finetuned on [natural-questions](https://huggingface.co/datasets/sentence-transformers/natural-questions) and evaluated on [NanoBEIR](https://huggingface.co/collections/zeta-alpha-ai/nanobeir-66e1a0af21dfd93e620cd9f6).
> * https://huggingface.co/tomaarsen/mpnet-base-nq-prompts: See above, except with `prompts={"query": "query: ", "answer": "document: "}` & `query_prompt` and `corpus_prompt` provided to the NanoBEIR evaluator. Identical seeds, etc.
> The former consistently performs slightly worse than the model with the `prompts`: image
> 
> Also, the prompts model shows the prompts in the model card easily: https://huggingface.co/tomaarsen/mpnet-base-nq-prompts#natural-questions
> 

Neat. I like the way prompting helps to disentangle the representations of query and documents even in smaller models

> Lastly, I built an extensive training suite for this feature because there are a LOT of moving parts between training, evaluation, iterable datasets, and the various prompt formats.
> 
> I'm curious about your thoughts on my proposal @ArthurCamara, as I know you're using this yourself too! And one final question:
> 
> * Do you think that `prompts` should be a parameter in `SentenceTransformerTrainer` or in `SentenceTransformerTrainingArguments`?
> 

Good question. I want to say it should be in the Arguments, so it can be easily swapped out when testing with different configurations. But I'm not sure how of a good UX it will be to pass a double-nested dictionary as an argument to training script (of course, reading from a json/yaml file is also an option).

> * Tom Aarsen

tomaarsen:
Thanks a bunch for spearheading this. I didn't expect that the prompts would have such a notable impact (0.66% and 0.90% relative NDCG@10 across mpnet-base and bert-base-uncased, respectively), but I'm glad that they do. 

This will be included as one of the 4 major features in Monday's v3.3 release, alongside the NanoBEIREvaluator which will be another major feature. I really appreciate your work on these.

- Tom Aarsen

PR #2965, opened 2024-09-27
"query" "answer" columns selection in hard negative mining function 

alperctnkaya:
Queries are now directly assigned from the "query" column of the dataset.
Positives are assigned from the "answer" column for clarity and consistency.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -67,6 +67,7 @@ exclude_patterns = [
 intersphinx_mapping = {
     "datasets": ("https://huggingface.co/docs/datasets/main/en/", None),
     "transformers": ("https://huggingface.co/docs/transformers/main/en/", None),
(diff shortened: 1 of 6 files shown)


tomaarsen:
Hello!

The choice to use the first two columns regardless of their names was a bit deliberate: it allows people to use any kinds of pair datasets, including e.g. similarity tasks like https://huggingface.co/datasets/sentence-transformers/all-nli, or multilingual paraphrase datasets like https://huggingface.co/datasets/sentence-transformers/parallel-sentences-talks, but also just question-answer dataset with different columns like "question" and "passage".

Perhaps if it's not clear from the docstring etc., then we should try and fix this in another way via documentation?

- Tom Aarsen

alperctnkaya:
Hi Tom,

Thanks for the clarification! I see the benefit of using the first two columns for flexibility with different datasets. However, I think passing column names as parameters would make it clearer and more explicit for users, ensuring it works across various datasets without confusion.

PR #2977, opened 2024-10-10
Refine mine_hard_negatives arguments

bakrianoo:
This PR to refine `mine_hard_negatives` function.

### Update 1

support the passing of the column names to make the function more flexible for any dataset structure.
The function now accepts:
```
`query_column_name` (str, optional): The column name in `dataset` that contains the queries. Defaults to None, in which case the first column in `dataset` will be used.
`positive_column_name` (str, optional): The column name in `dataset` that contains the positive candidates. Defaults to None, in which case the second column in `dataset` will be used.
```

### Update 2

The function now allows passing the `min_score` to accept for a negative candidate. I found controlling this argument can produce more confident results.

```
`min_score` (float, optional): Minimum score to consider as a negative. Defaults to None.
```

### Update 3

check if the tensor is empty before applying the reduction functions.

### DIFF
--- a/sentence_transformers/util.py
+++ b/sentence_transformers/util.py
@@ -963,11 +963,11 @@ def mine_hard_negatives(
             ("mean", torch.mean),
             ("median", torch.median),
             ("std", torch.std),
-            ("min", torch.min),
-            ("25%", lambda scores: torch.quantile(scores.float(), q=0.25)),
-            ("50%", lambda scores: torch.quantile(scores.float(), q=0.5)),
-            ("75%", lambda scores: torch.quantile(scores.float(), q=0.75)),
-            ("max", torch.max),
+            ("min", lambda scores: torch.min(scores) if scores.numel() > 0 else float('inf')),
+            ("25%", lambda scores: torch.quantile(scores.float(), q=0.25) if scores.numel() > 0 else float('inf')),
+            ("50%", lambda scores: torch.quantile(scores.float(), q=0.5) if scores.numel() > 0 else float('inf')),
+            ("75%", lambda scores: torch.quantile(scores.float(), q=0.75) if scores.numel() > 0 else float('inf')),
+            ("max", lambda scores: torch.max(scores) if scores.numel() > 0 else float('-inf')),
         ]:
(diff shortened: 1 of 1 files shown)


sentence_transformers/util.py
```diff
@@ -621,13 +625,16 @@ def mine_hard_negatives(
     Args:
         dataset (Dataset): A dataset containing (anchor, positive) pairs.
         model (SentenceTransformer): A SentenceTransformer model to use for embedding the sentences.
+        query_column_name (str, optional): The column name in `dataset` that contains the queries. Defaults to None, in which case the first column in `dataset` will be used.
```
tomaarsen, line 628:
```suggestion
        anchor_column_name (str, optional): The column name in `dataset` that contains the anchors or queries. Defaults to None, in which case the first column in `dataset` will be used.
```
Could we rename this to `anchor`, fitting with the `anchor, positive, negative` scheme? That way, people don't get the impression that this can only be used for "query-document" information retrieval datasets, but also e.g. for semantic textual similarity (although I imagine 90% of usage is for IR).

tomaarsen:
Hello!

Thanks a bunch for this, I quite like what this looks like already. I made one small comment regarding the `query_column_name` name, but otherwise it's looking good.
Could you also do:
```
pip install pre-commit
pre-commit install
pre-commit run --all
```
This should fix the code quality action: https://github.com/UKPLab/sentence-transformers/actions/runs/11270007656/job/31340462266?pr=2977

- Tom Aarsen

sentence_transformers/util.py
bakrianoo: Thank you.
Would you suggest renaming:
`query_column_name`: to `anchor_column_name`
right?

tomaarsen: Indeed! I think it looks good now

bakrianoo:
@tomaarsen 
I did the `pre-commit` exercise. Please check.

tomaarsen:
Well done, thanks a bunch for this! It'll be included in the upcoming release.

- Tom Aarsen

PR #2992, opened 2024-10-17
[fix] Added model2vec import compatible with current and newer version

Pringled:
Hi!

# Pull Request overview
- Make sentence-transformers compatible with current and newer version of Model2Vec

# Details
Starting from model2vec version 0.3, inference and distillation will be split up since inference will only need Numpy. This makes it much faster to initialize and install, since there is no dependency on Torch anymore for inference. This PR makes sentence-transformers compatible with both the old and new version, including an updated ImportError message for from_disillation. 

Relevant Model2Vec commit: https://github.com/MinishLab/model2vec/commit/010a7b7347e72e6fa5a92e52314674359253a409

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -61,7 +61,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install '.[train, onnx, openvino]'
+          python -m pip install '.[train, onnx, openvino, dev]'
 
       - name: Install model2vec
         run: python -m pip install model2vec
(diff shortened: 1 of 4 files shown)


tomaarsen:
Hello!

This looks solid I think! Also, torch-less inference on `model2vec` itself is quite nice, well done.

Have you tested `from_model2vec` with the latest `model2vec`? I think this could fail: https://github.com/Pringled/sentence-transformers/blob/92a9f27136176b3a2cfc8b17a5507412d177bd73/sentence_transformers/models/StaticEmbedding.py#L202-L203

- Tom Aarsen

tomaarsen:
I added some simple tests to make sure that things don't break in the future.

Pringled:
@tomaarsen good catch, those need to be converted! Will update the code

Pringled:
Nice! That should work with both the old and new versions 💪

PR #2993, opened 2024-10-17
[`enh`] Add Support for multiple adapters on Transformers-based models

carlesonielfa:
Hi!
## Pull Request overview
Adds support for loading, activating and deactivating adapters for a SentenceTransformer model when its underlying model is Transformers-based.

## Details
Adds a wrapper mixin so that the methods to manipulate adapters are exposed on the SentenceTransformer model. It includes checks to ensure that the underlying model supports it.

I wanted to implement it in a manner that won't break or require maintenance if the transformers library changes the signature for these methods, this however compromises typing and hints, so if that is more of a concern than maintenance, the implementation could be changed to a more hard-coded one.

As it is only a wrapper, the functionality follows that documented on the [hf docs for transformers ](https://huggingface.co/docs/transformers/v4.45.2/en/peft#load-adapters-with--peft).

### Usage example

Pre-trained adapter
```python

model = SentenceTransformer("my_org/mymodel")
model.load_adapter("my_org/my_pretrained_adapter", "my_adapter")
```

Adapter from scratch for training purposes
```python
from peft import LoraConfig, TaskType

model = SentenceTransformer("my_org/mymodel")

peft_config = LoraConfig(
    target_modules=["query", "key", "value"],
    task_type=TaskType.FEATURE_EXTRACTION,
    inference_mode=False,
    r=8,
    lora_alpha=32,
    lora_dropout=0.1
)
model.add_adapter(peft_config)

```

### Additional considerations
- Consider incorporating adapter switching to the `SentenceTransformer.encode` function
- Test with a pretrained model is commented out on this PR waiting for feedback from a maintainer on if an adapter model should be uploaded to the sentence-transformers hf hub
- Might be worth including documentation about the feature

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,7 @@ train = ["datasets", "accelerate>=0.20.3"]
 onnx = ["optimum[onnxruntime]>=1.23.1"]
 onnx-gpu = ["optimum[onnxruntime-gpu]>=1.23.1"]
 openvino = ["optimum-intel[openvino]>=1.20.0"]
-dev = ["datasets", "accelerate>=0.20.3", "pre-commit", "pytest", "pytest-cov"]
+dev = ["datasets", "accelerate>=0.20.3", "pre-commit", "pytest", "pytest-cov", "peft"]
 
 [build-system]
 requires = ["setuptools>=42", "wheel"]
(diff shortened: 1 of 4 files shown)


tomaarsen:
I've pushed back the release from today to Monday, should be a bit safer.

PR #3000, opened 2024-10-18
[`peft`] If AutoModel is wrapped with PEFT for prompt learning, then extend the attention mask

### LINKED ISSUE #2995: `Transformer` Class in HF Models Does Not Properly Update `attention_mask` with Prompt Tuning, Leading to Mismatch in Downstream Layers
#### Problem Description:

I'm using the following `Transformer` class in combination with **PEFT adapters** (e.g., prompt tuning) that modify the sequence length by adding virtual tokens. This creates a mismatch between the size of the `token_embeddings` and the `attention_mask`, causing downstream layers (e.g., pooling layers) to break due to this inconsistency.

Here is the `Transformer` class that is supposed to handle the attention mask:

```python
class Transformer(nn.Module):
    def forward(self, features: dict[str, torch.Tensor], **kwargs) -> dict[str, torch.Tensor]:
        """Returns token_embeddings, cls_token"""
        trans_features = {"input_ids": features["input_ids"], "attention_mask": features["attention_mask"]}
        if "token_type_ids" in features:
            trans_features["token_type_ids"] = features["token_type_ids"]

        output_states = self.auto_model(**trans_features, **kwargs, return_dict=False)
        output_tokens = output_states[0]

        features.update({"token_embeddings": output_tokens, "attention_mask": features["attention_mask"]})

        if self.auto_model.config.output_hidden_states:
            all_layer_idx = 2
            if len(output_states) < 3:  # Some models only output last_hidden_states and all_hidden_states
                all_layer_idx = 1

            hidden_states = output_states[all_layer_idx]
            features.update({"all_layer_embeddings": hidden_states})

        return features
```

#### Issue:

In the code, it is expected that the **`attention_mask`** is updated and propagated correctly. However, the `features.update({"token_embeddings": output_tokens, "attention_mask": features["attention_mask"]})` line does **not actually update the `attention_mask`** when virtual tokens are added by the PEFT adapters. This line simply adds the existing `attention_mask` to the updated features without reflecting the changes in sequence length, which leads to a mismatch when the downstream layers (such as pooling) use this outdated `attention_mask`.

When using prompt-tuning adapters (like in the example below), the sequence length increases by `num_virtual_tokens`, but the `attention_mask` remains the same, causing a **dimension mismatch** when passing `token_embeddings` and `attention_mask` to subsequent layers, such as the pooling layer.

Here’s how I’m using prompt-tuning with this model:

```python
sentence_transformer = SentenceTransformer(model_name)

NUM_VIRTUAL_TOKENS = 1
peft_config = PromptTuningConfig(
    task_type=TaskType.FEATURE_EXTRACTION,
    prompt_tuning_init=PromptTuningInit.RANDOM,
    num_virtual_tokens=NUM_VIRTUAL_TOKENS
)

sentence_transformer._modules["0"].auto_model = get_peft_model(
    sentence_transformer._modules["0"].auto_model, peft_config
)
```

### Error:

When I run the following code:

```python
batch_size = 1
seq_len = 53
random_input_ids = torch.randint(0, 30522, (batch_size, seq_len)).to("cuda:0")
random_attention_mask = torch.ones(batch_size, seq_len).to("cuda:0")

outputs = sentence_transformer({"input_ids": random_input_ids, "attention_mask": random_attention_mask})
```

I get this error:

```
RuntimeError: The expanded size of the tensor (54) must match the existing size (53) at non-singleton dimension 1. Target sizes: [64, 54, 768]. Tensor sizes: [64, 53, 1]
```

This error occurs because the `token_embeddings` have the length updated to 54 (53 original tokens + 1 virtual token from prompt tuning), but the `attention_mask` remains at length 53, causing the mismatch in the next layer (e.g., Pooling).

### Cause:

The root cause is that, while the transformer model correctly modifies the `token_embeddings` by adding virtual tokens, the `attention_mask` is **not updated** to account for the new tokens. Specifically, the line `features.update({"attention_mask": features["attention_mask"]})` does not actually modify the `attention_mask` in line with the new sequence length, even though the prompt tuning has added tokens to the embeddings.

### Workaround:

To work around this issue, I created a **forward hook** that manually modifies the `attention_mask` before the pooling layer is called:

```python
def add_prefix_tokens_to_attention_mask(module, features: dict[str, torch.Tensor]):
    attention_mask = features[0]["attention_mask"]
    batch_size = features[0]["token_embeddings"].size(0)
    
    num_virtual_tokens = NUM_VIRTUAL_TOKENS
    prefix_attention_mask = torch.ones(batch_size, num_virtual_tokens).to(attention_mask.device)
    
    updated_attention_mask = torch.cat([prefix_attention_mask, attention_mask], dim=1)
    
    features[0].update(
        {
         "attention_mask": updated_attention_mask   
        }
    )
    
    return features

sentence_transformer._modules["1"].register_forward_pre_hook(add_prefix_tokens_to_attention_mask)
```

While this workaround fixes the problem, it feels hacky and is error-prone. Ideally, the `Transformer` class itself should properly update the `attention_mask` to reflect the changes in sequence length when virtual tokens are added.

### Suggested Solution:

To fix this, the `Transformer` class should ensure that the `attention_mask` is updated in sync with the `token_embeddings`. This can be done by modifying the logic to properly handle prompt-tuning or other sequence-altering adapters by ensuring the `attention_mask` reflects the updated sequence length.

### Summary:

- The `Transformer` class should properly handle updates to `attention_mask` when using prompt-tuning or sequence-altering adapters.
- Without this, there will be runtime errors due to dimension mismatches between `token_embeddings` and `attention_mask` in downstream layers.
- A workaround using forward hooks works but is not an ideal long-term solution.

This behavior could lead to more bugs if developers modify the `attention_mask` expecting it to be properly updated, but it remains outdated in the current design.

tomaarsen:
Resolves #2995, resolves https://github.com/huggingface/peft/issues/2154

Hello!

## Pull Request overview
* If AutoModel is wrapped with PEFT for prompt learning, then extend the attention mask

## Details
Sentence Transformer models are sometimes trained with the `AutoModel` wrapped in PEFT, as that can lead to decreased computation cost while training. In particular, when PEFT with prompt learning is used, then virtual tokens (or rather, just input_embeds) are added to the model, and the `attention_mask` is updated before the base `AutoModel` is called.

However, then the attention mask used in the Pooling module won't be updated. This PR fixes that.

## Concern
My primary concern now is that the model doesn't seem to be able to train well:
```python
import random
import logging
from datasets import load_dataset, Dataset
    SentenceTransformer,
    SentenceTransformerTrainer,
    SentenceTransformerModelCardData,
)
from sentence_transformers.losses import MultipleNegativesRankingLoss
from sentence_transformers.models import Pooling, Transformer
from sentence_transformers.evaluation import InformationRetrievalEvaluator
from peft import get_peft_model, PromptTuningConfig, TaskType, PromptTuningInit

logging.basicConfig(
    format="%(asctime)s - %(message)s", datefmt="%Y-%m-%d %H:%M:%S", level=logging.INFO
)

# 1. Load a model to finetune with 2. (Optional) model card data
cls_pooling = False
if cls_pooling:
    transformer = Transformer("microsoft/mpnet-base")
    pooling = Pooling(transformer.get_word_embedding_dimension(), "cls")
    model = SentenceTransformer(
        modules=[transformer, pooling],
        model_card_data=SentenceTransformerModelCardData(
            language="en",
            license="apache-2.0",
            model_name="MPNet base trained on GooAQ triplets",
        ),
    )
else:
    model = SentenceTransformer(
        "microsoft/mpnet-base",
        model_card_data=SentenceTransformerModelCardData(
            language="en",
            license="apache-2.0",
            model_name="MPNet base trained on GooAQ triplets",
        ),
    )

# Apply PEFT with PromptTuningConfig
peft_config = PromptTuningConfig(
    task_type=TaskType.FEATURE_EXTRACTION,
    prompt_tuning_init=PromptTuningInit.RANDOM,
    num_virtual_tokens=1
)
model[0].auto_model = get_peft_model(model[0].auto_model, peft_config)

# 3. Load a dataset to finetune on
dataset = load_dataset("sentence-transformers/gooaq", split="train")
dataset = dataset.add_column("id", range(len(dataset)))
dataset_dict = dataset.train_test_split(test_size=10_000, seed=12)
train_dataset: Dataset = dataset_dict["train"]
eval_dataset: Dataset = dataset_dict["test"]

# 4. Define a loss function

# 5. (Optional) Specify training arguments
run_name = "mpnet-base-gooaq-peft"
    # Required parameter:
    num_train_epochs=1,
    learning_rate=2e-5,
    warmup_ratio=0.1,
    bf16=True,  # Set to True if you have a GPU that supports BF16
    batch_sampler=BatchSamplers.NO_DUPLICATES,  # MultipleNegativesRankingLoss benefits from no duplicate samples in a batch
    eval_strategy="steps",
    eval_steps=1000,
    save_strategy="steps",
    save_steps=1000,
    save_total_limit=2,
    logging_steps=250,
    logging_first_step=True,
    run_name=run_name,  # Will be used in W&B if `wandb` is installed
)

# 6. (Optional) Create an evaluator & evaluate the base model
# The full corpus, but only the evaluation queries
# corpus = dict(zip(dataset["id"], dataset["answer"]))
random.seed(12)
queries = dict(zip(eval_dataset["id"], eval_dataset["question"]))
corpus = (
    {qid: dataset[qid]["answer"] for qid in queries} |
    {qid: dataset[qid]["answer"] for qid in random.sample(range(len(dataset)), 20_000)}
)
relevant_docs = {qid: {qid} for qid in eval_dataset["id"]}
dev_evaluator = InformationRetrievalEvaluator(
    corpus=corpus,
    queries=queries,
    relevant_docs=relevant_docs,
    show_progress_bar=True,
    name="gooaq-dev",
)
dev_evaluator(model)

# 7. Create a trainer & train
    model=model,
    args=args,
    train_dataset=train_dataset.remove_columns("id"),
    eval_dataset=eval_dataset.remove_columns("id"),
    loss=loss,
    evaluator=dev_evaluator,
)
trainer.train()

# (Optional) Evaluate the trained model on the evaluator after training
dev_evaluator(model)

# 8. Save the trained model
model.save_pretrained(f"models/{run_name}/final")

model.push_to_hub(run_name, private=True)
```

Regardless of whether I use mean or CLS pooling. 

@BenjaminBossan could you 1) verify that the PR diff looks solid at a glance and 2) let me know if a model with this config is supposed to train roughly as well as with "full" training?
```python
peft_config = PromptTuningConfig(
    task_type=TaskType.FEATURE_EXTRACTION,
    prompt_tuning_init=PromptTuningInit.RANDOM,
    num_virtual_tokens=1
)
```

- Tom Aarsen

### DIFF
--- a/sentence_transformers/models/Transformer.py
+++ b/sentence_transformers/models/Transformer.py
@@ -11,6 +11,7 @@ import huggingface_hub
 import torch
 from torch import nn
 from transformers import AutoConfig, AutoModel, AutoTokenizer, MT5Config, T5Config
+from transformers.utils import is_peft_available
 
 logger = logging.getLogger(__name__)
 
@@ -350,7 +351,23 @@ class Transformer(nn.Module):
         output_states = self.auto_model(**trans_features, **kwargs, return_dict=False)
         output_tokens = output_states[0]
 
-        features.update({"token_embeddings": output_tokens, "attention_mask": features["attention_mask"]})
+        # If the AutoModel is wrapped with a PeftModelForFeatureExtraction, then it may have added virtual tokens
+        # We need to extend the attention mask to include these virtual tokens, or the pooling will fail
+        if is_peft_available():
+            from peft import PeftModelForFeatureExtraction
+
+            if (
+                isinstance(self.auto_model, PeftModelForFeatureExtraction)
+                and self.auto_model.active_peft_config.is_prompt_learning
+            ):
+                batch_size = output_tokens.size(0)
+                attention_mask = features["attention_mask"]
+                prefix_attention_mask = torch.ones(
+                    batch_size, self.auto_model.active_peft_config.num_virtual_tokens, device=attention_mask.device
+                )
+                features["attention_mask"] = torch.cat((prefix_attention_mask, attention_mask), dim=1)
+
+        features["token_embeddings"] = output_tokens
 
         if self.auto_model.config.output_hidden_states:
             all_layer_idx = 2
@@ -358,7 +375,7 @@ class Transformer(nn.Module):
                 all_layer_idx = 1
 
             hidden_states = output_states[all_layer_idx]
-            features.update({"all_layer_embeddings": hidden_states})
+            features["all_layer_embeddings"] = hidden_states
 
         return features
 


BenjaminBossan:
Thanks a lot for quickly providing this solution to the issue. Regarding the question whether this should train well, I honestly don't have much experience with the prompt learning methods, so I can't say what I would expect. The fix itself looks correct to me, so I would proceed even if training does not work well at the moment.

Perhaps @VecherVhatuX or @mruniverse8 could give this branch a try and report back their results.

VecherVhatuX:
Oh my gosh, is this really a good idea? 

Using `torch.ones()` in the attention mask for virtual tokens feels, like, super wobbly, don’t you think? Shouldn’t we be using zeros instead? I mean, the model treating padding like it's the main character in a rom-com is just too much! 

Moreover, what’s the right way to handle this—ones or zeros? I’m just trying to make sure we’re not making a coding catastrophe!

tomaarsen:
`torch.ones()` is correct - my implementation is inspired by the PeftModelForFeatureExtraction: https://github.com/huggingface/peft/blob/e8259ff7bc72365ec730efb1f4005a95071d5dfd/src/peft/peft_model.py#L2616-L2619

It only differs in that I place the tensor on the correct device immediately rather than first initializing it and then moving it.

- Tom Aarsen

BenjaminBossan:
> Using `torch.ones()` in the attention mask for virtual tokens feels, like, super wobbly, don’t you think? Shouldn’t we be using zeros instead? I mean, the model treating padding like it's the main character in a rom-com is just too much!

Could you explain why using ones would be wobbly? The virtual tokens are valid tokens that could be attended to, right? They're not just padding.

VecherVhatuX:
Oh, I just had concerns about using ones for their attention masks, I do not have enough experience to make a good example. Just do not know what is better? To use ones or zeros for padding. If it's ok then,  no problem

tomaarsen:
I have confidence in the implementation I think, so I'll merge it. Thank you both for having a look at this!

- Tom Aarsen

PR #3007, opened 2024-10-20
Fix cache_dir issue with loading CLIPModel

### LINKED ISSUE #3005: 3.2.0 release breaks CLIPModel
After upgrading from version 3.1.1 to 3.2, loading `clip-ViT-L-14` gives the following error:

```python
>>> SentenceTransformer('clip-ViT-L-14')
  File "<stdin>", line 1, in <module>
  File "/path/to/python3.11/site-packages/sentence_transformers/SentenceTransformer.py", line 306, in __init__
    modules, self.module_kwargs = self._load_sbert_model(
                                  ^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/lib/python3.11/site-packages/sentence_transformers/SentenceTransformer.py", line 1722, in _load_sbert_model
    module = module_class(model_name_or_path, cache_dir=cache_folder, backend=self.backend, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: CLIPModel.__init__() got an unexpected keyword argument 'cache_dir'
```

This should be caused by the recent removal of the `try/except` block around the model loading code:

https://github.com/UKPLab/sentence-transformers/blob/29535eb44d270f5665e389aeb7e18cf04cb51bf0/sentence_transformers/SentenceTransformer.py#L1721-L1724

BoPeng:
Partially revert [this commit](https://github.com/UKPLab/sentence-transformers/commit/adbf0ba0daeefcffa8ff62d3e892a68f3ccb7a39#diff-b85567d4fdaffe34a3ccd8fe6cd1fcb15a986ebd34af373c71f1f5cf5efff021L1646) to fix #3005 .

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -1718,10 +1718,10 @@ print(similarities)
 
                 # Try to initialize the module with a lot of kwargs, but only if the module supports them
                 # Otherwise we fall back to the load method
-                # try:
-                module = module_class(model_name_or_path, cache_dir=cache_folder, backend=self.backend, **kwargs)
-                # except TypeError:
-                #     module = module_class.load(model_name_or_path)
+                try:
+                    module = module_class(model_name_or_path, cache_dir=cache_folder, backend=self.backend, **kwargs)
+                except TypeError:
+                    module = module_class.load(model_name_or_path)
             else:
                 # Normalize does not require any files to be loaded
                 if module_class == Normalize:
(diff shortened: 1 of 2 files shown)


BoPeng:
Already fixed in master by other means.

tomaarsen:
Hello!

As you already noticed, 3.2.0 works fine for e.g. https://huggingface.co/sentence-transformers/clip-ViT-L-14, but the issue does still persist if the CLIP model is saved in the root of the repository/directory. The removal of the try-except was a mistake from when I was debugging #2712, and so we should still re-introduce the try-except. This will fix the issue from https://github.com/UKPLab/sentence-transformers/issues/3005#issuecomment-2424969760, although we still won't have ONNX exporting for CLIP models (yet).

Thanks for opening your issue & PR. I'll include this in a quick patch very soon.

- Tom Aarsen

PR #3011, opened 2024-10-22
Update EmbeddingSimilarityEvaluator.py to save some examples and thei…

youssefkhalil320:
…r scores in a csv file

Add new option to save some examples from the evaluating dataset with their similarity score in a csv file to help the user the see the performance of the model.


tomaarsen:
Hello!

The evaluator's job is to produce metrics; example-level dumps are easy to add by subclassing or by capturing them via a callback. I'd rather not bake a bespoke save-examples format into the evaluator that we'd then have to maintain. Closing this.

- Tom Aarsen

PR #3020, opened 2024-10-24
add an ImportError to tell the user that `datasets` must be install to fit a model

### LINKED ISSUE #3019: Add an error to tell the user `datasets` is required to fit a model
The package `datasets` seems to be required in order to fit à model :

https://github.com/UKPLab/sentence-transformers/blob/f286d9f210824d6ea1563e789f49894b19c24f0e/sentence_transformers/fit_mixin.py#L264

We could raise an error to tell it to the user : 

```python
if not is_datasets_available():
    raise ImportError("Please install `datasets` to use this function: `pip install datasets`.")
```

h4c5:
Minor change to add an ImportError when `datasets` is not installed and the user try to fit a model ( Closes #3019 )

### DIFF
--- a/sentence_transformers/fit_mixin.py
+++ b/sentence_transformers/fit_mixin.py
@@ -231,6 +231,9 @@ class FitMixin:
             checkpoint_save_total_limit: Total number of checkpoints to
                 store
         """
+        if not is_datasets_available():
+            raise ImportError("Please install `datasets` to use this function: `pip install datasets`.")
+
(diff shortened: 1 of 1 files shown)


tomaarsen:
Well spotted! Apologies for the slight delay.
I hadn't noticed this myself as `model.fit` is the old way of finetuning a model, nowadays the Trainer is preferred: https://sbert.net/docs/sentence_transformer/training_overview.html

I do still want to merge this though - it'll help those who still want to use the old approach or still have scripts for `model.fit`.

- Tom Aarsen

PR #3028, opened 2024-10-30
Fix model name typo in example

programmer-ke:
The model in the example: `msmarco-distilroberta-base-v3` is missing from huggingface.

The existing ones are `msmarco-distilbert-base-v3` and `msmarco-roberta-base-v3`.

Changed the model name to the former one since it downloads faster.


tomaarsen:
You're very right, there's only these: 
image

What an odd mistake, hah. Thanks for pointing it out!

- Tom Aarsen

Issue #3016, opened 2024-10-24
Skip/customize Model Card Callback


tRosenflanz:
Currently Trainer will create a model card on `__init__` by getting a chunk of the dataset and trying to infer the types from this chunk. This can be fairly slow if the dataset contains images since the chunk is defaulted to size 1000 and in my case adds considerable overhead.

It would be nice to either disable this behavior or customize it to run on a smaller batch of data.

I tried to remove the ModelCardCallback with the `trainer.remove_callback` but since it is called in the `__init__` it obviously doesn't help.

For context: I am wrapping a custom model with SentenceTransformer and my dataset is nested dictionaries so I have to override some collation and "flattening" of the data to get it working by subclassing the trainer. Surprisingly, this only takes a few lines of code to get working with most losses. 

However, there is no nice way of disabling the ModeLCard callback without copying most of the SentenceTransformerTrainer.__init__ and removing a few lines from it which would immediately break on a version upgrade that touches this class. 

If it is an acceptable solution, I can create a PR that simply moves the callback creation and it's call to a separate function that can be overridden by inheriting from the trainer class and/or add an argument that explicitly disables the model card.

tomaarsen:
Hello!

I'm glad to hear that the customization of the Trainer has been relatively straightforward so far - that is ideal. 

Normally I wouldn't like to add something that removes the model card generation altogether, as it's quite valuable to eventual users, but I think your solution of moving the callback creation into a separate function is very reasonable. Feel free to include a docstring that explains that you can override the function to remove the model card callback in case it's detrimental to whatever customization the user wants to do.

- Tom Aarsen

tomaarsen:
Resolved via #3047 & included in v3.3.0.
Thanks for leading this!

- Tom Aarsen

PR #3047, opened 2024-11-08
Moved Model Card Callback init in Trainer to a separate function

tRosenflanz:
Model Card Callback is now added/initialized in a separate function that can be disabled/customized by subclassing the trainer 

https://github.com/UKPLab/sentence-transformers/issues/3016

### DIFF
--- a/sentence_transformers/trainer.py
+++ b/sentence_transformers/trainer.py
@@ -10,18 +10,8 @@ from typing import TYPE_CHECKING, Any, Callable
 import torch
 from packaging.version import parse as parse_version
 from torch import nn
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Thanks a bunch for setting this up. I made some minor changes: renamed the method & added a docstring which should look like this in the new documentation: 

image

I'll merge this once the tests go green!

- Tom Aarsen

tomaarsen:
Looks like I can't move the `default_args_dict` to after the `super().__init__()` due to some magic by `accelerate`. I reverted that change.

- Tom Aarsen

Issue #3056, opened 2024-11-14
FileNotFoundError when using SentenceTransformerTrainingArguments(load_best_model_at_end=True) and Peft


GTimothee:
When I try to train a peft model with the load_best_model_at_end=True option enabled, it crashes, raising a FileNotFoundError. 

Here is the example code as a gist: https://gist.github.com/GTimothee/cb3551ba6eb14b04f7a06d63ea4616f9 

Error I get: 'FileNotFoundError: [Errno 2] No such file or directory: 'training_trial/checkpoint-250/pytorch_model.bin''

GTimothee:
Here you go https://github.com/UKPLab/sentence-transformers/pull/3057

reissaavedra:
hi @tomaarsen 
I would like to work on this issue.
Please let me know if there are any specific requirements I should follow. Looking forward to contributing!

GTimothee:
opened a ticket in transformers to request the change https://github.com/huggingface/transformers/issues/34747

zuhorer:
hi! Is some one working on this? I can take this up in case no owns this?

PR #3057, opened 2024-11-14
Fix filenotfounderror #3056

GTimothee:
fixes issue #3056 temporarily

The idea is to remove the error and wait for a future change in the transformers library


tomaarsen:
Should be fixed via #3470, closing this.

- Tom Aarsen

Issue #3053, opened 2024-11-13
Issue loading private Transformer model in version 3.3.0


J-Curwell:
### Issue:
We have a fine-tuned version of ember-v1 (https://huggingface.co/llmrails/ember-v1) hosted as a private model in our Hugging Face org. We could successfully load this model in the previous release of sentence-transformers (3.2.1) but as of 3.3.0 this same code no longer works.

### Steps to reproduce:
- Install sentence-transformers 3.3.0
- Ensure you do **not** have the `HF_TOKEN` environment variable set. Our architecture reads this directly from a secret manager rather than setting as an environment variable
- Run:
```
from sentence_transformers.models.Transformer import Transformer

args = {
    "token": <AUTH TOKEN>,
    "trust_remote_code": False,
    "revision": None,
    "local_files_only": False
}

transformer = Transformer(
    model_name_or_path=<PRIVATE MODEL PATH>,
    cache_dir=None,
    backend="torch",
    max_seq_length=512,
    do_lower_case=True,
    model_args=args,
    tokenizer_args=args,
    config_args=args
)
```

This same code worked before. I believe I have isolated the issue to this recent PR #2980 . Specifically the call to the `find_adapter_config_file` on new line 104, which does not pass in the auth token.

Is the suggested approach to always set the `HF_TOKEN` environment variable?

HenningDinero:
Can confirm - have the exact same issue.
I have tried parsing the tokens directly and that doesn't work.

If I downgrade to 3.0.1 then it works like a charm

tomaarsen:
Feel free to stick with 3.2.1 for now, unless you'd like to try out some of the new v3.3.0 features like Training with Prompts, NanoBEIR Evaluation, PEFT, Int8 quantization with OpenVINO, etc.
In that case, you can download the bleeding edge version until I bring out v3.3.1:
```
pip install git+https://github.com/UKPLab/sentence-transformers.git
```

Thanks for reporting this!

- Tom Aarsen

PR #3058, opened 2024-11-15
[Fix] Resolve loading private Transformer model in version 3.3.0

pesuchin:
Resolves: https://github.com/UKPLab/sentence-transformers/issues/3053

# Details

- Add token and local_files_only and revision to the find_adapter_config_file arguments.

# Check

I have confirmed that I can load a private Transformer model with an adapter with the following code.

code:
```python
from sentence_transformers.models.Transformer import Transformer

args = {
    "token": <AUTH TOKEN>,
    "trust_remote_code": False,
    "revision": None,
    "local_files_only": False
}

transformer = Transformer(
    model_name_or_path=<PRIVATE MODEL PATH>,
    cache_dir=None,
    backend="torch",
    max_seq_length=512,
    do_lower_case=True,
    model_args=args,
    tokenizer_args=args,
    config_args=args
)
```

logs: 
```
model.safetensors: 100%|██████████████████████████████████████████████████████████████████████████████████████| 2.27G/2.27G [01:17<00:00, 24.1MB/s]
tokenizer_config.json: 100%|██████████████████████████████████████████████████████████████████████████████████| 1.17k/1.17k [00:00<00:00, 1.03MB/s]
sentencepiece.bpe.model: 100%|████████████████████████████████████████████████████████████████████████████████| 5.07M/5.07M [00:00<00:00, 9.80MB/s]
tokenizer.json: 100%|█████████████████████████████████████████████████████████████████████████████████████████| 17.1M/17.1M [00:00<00:00, 24.9MB/s]
special_tokens_map.json: 100%|████████████████████████████████████████████████████████████████████████████████████| 964/964 [00:00<00:00, 1.87MB/s]
```

### DIFF
--- a/sentence_transformers/models/Transformer.py
+++ b/sentence_transformers/models/Transformer.py
@@ -101,7 +101,14 @@ class Transformer(nn.Module):
 
     def _load_config(self, model_name_or_path: str, cache_dir: str | None, backend: str, config_args: dict[str, Any]):
         """Loads the configuration of a model"""
-        if find_adapter_config_file(model_name_or_path) is not None:
+        if (
+            find_adapter_config_file(
+                model_name_or_path,
+                token=config_args.get("token"),
+                local_files_only=config_args.get("local_files_only"),
+            )
+            is not None
+        ):
             if not is_peft_available():
                 raise Exception(
                     "Loading a PEFT model requires installing the `peft` package. You can install it via `pip install peft`."
--- a/sentence_transformers/models/Transformer.py
+++ b/sentence_transformers/models/Transformer.py
@@ -105,6 +105,7 @@ class Transformer(nn.Module):
             find_adapter_config_file(
                 model_name_or_path,
                 token=config_args.get("token"),
+                revision=config_args.get("revision"),
                 local_files_only=config_args.get("local_files_only"),
             )
             is not None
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

I'm not going to add a test for this for now - it would just be a bit too messy to have a private model that everyone can somehow access, but this does seem to resolve the problem. 
Also, I could have sworn that `dict.get()` gave a `KeyError` by default if it failed, but I suppose not, haha.

I updated the default for `local_files_only` to False, as that's the default in `find_adapter_config_file` as well - the others can stay `None`.

Thanks for tackling this so quickly @pesuchin, I'd like to quickly merge this & bring it out in a patch release.

cc @J-Curwell, @HenningDinero thanks for reporting this! 

- Tom Aarsen

HenningDinero:
> Hello!
> 
> I'm not going to add a test for this for now - it would just be a bit too messy to have a private model that everyone can somehow access, but this does seem to resolve the problem. Also, I could have sworn that `dict.get()` gave a `KeyError` by default if it failed, but I suppose not, haha.
> 
> I updated the default for `local_files_only` to False, as that's the default in `find_adapter_config_file` as well - the others can stay `None`.
> 
> Thanks for tackling this so quickly @pesuchin, I'd like to quickly merge this & bring it out in a patch release.
> 
> cc @J-Curwell, @HenningDinero thanks for reporting this!
> 
> * Tom Aarsen

It is indeed difficult to make a test to a private-public repo 😅 
But yeah, `dict.get` does not raise an error (use `dict[]` instead if you want ;-))

J-Curwell:
Thank you for resolving so quickly @pesuchin @tomaarsen! 🥇

PR #3066, opened 2024-11-18
[`training`] Pass `steps`/`epoch`/`output_path` to Evaluator during training

### LINKED ISSUE #3062: RerankingEvaluator does not write the csv file when the backward compatible `fit()` method is used. Possible fix suggested.
* `sentence-transformers` version tested: 3.3 (branch: 3.3-release)
* Context: we are migrating a production pipeline that fine-tunes a model from `sentence-transformers` 2.x to 3.x 
* In 2.x we used the now deprecated method `model.fit()` as step toward 3.x we tested the `fit()` method provided for backward compatibility. However, a `RerankingEvaluator` we pass during training does not write the `csv` file with the metrics, as it did before. This file is very useful to us.

* Below my notes on the issue.

`RerankingEvaluator` needs the `output_path` to be set in order to write the csv file:

https://github.com/UKPLab/sentence-transformers/blob/ea49e01d53f7919b10b6a458553a2546bc44521f/sentence_transformers/evaluation/RerankingEvaluator.py#L138

However, if my understanding is correct, in the `fit_mixin.py` the class `EvaluatorCallback` calls the `evaluator` without passing the `output_path` : https://github.com/UKPLab/sentence-transformers/blob/ea49e01d53f7919b10b6a458553a2546bc44521f/sentence_transformers/fit_mixin.py#L117

which causes that the file is not written to disk since `output_path` is `None`. Note that the flag `write_csv: bool` is by default set to `True`, which is the other condition for the `RerankingEvaluator` to dump the file: https://github.com/UKPLab/sentence-transformers/blob/ea49e01d53f7919b10b6a458553a2546bc44521f/sentence_transformers/evaluation/RerankingEvaluator.py#L138 .

I think a possible fix would be to modify the [call to the evaluator in `fit_mixin.py`](https://github.com/UKPLab/sentence-transformers/blob/ea49e01d53f7919b10b6a458553a2546bc44521f/sentence_transformers/fit_mixin.py#L117 ) from

```evaluator_metrics = self.evaluator(model, epoch=state.epoch)```

to 

```evaluator_metrics = self.evaluator(model, epoch=state.epoch, output_path=args.output_dir)```

I do understand this method is provided only to support the transition and it will be removed, but I think it would be nice to have it working as expected in the meantime :)  .

### PR DESCRIPTION
**Pull request by tomaarsen**
Resolves #3062

Hello!

## Pull Request overview
* Pass `steps`/`epoch`/`output_path` to Evaluator during training
  * Both for the deprecated `model.fit()` and the `SentenceTransformerTrainer`
* Fix EmbeddingSimilarityEvaluator and NanoBEIREvaluator who failed when writing to disk

## Details
The `model.fit` nor `SentenceTransformerTrainer` passed all available parameters down to the evaluator(s). In particular, the `output_path` wasn't sent, which meant that no csv files could be written for the evaluators. Beyond that, 2 evaluators failed when writing to disk. Some better tests for the evaluators would be appreciated, as this could've been caught.

- Tom Aarsen

### DIFF
--- a/sentence_transformers/evaluation/EmbeddingSimilarityEvaluator.py
+++ b/sentence_transformers/evaluation/EmbeddingSimilarityEvaluator.py
@@ -225,9 +225,9 @@ class EmbeddingSimilarityEvaluator(SentenceEvaluator):
                         steps,
                     ]
                     + [
-                        metrics[f"{fn_name}_{m}"]
+                        metrics[f"{metric}_{fn_name}"]
                         for fn_name in self.similarity_fn_names
-                        for m in ["pearson", "spearman"]
+                        for metric in ["pearson", "spearman"]
                     ]
                 )
 
--- a/sentence_transformers/evaluation/NanoBEIREvaluator.py
+++ b/sentence_transformers/evaluation/NanoBEIREvaluator.py
@@ -326,20 +326,20 @@ class NanoBEIREvaluator(SentenceEvaluator):
             output_data = [epoch, steps]
             for name in self.score_function_names:
                 for k in self.accuracy_at_k:
-                    output_data.append(per_dataset_results[name]["accuracy@k"][k])
+                    output_data.append(agg_results[f"{name}_accuracy@{k}"])
 
                 for k in self.precision_recall_at_k:
-                    output_data.append(per_dataset_results[name]["precision@k"][k])
-                    output_data.append(per_dataset_results[name]["recall@k"][k])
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by bluebalam** [comment]
Thank you very much @tomaarsen for the quick response 👍🏾

**Comment by tomaarsen** [comment]
Gladly! You can use this already with:
```
```
And otherwise this will be included in the next release.

- Tom Aarsen

PR #3068, opened 2024-11-18
support cached losses in combination with matryoshka loss

Marcel256:
Proof of concept for supporting cached losses in combination with the matryoshka loss

### DIFF
--- a/sentence_transformers/losses/MatryoshkaLoss.py
+++ b/sentence_transformers/losses/MatryoshkaLoss.py
@@ -27,6 +27,14 @@ def shrink(tensor: Tensor, dim: int) -> Tensor:
 
 
 class ForwardDecorator:
+    """
+    This decorator is used to cache the output of the Sentence Transformer's forward pass,
+    so that it can be shrank and reused for multiple loss calculations. This prevents the
+    model from recalculating the embeddings for each desired Matryoshka dimensionality.
+
+    This decorator is applied to `SentenceTransformer.forward`.
+    """
+
     def __init__(self, fn) -> None:
         self.fn = fn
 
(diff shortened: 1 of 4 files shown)


Marcel256:
I cleaned up the Matryoshka loss a bit more. Moreover, I also see potential to further reduce code duplication in the cached loss classes, which could be done in a separate PR. Are there any other points which I should adress in this PR?

sentence_transformers/losses/MatryoshkaLoss.py
```diff
@@ -165,12 +158,13 @@ def __init__(
         dims_weights = zip(matryoshka_dims, matryoshka_weights)
         self.matryoshka_dims, self.matryoshka_weights = zip(*sorted(dims_weights, key=lambda x: x[0], reverse=True))
         self.n_dims_per_step = n_dims_per_step
+        self.cached_losses = (CachedMultipleNegativesRankingLoss, CachedGISTEmbedLoss, CachedMultipleNegativesSymmetricRankingLoss)
```
tomaarsen, line 161: Nice - this has a niche but useful bonus effect that people interested in creating their own Cached loss can extend this after initializing the loss.

E.g. 
```python
loss = CachedCustomLoss(model, ...)
loss = MatryoshkaLoss(model, loss, matryoshka_dims=[768, 512, 256, 128, 64, 32])
loss.loss.calculate_loss = CachedLossDecorator(...)
loss.cached_losses = CachedCustomLoss
```

tomaarsen:
I think this PR is looking quite solid. All I think that remains is to run
```
pip install pre-commit
pre-commit install # Set a pre-commit hook that runs formatting/linting code before every commit (in this repository only)
pre-commit run --all # Run the aforementioned hook separate from doing a commit
```
It should satisfy the quality code check.

- Tom Aarsen

Marcel256:
Thank you! The formatting should be fine now

PR #3076, opened 2024-11-20
converting np float result to float in binary classification evaluator 

### LINKED ISSUE #3075: BinaryClassificationEvaluator returns np.float32() and np.float64()
I came across the problem where

```python
output_scores[similarity_fn_name] = {
  "accuracy": acc,
  "accuracy_threshold": acc_threshold,
  "f1": f1,
  "f1_threshold": f1_threshold,
  "precision": precision,
  "recall": recall,
  "ap": ap,
}
```

is returning outputs in `np.float32()` and `np.float64()` which cause problems during model saving, as the `json/encoder.py` sees those types as `not JSON serializable`.

By changing the code snippet to

```python
output_scores[similarity_fn_name] = {
  "accuracy": acc.item(),
  "accuracy_threshold": acc_threshold.item(),
  "f1": f1.item(),
  "f1_threshold": f1_threshold.item(),
  "precision": precision,
  "recall": recall.item(),
  "ap": ap.item(),
}
```

the elements get copied to a standard Python scalar each (note: `precision` already is) and then behave like expected during saving.

JINO-ROHIT:
Fixes #3075 

A simple fix to change the result np.floats to float.

I had two options -
1. using .item()
2. using type cast float()

float() seemed slightly faster so i went with it.

Who can review
@tomaarsen

### DIFF
--- a/sentence_transformers/evaluation/BinaryClassificationEvaluator.py
+++ b/sentence_transformers/evaluation/BinaryClassificationEvaluator.py
@@ -288,14 +288,14 @@ class BinaryClassificationEvaluator(SentenceEvaluator):
             logger.info(f"Matthews Correlation with {name}: {mcc * 100:.2f}\n")
 
             output_scores[similarity_fn_name] = {
-                "accuracy": acc,
-                "accuracy_threshold": acc_threshold,
-                "f1": f1,
-                "f1_threshold": f1_threshold,
-                "precision": precision,
-                "recall": recall,
-                "ap": ap,
-                "mcc": mcc,
+                "accuracy": float(acc),
+                "accuracy_threshold": float(acc_threshold),
+                "f1": float(f1),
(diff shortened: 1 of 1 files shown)


tomaarsen (approved):
I like `float` - it keeps working even if sklearn or something updates their output type from np.float32/np.float64 to regular float, unlike `.item()`.

I'll try this out locally tomorrow morning I suspect.

Thanks!

tomaarsen:
Apologies for the delay!
I tested this locally and it works like a charm. Thanks for reporting @chr-werner and thanks for fixing @JINO-ROHIT!

- Tom Aarsen

JINO-ROHIT:
Hiya @tomaarsen 
Just remembered that this PR Id raised for BinaryClassificationEvaluator.py where I do a float on the results arent needed anymore with this fix . PR #3076 

Should we revert them back since it looks kinda ugly?

Issue #3061, opened 2024-11-15
Error when trying to load Adapter model of revision in repository on HF


pesuchin:
## Summary

When trying to load an Adapter Model by specifying a revision in an private and public repository on HF, the following error occurs:

```bash
 File "/Users/pesuchin/project/sentence-transformers/examples/debug2.py", line 7, in <module>
    model = SentenceTransformer("pesuchin/revision-pesu-test", 
  File "/Users/pesuchin/project/sentence-transformers/sentence_transformers/SentenceTransformer.py", line 308, in __init__
    modules, self.module_kwargs = self._load_sbert_model(
  File "/Users/pesuchin/project/sentence-transformers/sentence_transformers/SentenceTransformer.py", line 1728, in _load_sbert_model
    module = module_class(model_name_or_path, cache_dir=cache_folder, backend=self.backend, **kwargs)
  File "/Users/pesuchin/project/sentence-transformers/sentence_transformers/models/Transformer.py", line 78, in __init__
    self._load_model(model_name_or_path, config, cache_dir, backend, **model_args)
  File "/Users/pesuchin/project/sentence-transformers/sentence_transformers/models/Transformer.py", line 138, in _load_model
    self.auto_model = AutoModel.from_pretrained(
  File "/Users/pesuchin/project/sentence-transformers/.venv/lib/python3.9/site-packages/transformers/models/auto/auto_factory.py", line 525, in from_pretrained
    config, kwargs = AutoConfig.from_pretrained(
  File "/Users/pesuchin/project/sentence-transformers/.venv/lib/python3.9/site-packages/transformers/models/auto/configuration_auto.py", line 976, in from_pretrained
    config_dict, unused_kwargs = PretrainedConfig.get_config_dict(pretrained_model_name_or_path, **kwargs)
  File "/Users/pesuchin/project/sentence-transformers/.venv/lib/python3.9/site-packages/transformers/configuration_utils.py", line 633, in get_config_dict
    config_dict, kwargs = cls._get_config_dict(pretrained_model_name_or_path, **kwargs)
  File "/Users/pesuchin/project/sentence-transformers/.venv/lib/python3.9/site-packages/transformers/configuration_utils.py", line 690, in _get_config_dict
    resolved_config_file = cached_file(
  File "/Users/pesuchin/project/sentence-transformers/.venv/lib/python3.9/site-packages/transformers/utils/hub.py", line 432, in cached_file
    raise EnvironmentError(
[...]

PR #3079, opened 2024-11-22
[fix] revision of the adapter model can now be specified.

pesuchin:
Resolved: https://github.com/UKPLab/sentence-transformers/issues/3061

## Solution

I modified the `_load_peft_model` method so that when the config is `PeftConfig`, it handles loading both the base model and the adapter model. Additionally, when loading the base model, I temporarily save and remove the `revision` from `model_args` before loading the base model.

This ensures that the base model is loaded using the default revision instead of the adapter model's revision, resolving the error.

## Verification

The following three codes were tested and worked correctly.

- verification1:
```python
from sentence_transformers.models.Transformer import Transformer

args = {
    "token": <your token>,
    "trust_remote_code": False,
    "revision": <your repository revision>,
    "local_files_only": False
}

transformer = Transformer(
    model_name_or_path=<your model_path>,
    cache_dir=None,
    backend="torch",
    max_seq_length=512,
    do_lower_case=True,
    model_args=args,
    tokenizer_args=args,
    config_args=args
)
```

- verification2:
```python
model = SentenceTransformer("sentence-transformers-testing/stsb-bert-tiny-safetensors")
model.encode("Hello, World!")
```

- verification3:
```python

kwargs = {
    "revision": <your revision>,
}
model = SentenceTransformer(<your model repository path>,
                            use_auth_token=<your token>,
                            model_kwargs=kwargs,
                            config_kwargs=kwargs,
                            tokenizer_kwargs=kwargs)
model.encode("Hello, World!")
```

### DIFF
--- a/sentence_transformers/models/Transformer.py
+++ b/sentence_transformers/models/Transformer.py
@@ -132,13 +132,19 @@ class Transformer(nn.Module):
         if backend == "torch":
             if isinstance(config, T5Config):
                 self._load_t5_model(model_name_or_path, config, cache_dir, **model_args)
+                return
             elif isinstance(config, MT5Config):
                 self._load_mt5_model(model_name_or_path, config, cache_dir, **model_args)
-            else:
-                self.auto_model = AutoModel.from_pretrained(
-                    model_name_or_path, config=config, cache_dir=cache_dir, **model_args
-                )
-            self._load_peft_model(model_name_or_path, config, cache_dir, **model_args)
+                return
+            elif is_peft_available():
+                from peft import PeftConfig
+
+                if isinstance(config, PeftConfig):
+                    self._load_peft_model(model_name_or_path, config, cache_dir, **model_args)
+                    return
+            self.auto_model = AutoModel.from_pretrained(
+                model_name_or_path, config=config, cache_dir=cache_dir, **model_args
+            )
         elif backend == "onnx":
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

I see a minor issue here that I'd like to see resolved:
* The T5 and MT5 models can also have PEFT Adapters, but no longer with this PR. 

I had a crack at this, and I think the easiest solution is actually to have the `_load_config` return whether the model is a PEFT Model or not. That simplifies everything else a lot. Then we can load the base model like before, but *beforehand* we update the model_args and then *afterwards* we load the PEFT adapter on top.

I also added a simple test with the https://huggingface.co/sentence-transformers-testing/stsb-bert-tiny-lora model.

I'm curious to hear what you think!

- Tom Aarsen

tomaarsen:
Apologies for expanding the PR a bit more via 7ee9fa84d9ddc5b5a3bf4a0784f86255be6a338a. In https://github.com/UKPLab/sentence-transformers/pull/3079/files you can select only the first 2 commits of this PR by shift clicking the first 2
image

That's the easiest way to understand my changes in ec366b5f0c8da26b752662e2e17f0a95e0ac36d1

- Tom Aarsen

pesuchin:
@tomaarsen 
Hello!
Thank you for your suggested revisions.
I think it’s a great idea.
The implementation has become much simpler.
Thank you for adding a simple test as well.

PR #3104, opened 2024-11-29
Fix cross encoder device issue

### LINKED ISSUE #3078: `CrossEncoder` is not pushed to cuda until predict is called, even if cuda is specified as device.
Hi, this is more like a question rather than a bug or issue.

When I specify the target device during initialization of any CrossEncoder, the model is not pushed to that device until the `predict` or the `fit` method is called, until then the model is kept in cpu.

```python

model2 = CrossEncoder("mixedbread-ai/mxbai-rerank-large-v1", device="cuda:0")
print(model2.model.device)
# cpu
```

I mean I expect the model to be pushed to the specified device during initialization and until I am calling predict it is taking up my system Ram.
Is there any high level reason(s) why this is the case?

susnato:
Fixes #3078

- fixed the previous issue and now the model is pushed to the specified device when the object is created.
- added new method `to` which directly pushes the model to the specified device, removing the need to use model.model.to
- added 2 tests to make sure it works as expected

### DIFF
--- a/sentence_transformers/cross_encoder/CrossEncoder.py
+++ b/sentence_transformers/cross_encoder/CrossEncoder.py
@@ -606,5 +606,16 @@ class CrossEncoder(PushToHubMixin):
         return self.model.to(device)
 
     @property
+    def _target_device(self) -> torch.device:
+        logger.warning(
+            "`CrossEncoder._target_device` has been removed, please use `CrossEncoder.device` instead.",
+        )
+        return self.device
+
+    @_target_device.setter
+    def _target_device(self, device: int | str | torch.device | None = None) -> None:
+        self.to(device)
+
+    @property
     def device(self) -> torch.device:
         return self.model.device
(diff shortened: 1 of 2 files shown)


susnato:
cc: @tomaarsen

tomaarsen:
Hello!

Well done here @susnato! I made some tiny adjustments, primarily surrounding the removal of `_target_device`. Despite it being a "private attribute", I imagine that some users updated `model._target_device` to move the model from devices, so I've added a property + setter to ensure that it's still possible (although it's not recommended).
We have something similar for the `SentenceTransformer` class.

What do you think? Oh, I'm also going to try the Copilot reviewer on this.

- Tom Aarsen

tomaarsen (approved):

susnato:
hi @tomaarsen , seems good to me to have the _target_device. 

>  I'm also going to try the Copilot reviewer on this.

lol, I didn't knew this even existed!

tomaarsen:
> lol, I didn't knew this even existed!

It's very new! I only got access a few days ago, and this is the second time that I've used it. Seems like it didn't help much this time, haha.

I'll merge this now as I think we're good to go. Thanks for the help!

- Tom Aarsen

PR #3110, opened 2024-12-03
[`fix`] Fix breaking change in PyLate when loading modules

### LINKED ISSUE #3108: Breaking change for module loading in PyLate
Hello,

Since the version 3.1, the module are loaded with [get_class_from_dynamic_module](https://github.com/UKPLab/sentence-transformers/commit/6257cb00ef87ddc8f445f0e436377ea4c48879b2#diff-b85567d4fdaffe34a3ccd8fe6cd1fcb15a986ebd34af373c71f1f5cf5efff021R1466) when using ```trust_remote_code=True```
However, this function [breaks if the type of the modules contains more than one dot](https://github.com/huggingface/transformers/blob/31299670cda29f25fbc655f6f166e7b8cc21c89f/src/transformers/dynamic_module_utils.py#L536)
The Dense module from PyLate is saved with the type pylate.models.Dense.Dense, [due to this line](https://github.com/UKPLab/sentence-transformers/blob/ba8cb2ef1d7a5ad6b9169eddee469e900456532c/sentence_transformers/SentenceTransformer.py#L1189).

Before 3.1, the type of the module was pylate.models.Dense, but this still break when parsing on the dot. As the PyLate module still loads correctly with the ```import_from_string``` function (that was used to load the modules prior to 3.1), one of the solution is to broaden a bit the exception to prevent crash if the file is not there locally AND also if the parsing fail (```ValueError: too many values to unpack (expected 2)```)
The other possibility is to add ```class_ref.startswith("pylate.")``` to [this test](https://github.com/UKPLab/sentence-transformers/blob/ba8cb2ef1d7a5ad6b9169eddee469e900456532c/sentence_transformers/SentenceTransformer.py#L1550). It is a bit hardcoded but prevents not catching an exception earlier if there is some other ValueError

Let me know what solution you prefer and I can create a PR for this

tomaarsen:
Resolves #3108

Hello!

## Pull Request overview
* Prevent crash when calling `get_class_from_dynamic_module` with incorrect `class_ref`.

## Details
If the `class_ref` in `_load_module_class_from_ref` does not start with `sentence_transformers.`, and it's not a remote module, but e.g. `pylate`'s `Dense`, then `get_class_from_dynamic_module` will fail due to `a, b = c.split(".")` where `c` has more than 1 dot.

I think the cleanest solution is to ignore the `ValueError` - as otherwise we risk missing some edge cases if we e.g. expand on the `if trust_remote_code ...` before the `get_class_from_dynamic_module` call.

@NohTow I believe you already verified that this solution works, right?

- Tom Aarsen

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -1559,8 +1559,8 @@ print(similarities)
                     revision=revision,
                     code_revision=code_revision,
                 )
-            except OSError:
-                # Ignore the error if the file does not exist, and fall back to the default import
+            except (OSError, ValueError):
+                # Ignore the error if 1) the file does not exist, or 2) the class_ref is not correctly formatted/found
                 pass
 
         return import_from_string(class_ref)


tomaarsen:
> Small nitpick: the PR does not "prevent calling get_class_from_dynamic_module", it actually calls it but fallback to the other loading method when crashing.

Yes, let me update my phrasing. I meant "prevent calling ... from resulting in crash", but it wasn't clear:
> Prevent calling `get_class_from_dynamic_module` with incorrect `class_ref` resulting in a crash.

- Tom Aarsen

PR #3114, opened 2024-12-03
fix backward pass for cached losses

Marcel256:
Here is a draft for a fix of the backward pass in the cached losses, while still maintaining the compatibility with the matryoshka loss. The problem is a bit more difficult than I originally thought because we need to detach the tensor before doing the minibatch loss computation.

If anyone has suggestions for improvements, let me know 😀

### DIFF
--- a/sentence_transformers/losses/CachedGISTEmbedLoss.py
+++ b/sentence_transformers/losses/CachedGISTEmbedLoss.py
@@ -229,7 +229,7 @@ class CachedGISTEmbedLoss(nn.Module):
 
     def calculate_loss_and_cache_gradients(self, reps: list[list[Tensor]], reps_guided: list[list[Tensor]]) -> Tensor:
         """Generalized function to calculate the cross-entropy loss and cache the gradients wrt. the embeddings."""
-        loss = self.calculate_loss(reps, reps_guided)
+        loss = self.calculate_loss(reps, reps_guided, with_backward=True)
         loss = loss.detach().requires_grad_()
 
         self.cache = [[r.grad for r in rs] for rs in reps]
--- a/sentence_transformers/losses/CachedGISTEmbedLoss.py
+++ b/sentence_transformers/losses/CachedGISTEmbedLoss.py
@@ -230,14 +230,15 @@ class CachedGISTEmbedLoss(nn.Module):
     def calculate_loss_and_cache_gradients(self, reps: list[list[Tensor]], reps_guided: list[list[Tensor]]) -> Tensor:
         """Generalized function to calculate the cross-entropy loss and cache the gradients wrt. the embeddings."""
         loss = self.calculate_loss(reps, reps_guided)
-        loss.backward()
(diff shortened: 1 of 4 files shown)


sentence_transformers/losses/CachedGISTEmbedLoss.py
```diff
@@ -230,14 +230,15 @@ def embed_minibatch_iter(
     def calculate_loss_and_cache_gradients(self, reps: list[list[Tensor]], reps_guided: list[list[Tensor]]) -> Tensor:
         """Generalized function to calculate the cross-entropy loss and cache the gradients wrt. the embeddings."""
         loss = self.calculate_loss(reps, reps_guided)
```
tomaarsen, line 232:
```suggestion
        loss = self.calculate_loss(reps, reps_guided, with_backward=True)
```
I believe this is missing for CGIST.

tomaarsen:
My initial tests with CMNRL and CMNRL + Matryoshka are promising, and I can indeed also reproduce the high memory usage on `master`.

Matryoshka on top of CMNRL only seems to add about 20%-25% training time.

CMNRL does seem a decent bit slower than pure MNRL (more than the 10-20% that I thought it was), but based on your changes in this PR and the previous, that shouldn't be related to you at all.

- Tom Aarsen

tomaarsen:
Nice! I assume that 0292b9b48db27cce603517b55ec2e1fa36629bab means that we perform evaluations more efficiently?

Marcel256:
Yes, with [0292b9b](https://github.com/UKPLab/sentence-transformers/pull/3114/commits/0292b9b48db27cce603517b55ec2e1fa36629bab) the loss can now also be called with torch.no_grad(). 
I trained several models with this fix and the results are looking good. No more memory issues 😄

tomaarsen:
Thanks a bunch for having another look at this @Marcel256 🤗 

- Tom Aarsen

kuanhsieh:
Thanks a lot @tomaarsen and @Marcel256 for this fix, this would be extremely helpful for my work on fine-tuning embedding models using the `CachedMultipleNegativesRankingLoss` together with the `MatryoshkaLoss`. 

Is there an expected timeline for when there will be a new release of `sentence-transformers` which will contain this fix? Thanks a lot!

tomaarsen:
Hello @kuanhsieh!

Normally my releases are just whenever there's been a sufficiently big change that warrants an update, which has historically been about once a month. However, due to the holidays/new years, and also the ModernBERT release, things have been a bit slower.

I can however try to prioritize a smaller release in the near future. Perhaps some time next week?

- Tom Aarsen

kuanhsieh:
Thank you very much for update and for all the work on ModernBERT! I will use the Github version in the meantime.

tomaarsen:
Hello @kuanhsieh, I've made the v3.4.0 release: https://github.com/UKPLab/sentence-transformers/releases/tag/v3.4.0

- Tom Aarsen

PR #3122, opened 2024-12-08
raises error when dataset is an empty list in NanoBEIREvaluator

### LINKED ISSUE #3103: NanoBeirEvaluator takes in empty dataset
I noticed that NanoBeir takes in an empty dataset and returns no error during instantiation. But when the model gets passed, the error kinda seems confusing.
So -
1. is it fine to allow the evaluator to take in an empty dataset.
2. Should we change the error message saying "invalid datset or something"

Who can help?
@tomaarsen 

Snippet reproduction
```

datasets = []

#model = SentenceTransformer("sentence-transformers-testing/stsb-bert-tiny-safetensors")

evaluator = NanoBEIREvaluator( # this bit here returns no error
    dataset_names=datasets,
)

results = evaluator(model) #this raised an error
```

#####################################################################

Error log
```
{
	"name": "KeyError",
	"message": "'cosine_ndcg@10'",
	"stack": "---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[4], line 1
----> 1 evaluator(model)

sentence_transformers\\evaluation\\NanoBEIREvaluator.py:351, in NanoBEIREvaluator.__call__(self, model, output_path, epoch, steps, *args, **kwargs)
    348 if not self.primary_metric:
    349     if self.main_score_function is None:
    350         score_function = max(
--> 351             [(name, agg_results[f\"{name}_ndcg@{max(self.ndcg_at_k)}\"]) for name in self.score_function_names],
    352             key=lambda x: x[1],
    353         )[0]
    354         self.primary_metric = f\"{score_function}_ndcg@{max(self.ndcg_at_k)}\"
    355     else:

sentence-transformers\\sentence_transformers\\evaluation\\NanoBEIREvaluator.py:351, in <listcomp>(.0)
    348 if not self.primary_metric:
    349     if self.main_score_function is None:
    350         score_function = max(
--> 351             [(name, agg_results[f\"{name}_ndcg@{max(self.ndcg_at_k)}\"]) for name in self.score_function_names],
    352             key=lambda x: x[1],
    353         )[0]
    354         self.primary_metric = f\"{score_function}_ndcg@{max(self.ndcg_at_k)}\"
    355     else:

KeyError: 'cosine_ndcg@10'"
}
```

JINO-ROHIT:
Raises an error when dataset names are a empty list in NanoBEIREvaluator.

Fixes #3103 

Who can review?
@tomaarsen

### DIFF
--- a/sentence_transformers/evaluation/NanoBEIREvaluator.py
+++ b/sentence_transformers/evaluation/NanoBEIREvaluator.py
@@ -420,6 +420,8 @@ class NanoBEIREvaluator(SentenceEvaluator):
         )
 
     def _validate_dataset_names(self):
+        if self.dataset_names == []:
+            raise ValueError("dataset_names cannot be empty. Use None to evaluate on all datasets.")
         if missing_datasets := [
             dataset_name for dataset_name in self.dataset_names if dataset_name.lower() not in dataset_name_to_id
         ]:


sentence_transformers/evaluation/NanoBEIREvaluator.py
```diff
@@ -420,6 +420,8 @@ def _load_dataset(self, dataset_name: DatasetNameType, **ir_evaluator_kwargs) ->
         )
 
     def _validate_dataset_names(self):
+        if self.dataset_names == []:
```
tomaarsen, line 423:
```suggestion
        if len(self.dataset_names) == 0:
```
This is a personal preference, but I prefer checking lengths like this :)

JINO-ROHIT: done!

JINO-ROHIT: on second thought, theres also this case when datasets = [''], an empty string, 

currently it recognizes as an invalid dataset
```
ValueError: Dataset(s) [''] not found in the NanoBEIR collection.Valid dataset names are: ['climatefever', 'dbpedia', 'fever', 'fiqa2018', 'hotpotqa', 'msmarco', 'nfcorpus', 'nq', 'quoraretrieval', 'scidocs', 'arguana', 'scifact', 'touche2020']
```

would it be nicer to consider it as an empty dataset and use this check instead?

tomaarsen: I don't think we have to worry about this case, I don't think many people would use `[""]` and if they do, the error is already pretty descriptive.

JINO-ROHIT: cool then

tomaarsen:
Thanks again! 

- Tom Aarsen

PR #3126, opened 2024-12-10
raises ValueError when num_label !=1 when using Crossencoder.rank()

### LINKED ISSUE #3124: CrossEncoder .rank condition error in CrossEncoder.py
I get the following error when I use .rank method:

```
File /usr/local/lib/python3.12/dist-packages/sentence_transformers/cross_encoder/CrossEncoder.py:551, in CrossEncoder.rank(self, query, documents, top_k, return_documents, batch_size, show_progress_bar, num_workers, activation_fct, apply_softmax, convert_to_numpy, convert_to_tensor)
    548     if return_documents:
    549         results[-1].update({"text": documents[i]})
--> 551 results = sorted(results, key=lambda x: x["score"], reverse=True)
    552 return results[:top_k]

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
I use sentence_transformers v3.3.0.

A snippet:
```python

cross_encoder = sentence_transformers.CrossEncoder("amberoad/bert-multilingual-passage-reranking-msmarco", device='cpu', max_length=256)
cross_encoder.rank(query, docs)
```

JINO-ROHIT:
Raises ValueError when num_label !=1 when using Crossencoder.rank()

Fixes #3124 

Who can review?
@tomaarsen

### DIFF
--- a/sentence_transformers/cross_encoder/CrossEncoder.py
+++ b/sentence_transformers/cross_encoder/CrossEncoder.py
@@ -527,6 +527,8 @@ class CrossEncoder(PushToHubMixin):
                 'score': -5.082967,
                 'text': "The 'Harry Potter' series, which consists of seven fantasy novels written by British author J.K. Rowling, is among the most popular and critically acclaimed books of the modern era."}]
         """
+        if self.config.num_labels != 1:
+            raise ValueError("CrossEncoder.rank() only works for models with num_labels=1.")
         query_doc_pairs = [[query, doc] for doc in documents]
         scores = self.predict(
             sentences=query_doc_pairs,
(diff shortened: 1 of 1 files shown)


sentence_transformers/cross_encoder/CrossEncoder.py
```diff
@@ -527,6 +527,8 @@ def rank(
                 'score': -5.082967,
                 'text': "The 'Harry Potter' series, which consists of seven fantasy novels written by British author J.K. Rowling, is among the most popular and critically acclaimed books of the modern era."}]
         """
+        if self.config.num_labels != 1:
+            raise ValueError("CrossEncoder.rank() only works for models with num_labels=1.")
```
tomaarsen, line 531: Nice!
What do you think, should we maybe also refer people to `CrossEncoder.predict()`?
Something like:
```suggestion
            raise ValueError(
            	"CrossEncoder.rank() only works for models with num_labels=1. "
            	"Consider using CrossEncoder.predict() with input pairs instead."
            )
```
Otherwise people might just stop using Sentence Transformers altogether when `CrossEncoder.rank` fails.

JINO-ROHIT: true

tomaarsen:
I like it! Well done - I appreciate it.

- Tom Aarsen

PR #3127, opened 2024-12-10
[DRAFT] tests for nanobeir evaluator

JINO-ROHIT:
This is a draft PR for testing the NanoBeir Evaluator. 

To be completely honest, I have very limited experience in writing pytests so I came up with few (broken) cases. Id like your help @tomaarsen here in how can i increase coverage for the nanobeir module, what are the different elements i can test for if you dont mind? 😅


tomaarsen (approved):
I think this is looking quite solid! The only issue you had was that the `match` in `pytest.raises` is taken as regex, so you can escape it if there's special symbols for regex.

JINO-ROHIT:
thanks @tomaarsen making the changes!

Issue #3121, opened 2024-12-06
Changing Encoding Precision Yields error.


NovaBro:
When I run the following code, I get an error "RuntimeError: linalg.vector_norm: Expected a floating point or complex tensor as input. Got Char" I do not understand how this can be, since the output is a number, and not a char...

```python
from sentence_transformers import SentenceTransformer
from datasets import Dataset
import json, os, torch


device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
model = SentenceTransformer('multi-qa-mpnet-base-cos-v1')
print('start encoding')
sentences = [
    "The weather is so nice!",
    "It's so sunny outside.",
    "He's driving to the movie theater.",
    "She's going to the cinema.",
]

embeddings = model.encode(sentences, show_progress_bar=True, 
        precision='int8')
print(embeddings)
print(type(embeddings))
print(model.similarity(embeddings, embeddings))

```

JINO-ROHIT:
@NovaBro is there a particular reason why you need it to be int8, int8 throws away information compared to float.

If you still want to do this, you can try to convert the int embeddings into float , and then compare

```
from sentence_transformers import SentenceTransformer
from datasets import Dataset
import json, os, torch


device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
model = SentenceTransformer('multi-qa-mpnet-base-cos-v1')
print('start encoding')
sentences = [
    "The weather is so nice!",
    "It's so sunny outside.",
    "He's driving to the movie theater.",
    "She's going to the cinema.",
]

embeddings = model.encode(sentences, show_progress_bar=True, precision='int8')
print(embeddings)
print(type(embeddings))
print(model.similarity(np.float64(embeddings), np.float64(embeddings)))
```

NovaBro:
Hello, thank you for your response!
**There is no specific reason for int8 other than to try to speed up computation.**
This was my understanding from the documentation. 
I had gotten this to work earlier by converting embeddings back to float32 after encoding. 
```python
embeddings = embeddings.astype('float32')
```
Is this the intended use/method to speed up computation? 
If so, I am surprised that the doc does not mention that similarity() requires needs specific dtype. 

<img width="1146" alt="Screenshot 2024-12-08 at 9 38 39 AM" src="https://github.com/user-attachments/assets/6a7104cb-5f65-4ffd-95bf-89458121f651">

JINO-ROHIT:
[...]

PR #3131, opened 2024-12-11
Added a note to the documentation stating that the similarity method does not support embeddings other than non-quantized ones.

pesuchin:
Resolved: https://github.com/UKPLab/sentence-transformers/issues/3121

## Changes
- Updated the documentation for the similarity and similarity_pairwise methods to specify that only non-quantized embeddings are supported.
- Added type hints to indicate that only np.float32 is supported.

If you find any issues or have concerns about these changes, please feel free to suggest adjustments.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -20,11 +20,11 @@ from pathlib import Path
 from typing import Any, Callable, Literal, overload
 
 import numpy as np
+import numpy.typing as npt
 import torch
 import torch.multiprocessing as mp
 import transformers
 from huggingface_hub import HfApi
-from numpy import ndarray
 from torch import Tensor, device, nn
 from tqdm.autonotebook import trange
 from transformers import is_torch_npu_available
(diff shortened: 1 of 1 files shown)


sentence_transformers/SentenceTransformer.py
```diff
@@ -723,14 +723,15 @@ def similarity_fn_name(
     def similarity(self, embeddings1: Tensor, embeddings2: Tensor) -> Tensor: ...
 
     @overload
-    def similarity(self, embeddings1: ndarray, embeddings2: ndarray) -> Tensor: ...
+    def similarity(self, embeddings1: npt.NDArray[np.float32], embeddings2: npt.NDArray[np.float32]) -> Tensor: ...
```
tomaarsen, line 726: Nice, looks to be the correct usage. I wasn't aware of these.
https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray

tomaarsen:
Very nice work! Thanks for this.

- Tom Aarsen

PR #3162, opened 2025-01-11
Allow passing a custom batch sampler to the trainer

### LINKED ISSUE #3152: Improve API to use a custom batch_sampler in a SentenceTransformerTrainer
Currently only specific `batch_sampler` values are possible
There seems to be a need to enable users to create custom batch samplers

Examples from issues:
1. https://github.com/UKPLab/sentence-transformers/issues/3123
2. https://github.com/UKPLab/sentence-transformers/issues/2707
3. https://github.com/UKPLab/sentence-transformers/issues/2954

I believe that the current solution (that i found the hard way) that is suggested in the issues is to subclass `SentenceTransformerTrainer` and override `get_batch_sampler`, is not documented well enough and isn't straightforward

Is there any reason not to just accept anything that inherits from `DefaultBatchSampler` (for example) as a parameter?

alonme:
resolves #3152

Also, I believe there was an issue with the typing and usage of the seed parameter, given that 0 is falsy, it would be ignored if it was passed as a seed.

Missing:
1. Documentation - wanted to make sure this makes sense before writing docs


sentence_transformers/sampler.py
```diff
@@ -77,21 +97,17 @@ def __init__(
         dataset: Dataset,
         batch_size: int,
         drop_last: bool,
-        valid_label_columns: list[str] = None,
-        generator: torch.Generator = None,
-        seed: int = 0,
```
alonme, line 82: I believe there was an issue with the typing and usage of the seed parameter, given that 0 is falsy, it would be ignored if it was passed as a seed.

sentence_transformers/trainer.py
```diff
@@ -574,10 +579,14 @@ def get_batch_sampler(
...
         """
         if isinstance(dataset, IterableDataset):
-            if self.args.batch_sampler != BatchSamplers.BATCH_SAMPLER:
+            if self.args.batch_sampler != BatchSamplers.BATCH_SAMPLER and self._batch_sampler is not None:
                 logger.warning("When using an IterableDataset, you cannot specify a batch sampler.")
             return None
 
+        # If a batch sampler has been explicitly set, we return that
+        if self._batch_sampler:
+            return self._batch_sampler
```
tomaarsen, line 588: I like the changes in the `sampler.py`, but I'm not sure if this is the best option.
This here requires the user to initialize the Batch Sampler themselves, which prevents Sentence Transformers from updating the dataset(s). This breaks the `prompts` feature here: https://github.com/UKPLab/sentence-transformers/blob/c68bf68299a4435c6a48ea15d789fef596bf1444/sentence_transformers/trainer.py#L284-L291

Additionally, it prevents multi-dataset training setups, because there's only one batch sampler possible.

So, I think a good solution is for the `batch_sampler` argument to be 1) a class (not instance) that subclasses `DefaultBatchSampler` or 2) a function that returns a subclass of `DefaultBatchSampler` given `dataset`, `batch_size`, `drop_last`, `valid_label_columns`, `generator`, `seed`, and *args & **kwargs.

What do you think?

atuysuzoglu:
Hi @tomaarsen, did you folks come up with a solution to this?
I would like to use my custom batch sampler but could not do it by overriding the SentenceTransformer class with my class SentenceTransformerWithCustomSampler that implements get_batch_sampler function.
Would appreciate your help.
Thank you!

tomaarsen:
I'd like to add this updated behaviour, so I've added it to this PR myself. In short, you can now:
1. pass a `DefaultBatchSampler` subclass to `...TrainingArguments(..., batch_sampler=MyBatchSampler)`
2. pass a function that returns a `DefaultBatchSampler` instance to `...TrainingArguments(..., batch_sampler=get_my_batch_sampler)`
3. pass a `MultiDatasetDefaultBatchSampler` subclass to `...TrainingArguments(..., multi_dataset_batch_sampler=MyMultiDatasetBatchSampler)`
4. pass a function that returns a `MultiDatasetDefaultBatchSampler` instance to `...TrainingArguments(..., multi_dataset_batch_sampler=get_my_multi_dataset_batch_sampler)`

If you can, feel free to help me review this PR, or let me know if these changes match your goals/expectations, then I can include them in the next release.

- Tom Aarsen

PR #3163, opened 2025-01-11
testcases for community detection

JINO-ROHIT:
This PR adds few tests cases for the community detection module.

Also should we look to cover a case for empty inputs? Currently it accepts empty inputs and throws a embedding computation error.

Who can review?
@tomaarsen

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -5,7 +5,9 @@ import sklearn
 import torch
 
 from sentence_transformers import SentenceTransformer, util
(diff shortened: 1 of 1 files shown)


tomaarsen:
This is excellent, thanks a lot! I renamed some tests to make it easier to tell that they're `community_detection` related. Other than that I just did `pre-commit run --all` to ensure the quality tests pass.

- Tom Aarsen

PR #3167, opened 2025-01-12
[DRAFT] pytests for losses

JINO-ROHIT:
Im writing this PR to help write test cases for all the losses .

Im starting with ContrastiveLoss and I wanted to get some early feedback on this. Like you mentioned, the idea is have a single parametrized func and pass all loss cases into it.

wdyt
@tomaarsen

### DIFF
--- a/tests/losses/test_loss.py
+++ b/tests/losses/test_loss.py
@@ -2,41 +2,99 @@ from __future__ import annotations
 
 import pytest
 import torch
+from datasets import Dataset
+from torch import nn
 
-from sentence_transformers import InputExample, SentenceTransformer
-from sentence_transformers.losses import ContrastiveLoss
+from sentence_transformers import SentenceTransformer
+from sentence_transformers.losses import (
+    CachedGISTEmbedLoss,
+    CachedMultipleNegativesRankingLoss,
+    GISTEmbedLoss,
+    MultipleNegativesRankingLoss,
+    TripletLoss,
+)
+from sentence_transformers.util import batch_to_device
 
-model = SentenceTransformer('all-MiniLM-L6-v2')
+
+@pytest.fixture(scope="module")
+def guide_model():
+    return SentenceTransformer("sentence-transformers-testing/stsb-bert-tiny-safetensors")
 
 
-example_pairs = [
-    InputExample(texts=['This is a positive example', 'This is a similar positive example'], label=1.0),
-    InputExample(texts=['This is a positive example', 'This is a negative example'], label=0.0),
-]
+def get_anchor_positive_negative_triplet():
+    return {
+        "losses": [
+            (MultipleNegativesRankingLoss, {}),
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

I think this is heading in the right direction, but I do have some suggestions:

In particular, I think there's a risk that our test implementation has similar/the same bug as our real implementation. However, we still want to ensure that the loss "works".

So, one option is to create 2 batches. If we want to use the same model for both, then we have to create 1 "good" batch and 1 "bad" batch. The former is normal data, whereas the latter is opposite of reality: the positive is the negative, or related texts are marked as 0.0 similarity, etc.

Then, the same trained model should give a low loss for the "good" batch and a high loss for the "bad" batch. We should in both cases still test that the output 1) is a torch Tensor, 2) not `nan`, 3) not `0.0`, 4) has `requires_grad`, etc.

I think this is safer than just calculating an "expected loss", because if our implementation is buggy, that expected loss is probably also wrong.

And then we'd have to create the good and bad batches for each row in the [Loss Overview](https://sbert.net/docs/sentence_transformer/loss_overview.html), and then use each of the appropriate loss functions with those good/bad batches and ensure that `loss(good_batch) < loss(bad_batch)`.

Does that make sense?

- Tom Aarsen

JINO-ROHIT:
umm i must admit i dont completely understand, do you mind giving an example?

JINO-ROHIT:
yeap, thanks so much tom, working on it!

tomaarsen:
I think there is a way for `pytest` to load a fixture by name, but I think it's almost as hacky as this, so I think this is fine to keep for now! If I ever run into it again, then I can update things.

JINO-ROHIT:
cool then , ill add for other losses

JINO-ROHIT:
wb for losses like softmax? 
this condition fails
```
assert correct_loss < incorrect_loss, "Loss should be lower for correct data than for incorrect data"
```
I assume its since the test tiny model hasnt been trained and the random init makes the losses inconsistent ?

tomaarsen:
Oh, I hadn't thought about that the tiny model might not be trained enough, hah. 
I think I did train them a bit, perhaps you can check if this one does work? https://huggingface.co/sentence-transformers-testing/all-nli-bert-tiny-dense

And otherwise we'll have to use a bigger, normal model, like all-MiniLM-L6-v2.

- Tom Aarsen

JINO-ROHIT:
yeah, this models seems to be working for now. thanks!

JINO-ROHIT:
tried using a larger model and the smaller ones as well, the losses are still kinda random.

Also i think ive added the losses that can be tested together, i think the other losses have to be handled differently, wdyt?

tomaarsen:
Hello!

This was tagged DRAFT in early 2025 and the loss directory layout has changed substantially with the v5.4 refactor (per-archetype subpackages, snake_case filenames). I think I'll close this now as it's also stale.

- Tom Aarsen

PR #3174, opened 2025-01-15
fix: propagate cache dir to find adapter

lauralehoczki11:
`cache_dir` is not currently propagated to the `find_adapter_config_file` function therefore the default cache directory is always used. This has caused us issues when trying to cache models running on a pyspark enabled Dataproc cluster as the agent does not have a `home`.


tomaarsen:
Hello!

Well spotted; thank you for this simple and effective patch.

- Tom Aarsen

PR #3186, opened 2025-01-22
Fix #3185: Support `convert_to_numpy` for `output_value="token_embeddings"`

### LINKED ISSUE #3185: `convert_to_numpy=True` not working with `output_as="token_embeddings"`; returns `torch.Tensor` type
Hey ST Team! 👋

I was trying to get the "token_embeddings" from a `bge-base-en-v1.5` model, and converting it to `numpy.ndarray` but it was not functioning as expected.

Have a look at the screenshot below 👇

Image

As a workaround, I was using `.cpu().numpy()` which was working. 

Thanks! 😊

bhavnicksm:
Hey UKPLab and ST Team! 👋

Just added a fix for #3185

Let me know if it looks good or any changes I need to make to the PR!

Thanks 😊

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -671,9 +671,9 @@ class SentenceTransformer(nn.Sequential, FitMixin, PeftAdapterMixin):
         elif convert_to_numpy:
             if not isinstance(all_embeddings, np.ndarray):
                 if all_embeddings and all_embeddings[0].dtype == torch.bfloat16:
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

I added some tests so that the current behaviour is pretty much what is expected. Sadly, there's an issue with the expected output if the `output_value` is None currently.
I'm rather unhappy with the 2 `convert_to_...` arguments, I would have preferred just 1 parameter for that. I also don't like all the if-else's at the end of the method :/ 

But for now, perhaps we can re-add the removed code, but with `if output_value is None` to resolve the test failures?

- Tom Aarsen

Issue #3199, opened 2025-01-27
Usage when huggingface is not accessible?


Davidyz:
I noticed [here](https://github.com/UKPLab/sentence-transformers/blob/3fbc1833b89e4022915da88d76f3ee1b1984b564/sentence_transformers/model_card.py#L736) that access to the huggingface API is compulsory when calling `set_base_model` method, which happens when creating a new instance of the `SentenceTransformer` class. This blocks the model initialisation when the environment has internet access but cannot access huggingface. Is there a way to get around this? The `local_files_only` parameter in the `SentenceTransformer` class made it look like it's a local mode, but in fact it still requires an internet connection, which is quite frustrating in deployment.

Davidyz:
I implemented [a fix in my own fork](https://github.com/Davidyz/sentence-transformers/commit/71ddba380190743433569dc468a39a9244f9d2be) that would address the issue. I can submit a PR if the maintainers agree.

ir2718:
Hi,

Since you mentioned deployment, an easier option would be to just convert the model to ONNX.

Davidyz:
> Hi,
> 
> Since you mentioned deployment, an easier option would be to just convert the model to ONNX.

Hi thanks for your reply. I'm not using sentence-transformer directly. I'm using it as the embedding function for [chromadb](https://www.trychroma.com/) so the best I can do is to set the `backend` option to `onnx`. Even if I do that, there will be a `requests.exceptions.ConnectionError` being raised. I also want to allow users to select the model to use for embedding, so I cannot just load a hardcoded ONNX model. Is there a way to bypass the request to huggingface API?

ir2718:
I thought you were deploying a single model for a customer. In your particular case, if you've got a set of small models you need to support, you can still convert them and use them with `onnxruntime`. But if the fork fixes your issues, that's great.

Also, I'm curious, have you tried this out with a model that is not cached? I'm not sure how offline mode would work if you need remote code for doing inference, eg. [NV-Embed-v2](https://huggingface.co/nvidia/NV-Embed-v2/blob/main/modeling_nvembed.py).

Davidyz:
> I thought you were deploying a single model for a customer. In your particular case, if you've got a set of small models you need to support, you can still convert them and use them with `onnxruntime`. But if the fork fixes your issues, that's great.
> 
[...]

PR #3202, opened 2025-01-28
Don't call `set_base_model` when `local_files_only=True`

Davidyz:
This avoids calling hfAPI in `set_base_model` when `local_files_only=True`. This means the code will run without creating connections to the internet.

Solves #3199 .


tomaarsen (approved):
This is looking great; thanks for the issue report and setting up this PR. I'll merge this once the tests go green

tomaarsen:
Thanks a bunch!

- Tom Aarsen

PR #3222, opened 2025-02-07
[`v4`] CrossEncoder Training refactor - MultiGPU, loss logging, bf16, etc.

tomaarsen:
Hello!

## Pull Request overview
* Introduce a new training loop, using a `CrossEncoderTrainer`, `CrossEncoderTrainingArguments`, and loss functions. Brings features such as:
  * Multi-GPU training (DP, DDP), multi-dataset training, bf16, loss logging (Weights & Biases, Tensorboard, the terminal, etc.), many additional hyperparameters, automatic model card generation, etc.
* A new MultipleNegativesRankingLoss and CachedMultipleNegativesRankingLoss (a.k.a. InfoNCE and InfoNCE with GradCache) to train with anchor-positive pairs, anchor-positive-negative triplets, and anchor-positive-negative1-...-negativeN tuples.

## TODOs
* [x] Additional documentation
* [x] CE Tests - training & inference
* [x] Implementing more losses from the literature (can also be done after the initial release)
   * [x] MarginMSE
* [x] Determine solid defaults for MNRL and CMNRL
   * [x] Sigmoid with scale=10.0 seems good enough for now
* [x] Test MSELoss - currently never been ran
* [x] Update existing training scripts to the new flow
   * [x] Remove old scripts in the original style
* [x] Rename `fit` to `old_fit` and create a new `fit` method that depends on `CrossEncoderTrainer`. Goal: no real backwards incompatibility with existing training scripts.
* [x] add CE NanoBEIR evaluator

## Details
Overall, the goal of this refactor is to introduce feature parity between the Cross Encoder training and the Sentence Transformer training. Luckily, the work done for the ST trainer can be extended rather easily, so the refactor is not as big as it was for the SentenceTransformer class in v3.0.

Notably, training now centers around:
1. a training [`Dataset`](https://huggingface.co/docs/datasets/main/en/package_reference/main_classes#datasets.Dataset) or [`DatasetDict`](https://huggingface.co/docs/datasets/main/en/package_reference/main_classes#datasets.DatasetDict). This class is much more suited for sharing & efficient modifications than lists/DataLoaders of `InputExample` instances. A `Dataset` can contain multiple text columns that will be fed in order to the corresponding loss function. So, if the loss expects (anchor, positive, negative) triplets, then your dataset should also have 3 columns. The names of these columns are irrelevant at this time. If there is a "label" column, it is treated separately, and used as the labels during training.
    A `DatasetDict` can be used to train with multiple datasets at once, e.g.:
    ```python
[...]

PR #3225, opened 2025-02-11
Additional Trainer Argument for features of different modalities

mengerj:
Hi,

I have been working on multi-modal embedding models based on the sentence-transformers framework and found that I need to hard code keys into the "collect_features" method within the trainer. I simply **added an argument to allow passing a list of additional keys** that might be present in your features (similar to "pixel_values" for CLIP). I would appreciate a merge so I don't need to rely on my fork for the multi-modal embedding model.

Best,
Jonatan


tomaarsen:
Hello!

I think this is indeed quite valuable to support, so I'd like to help you out here. However, I think we might be better off going in a different direction that should work out of the box without the user having to ever specify anything: https://github.com/UKPLab/sentence-transformers/pull/3276

Feel free to give me some feedback there!

Also, apologies for the delays. I've been focusing on the huge #3222, which is now nearing completion.

- Tom Aarsen

PR #3238, opened 2025-02-18
Changed value error for missing model into FileNotFoundError

PhorstenkampFuzzy:


tomaarsen (approved):
This strikes me as very sensible, thank you!

PR #3244, opened 2025-02-21
Model Card Compatability with Asymmetric Models

OsamaS99:
Hello @tomaarsen,

So I came across another issue with the asymmetric models compatibility. So without that fix, I get `ZeroDivisionError: integer division or modulo by zero`


tomaarsen:
Thanks a bunch!

- Tom Aarsen

PR #3253, opened 2025-02-28
feat: add 'Path' parameter for ModelCard template

sam-hey:
Allowing users to pass the `Path` to a template to create the model card from it:

`SentenceTransformerModelCardData( model_card_template_path=...)`

This eliminates the need to override `_create_model_card()`, making it easier for `PyLate`, and enables users to provide their own model template.

cc @NohTow

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -1266,7 +1266,7 @@ class SentenceTransformer(nn.Sequential, FitMixin, PeftAdapterMixin):
                 )
         else:
             try:
-                model_card = generate_model_card(self)
+                model_card = generate_model_card(self, self.model_card_data.model_card_template_path)
             except Exception:
                 logger.error(
                     f"Error while generating model card:\n{traceback.format_exc()}"
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -1266,7 +1266,7 @@ class SentenceTransformer(nn.Sequential, FitMixin, PeftAdapterMixin):
                 )
         else:
             try:
-                model_card = generate_model_card(self, self.model_card_data.model_card_template_path)
+                model_card = generate_model_card(self)
             except Exception:
                 logger.error(
                     f"Error while generating model card:\n{traceback.format_exc()}"
--- a/sentence_transformers/model_card.py
+++ b/sentence_transformers/model_card.py
@@ -253,7 +253,6 @@ class SentenceTransformerModelCardData(CardData):
             e.g. "semantic textual similarity, semantic search, paraphrase mining, text classification, clustering, and more".
         tags (`Optional[List[str]]`): A list of tags for the model,
             e.g. ["sentence-transformers", "sentence-similarity", "feature-extraction"].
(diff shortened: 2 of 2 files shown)


tomaarsen:
Hello!

So I understand correctly, what do you intend to override exactly for PyLate? the `model.model_card_data.model_card_template_path = "..."`? (I'm not really sure what the current problem is that you're having on PyLate.)

If so, we might not have to introduce the `template_path` argument in `generate_model_card` at all, it can just grab it from `model.model_card_data.model_card_template_path`.

Also, as a nit: the `model_card_` part of `model_card_template_path` is already kind of implied by that attribute living on `SentenceTransformerModelCardData`, so just `template_path` might make more sense.

- Tom Aarsen

NohTow:
Basically, IIUC, right now the main issue is that we have to overwrite `_create_model_card` in PyLate to be able to call the PyLate `generate_model_card` that is basically the same as ST, up to the `ModelCard.from_template` that points towards our local .md (and the HF emoji of the dog, but I am not even sure where it is used ahah).

It was working before as I was overwriting `_create_model_card` but removed it because you change it and I was not recalling why I did overwrite it. Honestly ; I am fine with overwriting it if you do not like adding a path, we already overwrite other stuff for the card. PyLate was mostly build by extending ST without asking for modifications (except when they is no way around it), but obviously sometimes the extension could be prettier with some modifications, so up to you @tomaarsen.

tomaarsen:
> (and the HF emoji of the dog, but I am not even sure where it is used ahah).

In ST it's (only) used in the inference snippet:
image

I couldn't get it working in Jinja itself, so I used this trick instead:
https://github.com/UKPLab/sentence-transformers/blob/5c362dc1f7b069ebad20c44380566f40bfedb4e5/sentence_transformers/model_card_template.md?plain=1#L81

I think this implementation is quite neat, so I'm open to moving forward with this if @NohTow agrees that it would be valuable for simplifying PyLate.

Also, apologies for the delay. I've been trying to focus on https://github.com/UKPLab/sentence-transformers/pull/3222 in preparation for a big v4.0 release.

- Tom Aarsen

tomaarsen:
Also, @NohTow, I'm guessing that you can just override `template_path` in your `SentenceTransformerModelCardData` subclass?
In that case, I'll move `template_path` around a bit and take it out of the `__init__`, as normal users don't need to update it.

- Tom Aarsen

NohTow:
>  if @NohTow agrees that it would be valuable for simplifying PyLate.

Well it allows us to not override a function just for this, so it is always good, as we saw it broke PyLate once!

> Also, @NohTow, I'm guessing that you can just override template_path in your SentenceTransformerModelCardData subclass?

I think we can simply feed the template_path in this [line](https://github.com/lightonai/pylate/blob/2d094a724866d6e15701781528368438081c0157/pylate/models/colbert.py#L228) no? 

Do you know if we ever feed a model card directly to the init? I am afraid it could break with re-using previous card that did not have this parameter though (and will default to ST template)?

sam-hey:
Hey Tom,

Thanks a lot for merging! Looking forward to seeing more awesome work on ST  :) 

Best
Sam

PR #3260, opened 2025-03-06
Fix: Prevent duplicate positive samples being used as negatives in GISTEmbedLoss (#2756)

### LINKED ISSUE #2756: Modify GISTEmbedLoss and CachedGISTEmbedLoss to automatically remove duplicate positives from being considered negatives

tjwjdrok:
## Summary
This PR fixes an issue in `GISTEmbedLoss` where duplicate positive samples were incorrectly considered as negatives.  
- Added masking logic to prevent positives from being selected as negatives.
- Improved in-batch negative sampling.

## Changes
- Updated `GISTEmbedLoss.py` to mask duplicate positive samples.
- Added a unit test in `tests/losses/test_gist_embed_loss.py`.

## Related Issue
Fixes #2756

### DIFF
--- a/sentence_transformers/losses/GISTEmbedLoss.py
+++ b/sentence_transformers/losses/GISTEmbedLoss.py
@@ -153,7 +153,6 @@ class GISTEmbedLoss(nn.Module):
         aa_sim[mask] = -torch.inf
         pp_sim[mask] = -torch.inf
 
-        # 기존 마스킹 로직 유지
         ap_sim[guided_ap_sim > guided_sim] = -torch.inf
         aa_sim[guided_aa_sim > guided_sim] = -torch.inf
         pp_sim[guided_pp_sim > guided_sim] = -torch.inf
@@ -165,7 +164,6 @@ class GISTEmbedLoss(nn.Module):
             an_sim = self.sim_matrix(anchor, negative)
             guided_an_sim = self.sim_matrix(anchor_guide, negative_guide)
             
-            # 중복된 음성 샘플을 제외
             mask = (an_sim == 1.0)
             an_sim[mask] = -torch.inf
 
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Thanks for opening this. I did some more tests with the proposal here, but I was surprised to see that the `mask` was always exclusively False, despite me using a dataset with a lot of overlap. I looked into it a bit more, and it seems like I made a critical error in my reasoning in #2756: whereas 2 embeddings of the same text are normally identical in inference, **but during training the Dropout makes them different!**.

This means that using e.g. `pp_sim == 1.0` does not work to identify identical input texts. This also explains why @3074 from @JINO-ROHIT ended up failing.

The only way to do this is to track the inputs from the start - before they're embedded. This means detokenizing the inputs (as you can't easily compare `input_ids` due to padding), computing a `[batch_size, batch_size]` mask that tracks all overlapping texts between all positives (and `[batch_size*2, batch_size*2]` if you also have a negative column). This gets expensive, slow, while using `batch_sampler="no_duplicates"` solves the problem already.

I was expecting that there would be a simple fix like what you proposed, but I'm afraid I don't think there is. I'll close this, as I think the original idea behind it is already flawed.

- Tom Aarsen

Issue #3241, opened 2025-02-19
TSADE data preparation


HuangBugWei:
It seems that in standard embedding model training with the Masked Language Modeling (MLM) objective, the masked tokens are randomly selected during training. The original implementation usually achieves this by passing a collate_fn to apply random masking at the data loading stage. 
https://github.com/huggingface/transformers/blob/9f51dc25357bcde280a02b59e80b66248b018ca4/examples/pytorch/language-modeling/run_mlm_no_trainer.py#L531-L538

However, in the current implementation of TSADE, it looks like the noisy transformation is only applied once during the dataset initialization, rather than dynamically at each iteration.
https://github.com/UKPLab/sentence-transformers/blob/82fa8415b60f85ffb30e8d29353e40bd8fd1bafc/examples/unsupervised_learning/TSDAE/train_tsdae_from_file.py#L55C1-L76C52

Is this intended behavior in TSADE?
If so, what are the considerations behind this design choice compared to dynamic masking during training?
Would it be beneficial to introduce a more dynamic masking strategy similar to the standard MLM approach?

If my understanding is incorrect, please let me know how TSADE handles token masking differently from the standard approach. Any insights or clarifications would be greatly appreciated.

Since @nreimers is one of the authors of the TSADE paper and also contributed to the original DenoisingAutoEncoderDataset.py implementation, I wanted to reach out for clarification regarding this design choice. 
Thanks for your time and for maintaining this great project!

PR #3265, opened 2025-03-12
Fix: dynamic noise addition during training in TSADE examples

HuangBugWei:
The TSADE examples currently use a fixed data deletion strategy, as mentioned in issue #3241. Upon reviewing the legacy code (specifically the [dataset preparation](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/datasets/DenoisingAutoEncoderDataset.py) for the v2.x training style), I found that noise addition was dynamic during training. In the current approach, the fixed deletion occurs because the deletion is only applied once at the beginning, with no further dynamic noise added. This PR aims to align the data preparation with the dynamic approach from the legacy code, ensuring that noise addition is performed randomly throughout the training process.

I use ```set_transform``` from ```datasets.arrow_dataset.Dataset``` to obtain this feature.

Notes:
* The other two example files have not been modified in this way.
* I used ```set_transform``` on the dataset rather than applying a ```data_collator``` to the trainer. This approach is simpler, though I’m unsure if it might have any unintended side effects.
* The current method may be somewhat redundant, as tokenization occurs at every step. A more efficient approach would involve storing the tokenized sentences (as a ```list[str]```) in the noisy column. Then, ```set_transform``` could randomly select the kept words and concatenate them into a noisy sentence. However, this approach may be confusing because it’s unclear why the noisy column would store the tokenized sentences in the first place.

@tomaarsen Your feedback and suggestions are much appreciated!

### DIFF
--- a/examples/unsupervised_learning/TSDAE/train_tsdae_from_file.py
+++ b/examples/unsupervised_learning/TSDAE/train_tsdae_from_file.py
@@ -49,7 +49,15 @@ with (
         line = line.strip()
         if len(line) >= 10:
             train_sentences.append(line)
-dataset = Dataset.from_dict({"text": train_sentences})
+
+dataset = Dataset.from_dict({"noisy": train_sentences, "text": train_sentences})
+
+
+def noise_transform(batch):
+    if "noisy" not in batch:
+        raise ValueError("Input batch must contain a 'noisy' key.")
+    batch["noisy"] = [noise_fn(text) for text in batch["noisy"]]
+    return batch
 
 
 def noise_fn(text, del_ratio=0.6):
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

When I rewrote the TSDAE script, I indeed forgot to consider that when applying multiple epochs, the subsequent epochs would use the exact same noise as the first one. Because of this, I think it indeed makes sense to use `set_transform` instead, which is exactly intended for this use case: an on-the-fly modification of the dataset that differs every epoch.

I can take over from here and try to adapt this to the other files as well. I also think we can remove `noise_fn` in favor of only having a `noise_transform` function.

- Tom Aarsen

tomaarsen:
I merged the two noisy functions, and propagated the changes to all 3 files. What do you think?

HuangBugWei:
Thank you for merging the functions. The merged noisy function is excellent and concise. However, I found and fixed a few minor issues and added more detailed documentation to the function.

Since I only modified one file, I would appreciate it if you could propagate these changes to the remaining two files after you have reviewed them.

tomaarsen (approved):
I've normalized the noise transform across all 3 scripts. I think it looks quite solid now, what do you think?

I'll merge this when the tests go green and if you're happy about my final commit.

- Tom Aarsen

HuangBugWei:
Thanks for propagating that. Your final commit looks good to me. 
The current test failures (timeout and file not found) seem unrelated to these changes.

Issue #1605, opened 2022-06-23
Is it possible to resume training from a saved checkpoint?


issam9:
Hi, i'm trying to resume training from checkpoints that i save while finetuning a sentence transformer. I tried the following but it starts training from scratch.

```
model = SentenceTransformer('checkpoint_path')
model.fit([(loader, loss)])
```

nreimers:
This is sadly currently not supported

steve-marmalade:
Came here looking for this, so I figured I'd leave a +1. Something equivalent to the `resume_from_checkpoint` flag in the huggingface [transformers.Trainer.train](https://huggingface.co/docs/transformers/main_classes/trainer#transformers.Trainer.train) would be awesome.

Is this a high level of effort to implement?

jaideep11061982:
@nreimers  at least if we do load from a saved best ckps , then it should have updated weights to start from epoch 0 ?

david-waterworth:
@nreimers any plans on implementing this? It's critical when training on sagemaker using spot instances - the training script must be able to be interruptible as sagemaker will stop/start it at will based on availability (and using spot instances is much cheaper).

tomaarsen:
Yes @david-waterworth, the intention is to include this in the upcoming Trainer refactor in #2449.

- Tom Aarsen

NRamirez01:
@tomaarsen My open PR adds resume_to_checkpoint to fit() in fitmixin.py which is what ollama uses at least for SentenceTransformersFinetuneEngine

PR #3269, opened 2025-03-17
Update to fit_mixin.fit to allow fine tuning to resume from a checkpoint

NRamirez01:
### **Summary**
Adding a parameter to fit_mixin.fit method that allows for a fine-tune to be continued from the latest checkpoint if a checkpoint exists in the checkpoint directory

### Motivations
A faulty PSU and many hours of fine-tuning going to waste. The method in fit_mixin.fix has a parameter that allow for checkpoints to be saved, and the method that it uses for training, Transformer's trainer.train has a parameter that takes in a str, or a bool, that allows for resuming from a checkpoint, this change takes advantage of that parameter so that the saved checkpoints can now be used.

### Results 
When a checkpoint file is found, fine-tuning will begin from the latest step found inside of training_state.json, the good thing about this is it allows for continuing a fine-tune without retraining the model on data that it has already been processed.

### Changes
Adds an optional parameter resume_from_checkpoint to fit_mixin.fit that prompts a check for checkpoints in checkpoint_directory and continues fine-tuning from there if it finds them

### Related Issue
It is related to #1605 and gives a way to use resume_from_checkpoint at least with the fit() method in fit_mixin

### DIFF
--- a/sentence_transformers/fit_mixin.py
+++ b/sentence_transformers/fit_mixin.py
@@ -183,7 +183,7 @@ class FitMixin:
         checkpoint_path: str = None,
         checkpoint_save_steps: int = 500,
         checkpoint_save_total_limit: int = 0,
-        resume_from_checkpoint: bool = False
+        resume_from_checkpoint: bool = False,
     ) -> None:
         """
         Deprecated training method from before Sentence Transformers v3.0, it is recommended to use
@@ -387,12 +387,11 @@ class FitMixin:
 
         if checkpoint_path is not None and resume_from_checkpoint:
             if os.path.exists(checkpoint_path) and os.path.isdir(checkpoint_path):
-                logger.info(
-                    f"Looking for checkpoints in: {checkpoint_path}"
-                )
+                logger.info(f"Looking for checkpoints in: {checkpoint_path}")
 
                 all_checkpoints = [
-                    checkpoint for checkpoint in os.listdir(checkpoint_path)
+                    checkpoint
+                    for checkpoint in os.listdir(checkpoint_path)
                     if checkpoint.startswith("checkpoint-") and checkpoint.split("-")[-1].isdigit()
                 ]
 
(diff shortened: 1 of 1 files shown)


NRamirez01:
Reverted a push that was stylistic because it did not play well with delinter.

tomaarsen:
I'll take care of the formatting, no worries :)
Normally I would not really update the `model.fit()` code anymore - it's primarily meant as backwards compatibility as I generally recommend using the `SentenceTransformerTrainer` now. However, as you mentioned, LlamaIndex and some others still rely on it, so I'm okay with making a change like this.

I'll play around with it myself in a while, before the v4.0 release that's coming soon.

- Tom Aarsen

tomaarsen:
You can also run the formatter yourself indeed! The details are here: https://github.com/UKPLab/sentence-transformers?tab=readme-ov-file#development-setup
In short, `pre-commit install` will (for this repository only) set up "pre commit hooks" that run the formatting loop whenever you try and commit something. Because these changes are already committed, you can then also run `pre-commit run --all` to run the formatting on everything (and not just the changes that you're trying to commit).

- Tom Aarsen

NRamirez01:
Code is now acceptable to ruff and ruff-format. Thank you for pointing that out and next time I will read the README!

sentence_transformers/fit_mixin.py
```diff
@@ -238,6 +239,8 @@ def fit(
                 steps
             checkpoint_save_total_limit: Total number of checkpoints to
                 store
+            resume_from_checkpoint: If true, searches for checkpoints
+            to continue training from.
```
tomaarsen, line 243:
```suggestion
            resume_from_checkpoint: If true, searches for checkpoints
                to continue training from.
```
Small nitpick to get the docs to render nicely

tomaarsen (approved):
Looks very solid! Well done, thanks. I ran it locally and it worked out of the box.

PR #3276, opened 2025-03-21
Update collect_features to allow different modalities more easily in the Trainer

tomaarsen:
Alternative to #3225

Hello!

## Pull Request overview
* Update collect_features to allow different modalities more easily in the Trainer

## Details
This is a follow-up of #3225 by @mengerj. He noticed that if you want to train with more custom modules that don't tokenize with `input_ids`, `pixel_values`, or `sentence_embedding`, then `collect_features` won't recognize your features. He proposed to allow users to provide more options than those 3, but I think a refactor of `collect_features` might be more useful:

Instead of relying on the suffixes in e.g. `query_input_ids`, or `sentence_1_pixel_values`, we could also rely on the data collator to provide us with information about which feature columns exist (e.g. `query`, `sentence_1`) - the data collator has easy access to this.

This means that all (custom) features should work out of the box, without the user having to specify anything special. What do you think, @mengerj?

- Tom Aarsen

### DIFF
--- a/sentence_transformers/data_collator.py
+++ b/sentence_transformers/data_collator.py
@@ -46,6 +46,7 @@ class SentenceTransformerDataCollator:
                 column_names.remove(label_column)
                 break
 
+        feature_columns = []
         for column_name in column_names:
             # If the prompt length has been set, we should add it to the batch
             if column_name.endswith("_prompt_length") and column_name[: -len("_prompt_length")] in column_names:
@@ -55,6 +56,9 @@ class SentenceTransformerDataCollator:
             tokenized = self.tokenize_fn([row[column_name] for row in features])
             for key, value in tokenized.items():
                 batch[f"{column_name}_{key}"] = value
+            feature_columns.append(column_name)
+
+        batch["feature_columns"] = feature_columns
 
         return batch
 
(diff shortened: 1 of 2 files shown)


mengerj:
Hi @tomaarsen,

I think the idea is good and I would like to work on it with you. I am currently on vacation, but in April I will take a closer look at it. Maybe we could also have a video chat to talk about the specifics and other ways to improve support for multiple modalities?

Best,
Jonatan

tomaarsen:
Thanks for having a quick look - sadly it seems like `accelerate` doesn't like it when I try to return a list of non-tensors - it will try and `concatenate` them, but that fails with strings. In short, this fix doesn't work as well as I would have hoped. An alternative is to look at the dataset `column_names` as those are also the features, but [`set_transform`](https://huggingface.co/docs/datasets/v3.4.1/en/package_reference/main_classes#datasets.Dataset.set_transform) makes it so it's possible for a dataset to return data with different names than `column_names`, which complicates things.

I'll put this on draft for now, so we can tackle it after v4.0+.

As for a video chat - I prefer working asynchronously over chat on GitHub, so that would be my preference.

- Tom Aarsen

PR #3295, opened 2025-03-28
[`fix`] Avoid "Only if model is wrapped" check which is faulty for FSDP

### LINKED ISSUE #3023: FSDP Training with Sentence Transformer
Given there are so many LLM-based models on top of MTEB benchmark nowadays, is there a canonical way to train with FSDP now? I'm trying to explore along this direction, but I just want to ask if there already exists some examples before I rebuild the wheel.

tomaarsen:
Closes #3023, see https://github.com/UKPLab/sentence-transformers/issues/3023#issuecomment-2760519375

Hello!

## Pull Request overview
* Avoid "Only if model is wrapped" check which is faulty for FSDP

## Details
According to experiments by @meshidenn, by understanding that `model != self.model` checks whether a model is wrapped is faulty. In FSDP, even `self.model` is wrapped, so then both are the same.

cc @ShengYun-Peng

- Tom Aarsen

### DIFF
--- a/sentence_transformers/trainer.py
+++ b/sentence_transformers/trainer.py
@@ -399,7 +399,6 @@ class SentenceTransformerTrainer(Trainer):
         # if the loss stores the model. Only called once per process
         if (
             model == self.model_wrapped
-            and model != self.model  # Only if the model is wrapped
             and hasattr(loss_fn, "model")  # Only if the loss stores the model
             and loss_fn.model != model  # Only if the wrapped model is not already stored
         ):


ShengYun-Peng:
Thanks, @tomaarsen! I'm curious if deepspeed would also benefit from this PR. To use deepspeed with model.encode, I assume all the downstream evaluation on MTEB will only happen for gpu 0, which may cause other gpus to hang if the postprocessing time is super long.

PR #3299, opened 2025-03-30
CachedGISTEmbedLoss Adding Margin

daegonYu:
## 🧪 Margin-enhanced GISTEmbedLoss: Inspired by NV-Retriever

Hello. I read the paper **[[NV-Retriever: Improving text embedding models with effective hard-negative mining](https://arxiv.org/pdf/2407.15831v2)](https://arxiv.org/pdf/2407.15831v2)** and applied its ideas to **GISTEmbedLoss** by introducing a margin-based filtering mechanism. Below, we share interesting experimental results and insights.

---

### 🔧 Key Update: Margin-Based False Negative Filtering

The updated `CachedGISTEmbedLoss` introduces a mechanism to **filter false negatives** using explicit **margin constraints**.

Previously, negative samples more similar than the positive were masked out, but now you can specify how *close* is *too close* by using either:

- `"absolute"`:  
  A negative is discarded if  
  `similarity(neg) > similarity(pos) - margin`
  
- `"percentage"`:  
  A negative is discarded if  
  `similarity(neg) > similarity(pos) * margin`

This filtering is guided by a separate **teacher model** to identify potentially mislabeled negatives. The filtering logic is encapsulated in `mask_false_negatives()`, which masks out these negatives in the similarity matrix by assigning `-inf`, thus ignoring them during softmax loss calculation.

#### ✅ Benefits:
- Reduces the impact of **false negatives** during training
- Improves model robustness and **training stability**
- Especially effective for **large batch sizes**, where false negatives are more frequent
- Inspired by techniques in the **NV-Retriever** paper

If `margin_strategy=None`, the loss behaves like the original GISTEmbedLoss (negatives more similar than the positive are masked, but no margin threshold is applied).

---

### ⚙️ Experimental Setup

To evaluate the effectiveness of this margin-based filtering, we trained models with two different batch sizes:

#### **Setting 1**
- Batch size: `512`  
- Learning rate: `2e-5`

#### **Setting 2**
- Batch size: `2048`  
- Learning rate: `4e-5`

These settings allow us to observe how **larger batches** (which naturally introduce more difficult negatives) interact with the margin strategy.

📌 The rest of the training setup (optimizer, scheduler, warmup, etc.) follows the official SBERT `Trainer` configuration:  
🔗 [[SBERT Training Overview – Trainer Section](https://sbert.net/docs/sentence_transformer/training_overview.html)](https://sbert.net/docs/sentence_transformer/training_overview.html)

We trained each model for **1 epoch** on a 100,000-sample subset of the [[AllNLI Triplet Dataset](https://huggingface.co/datasets/sentence-transformers/all-nli)](https://huggingface.co/datasets/sentence-transformers/all-nli):

```python

dataset = load_dataset("sentence-transformers/all-nli", "triplet")
train_dataset = dataset["train"].select(range(100_000))
```

---

### 📊 Experimental Results

Across both batch sizes, the following performance order (based on **cosine accuracy**) was observed:

```
CachedMultipleNegativesRankingLoss < CachedGISTEmbedLoss < CachedGISTEmbedLoss with Margin
```

Furthermore, the **magnitude of the margin** had a measurable effect:

- In `"absolute"` margin setting:
  - `margin = 0.1` outperformed `margin = 0.05`

- In `"percentage"` margin setting:
  - `margin = 0.9` outperformed `margin = 0.95`

These results demonstrate that:
> **Applying a margin effectively filters out false negatives, leading to more stable and effective training than previous loss functions.**

We believe that even stronger improvements can be observed when scaling to **larger datasets (>100k)**, where false negatives become even more prevalent.

---

### 🔗 Resources

- 📎 **Reference Paper**: [[NV-Retriever (arXiv)](https://arxiv.org/pdf/2407.15831v2)](https://arxiv.org/pdf/2407.15831v2)  
- 🧪 **Experimental Code (Colab)**: [[Colab Notebook](https://colab.research.google.com/drive/1-csyL71tXNXSL-FOeg1x_r9le7AZCVYx?usp=sharing)](https://colab.research.google.com/drive/1-csyL71tXNXSL-FOeg1x_r9le7AZCVYx?usp=sharing)  
- 📊 **Training Logs (W&B)**: [[W&B Experiment Page](https://wandb.ai/ysg5025/mpnet-GIST-margin?nw=oht3iaao7p)](https://wandb.ai/ysg5025/mpnet-GIST-margin?nw=oht3iaao7p)

### DIFF
--- a/sentence_transformers/losses/CachedGISTEmbedLoss.py
+++ b/sentence_transformers/losses/CachedGISTEmbedLoss.py
@@ -302,7 +302,7 @@ class CachedGISTEmbedLoss(nn.Module):
             pp_sim = self.sim_matrix(concatenated_reps[1][b:e], concatenated_reps[1])  # positive-positive similarity
 
             # This uses guided (teacher) similarity as a dynamic threshold to identify and suppress false negatives
-            def mask_false_negatives(guided_sim_mat, raw_sim_mat, positive_mask: Tensor | None = None):
+            def mask_false_negatives(guided_sim_mat, sim_mat, positive_mask: Tensor | None = None):
                 if self.margin_strategy == "absolute":
                     # Remove samples whose guided similarity is higher than (positive_sim - margin)
                     mask = guided_sim_mat > (guided_sim - self.margin)
@@ -316,15 +316,12 @@ class CachedGISTEmbedLoss(nn.Module):
                 if positive_mask is not None:
                     # Ensure true positive pairs are not masked out
                     mask = mask & ~positive_mask
-                raw_sim_mat[mask] = -torch.inf
-                return raw_sim_mat
+                sim_mat[mask] = -torch.inf
+                return sim_mat
 
             # Create a mask to protect true positive pairs in the anchor-positive matrix (i.e., diagonal elements)
-            positive_mask = torch.zeros_like(guided_ap_sim, dtype=torch.bool)
-            for i in range(guided_ap_sim.size(0)):
-                pos_idx = i + b
-                if pos_idx < guided_ap_sim.size(1):
-                    positive_mask[i, pos_idx] = True
+            positive_mask = torch.eye(*guided_ap_sim.shape, dtype=torch.bool, device=guided_ap_sim.device)
+            positive_mask = positive_mask.roll(b)
 
             # Apply false negative suppression to each similarity matrix using guided similarity as anchor
             ap_sim = mask_false_negatives(guided_ap_sim, ap_sim, positive_mask=positive_mask)  # anchor-positive
--- a/sentence_transformers/losses/CachedGISTEmbedLoss.py
+++ b/sentence_transformers/losses/CachedGISTEmbedLoss.py
@@ -107,7 +107,6 @@ class CachedGISTEmbedLoss(nn.Module):
             - Efficient Natural Language Response Suggestion for Smart Reply, Section 4.4: https://arxiv.org/pdf/1705.00652.pdf
             - Scaling Deep Contrastive Learning Batch Size under Memory Limited Setup: https://arxiv.org/pdf/2101.06983.pdf
             - GISTEmbed: Guided In-sample Selection of Training Negatives for Text Embedding Fine-tuning https://arxiv.org/abs/2402.16829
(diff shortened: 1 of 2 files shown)


daegonYu:
Hi,

During CI testing, one of the tests failed due to a network timeout when trying to download the test model from Hugging Face Hub.

I'd appreciate it if you could kindly take a look when you have a moment.  
Thank you!

tomaarsen:
Hello!

Thank you for the PR, no need to worry about the CI too much, it's been having issues all over the repository. I'll have a look at this next week!

- Tom Aarsen

daegonYu:
Hi!

That sounds great — I’m really glad to hear the W&B report was helpful!
Thanks for taking the time to integrate the changes into GISTEmbedLoss as well.
I like the use of eye and roll for the positive_mask — it’s clean and efficient. Nice touch!

Thanks!

tomaarsen:
I made another proposal in 2795cd046529532923b939f2618fbee3f326c19b & 68304f093d9bbf1bc34dc84d5d8facaacf3dfaf5 and would love your opinion: I removed `margin_strategy=None` as it's identical to `margin_strategy="absolute"` with `margin=0.0`. 

I think this will be slightly simpler for the end user, what do you think?

- Tom Aarsen

daegonYu:
That’s a great idea. I think it will make things more intuitive for the end user. Thanks for suggesting it!

daegonYu:
Thanks so much for merging this in!
Regarding the point you raised — if you're aiming to follow the NV-Retriever paper precisely, I believe relative_margin=0.95 would be more accurate, as it aligns with their original formulation.

If I may gently add a thought — the variable name relative_margin could potentially be a bit ambiguous. It’s unclear whether it refers to TopK-MarginPos or TopK-PercPos as defined in the NV-Retriever paper. To improve clarity for end users, especially those referencing the paper, it might be worth considering a more descriptive name like percentage_margin or perc_margin to explicitly convey its meaning.

tomaarsen:
We can definitely do that! I'm more than happy for us to update the margin in the GIST losses to align with the updated margin in mine_hard_negatives.

- Tom Aarsen

PR #3309, opened 2025-04-05
Update README.md

lakshminarasimmanv:
The Towards Data Science article no longer exist. Instead, I found the same in a different site.


tomaarsen:
Hello!

Thank you so much for finding the updated source! I noticed a while back that it had been removed, but I couldn't find a replacement source, so I left it.

I saw that there's a 2nd place where this same article is linked, so I also updated this one. Thank you!

- Tom Aarsen

PR #3314, opened 2025-04-08
Exclude transformers==4.51.0 due to upstream bug

VitoDrofenik:
transformers==4.51.0 introduced a bug that has been fixed in 4.51.1.

This PR updates the version specifier to explicitly exclude 4.51.0 using:

    transformers>=4.41.0,<5.0.0,!=4.51.0

This ensures no one installs the broken version by accident.

Reference:
- https://github.com/huggingface/transformers/releases/tag/v4.51.1

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Artificial Intelligence",
 ]
 dependencies = [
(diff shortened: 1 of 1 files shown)


tomaarsen:
With the transformers 4.51.1 release already out, I don't think it's necessary for Sentence Transformers to explicitly exclude this one breaking version - it'll only be installed if it's specifically requested.

- Tom Aarsen

PR #3323, opened 2025-04-14
Update (C)GIST losses to support "relative" margin instead of "percentage"

tomaarsen:
Follow-up of #3299 and #3321

Hello!

## Pull Request overview
* Update (C)GIST losses to support "relative" margin instead of "percentage"
* Use an inverted margin to the current unreleased implementation: a relative margin of 0.05 means that the anchor-negative similarity can be as most 95% of the anchor-positive similarity.

## Details
See discussion here: https://github.com/UKPLab/sentence-transformers/pull/3299#issuecomment-2798369903

cc @daegonYu Feel free to have a look at this :)

- Tom Aarsen


daegonYu:
I am very satisfied, this will help users avoid confusion with the variables used in mine_hard_negatives() . 

Thanks!

tomaarsen:
I agree completely, thanks for having a look!

- Tom Aarsen

PR #3324, opened 2025-04-14
[`typing`] Fix typing for CrossEncoder.to

### LINKED ISSUE #3322: Wrong method signature in CrossEncoder.py
The `to(device)` method returns a reference to the moved model [after moving to the given device], but the method signature declares `None` as return value:
```
   def to(self, device: int | str | torch.device | None = None) -> None:
        return self.model.to(device)
```
Causing my IDE to complain. :-) 

https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/cross_encoder/CrossEncoder.py#L656

tomaarsen:
Resolves #3322

Hello!

## Pull Request overview
* Fix typing for CrossEncoder.to

## Details
Since we're now subclassing `nn.Module`, we don't actually need to override `.to` at all, so I can just remove it and rely on the superclass' `to`, which has some nice overrides and should have correct typings with `Self` as the return. 

Thank you @jeyendranbalakrishnan for reporting this!

- Tom Aarsen

### DIFF
--- a/sentence_transformers/cross_encoder/CrossEncoder.py
+++ b/sentence_transformers/cross_encoder/CrossEncoder.py
@@ -653,9 +653,6 @@ class CrossEncoder(nn.Module, PushToHubMixin, FitMixin):
             return folder_url.pr_url
         return folder_url.commit_url
 
-    def to(self, device: int | str | torch.device | None = None) -> None:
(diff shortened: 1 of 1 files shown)


jeyendranbalakrishnan:
Thank you!

PR #3334, opened 2025-04-18
feature - mine hard negatives working with prompts

GivAlz:
Hi sentence transformers DEVs.

 This is my first PR here so first, thank you for your precious package that I have been using for a while!
 
 Recently I have tried to mine hard negatives on a difficult dataset and I noticed that some of the best models do require a prompt. This PR is a minor change to the functions and documentation that makes it possible to pass a prompt.

I have just a couple of comments:
- I have copied and modified the description of the parameters from model.encode, as I found the original description a bit confusing
- I have added a very basic test, which essentially checks some negatives are found on a mock up dataset

### DIFF
--- a/sentence_transformers/util.py
+++ b/sentence_transformers/util.py
@@ -324,6 +324,8 @@ def paraphrase_mining(
     max_pairs: int = 500000,
     top_k: int = 100,
     score_function: Callable[[Tensor, Tensor], Tensor] = cos_sim,
+    prompt_name: str | None = None,
+    prompt: str | None = None,
 ) -> list[list[float | int]]:
     """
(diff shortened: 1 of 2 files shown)


GivAlz:
I have added some tests that run locally (there were none for mine hard negatives).

For what I see most of the errors are due to huggingface not answering (and some other tests not failing, but not because of code I wrote).

Please let me know if there's anything I can do/should do.

tomaarsen:
I realised that we should split up the `prompt_name` & `prompt` into query/corpus, so I've done that now. Feel free to let me know what you think.

- Tom Aarsen

GivAlz:
oh...good catch! Thank you @tomaarsen !

Issue #3247, opened 2025-02-26
Loading a PEFT model uses `peft.peft_model.PeftModelForFeatureExtraction`; incompatible with ST PEFT Methods


tomaarsen:
Related: #3246

If you load a PEFT model like [sentence-transformers-testing/stsb-bert-tiny-lora](https://huggingface.co/sentence-transformers-testing/stsb-bert-tiny-lora), you can't call e.g. `add_adapter` on it because the underlying model is not a `transformers.integrations.peft.PeftAdapterMixin`, but a `peft.peft_model.PeftModelForFeatureExtraction`. 
They have different interfaces.

```python
from sentence_transformers import SentenceTransformer

# Download from the 🤗 Hub
peft_model = SentenceTransformer("sentence-transformers-testing/stsb-bert-tiny-lora")
print(peft_model.has_peft_compatible_model())
# ValueError: PEFT methods are only supported for Sentence Transformer models that use the Transformer module.
```

- Tom Aarsen

tomaarsen:
Fixed via #3405. Also see #3957 and #3975 for more details.

- Tom Aarsen

PR #3339, opened 2025-04-21
Add merge_adapter method to PeftAdapterMixin to safely merge PEFT adapters (#3247)

MrAnayDongre:
This commit introduces a `merge_adapter` method that checks for a callable `merge_and_unload` on the underlying model. If found, it merges the adapter and resets the configuration flag; otherwise, it raises a clear error. Tests cover both success and failure cases.


tomaarsen:
Hello!

Apologies for the huge delay. I never made progress with this, because I don't believe that the `transformers` models loaded with `AutoModel` and stored under `self[0].auto_model` ever have the `merge_and_unload` method. As a result, this method will always throw an error.

- Tom Aarsen

tomaarsen (changes requested):
See the above, I believe the `merge_and_unload` often (always?) doesn't exist.

PR #3342, opened 2025-04-22
[`ci`] Attempt to avoid 429 Client Error in CI

tomaarsen:
Hello!

## Pull Request overview
* Attempt to avoid 429 Client Error in CI

## Details
This PR aims to fix the CI by reducing the number of requests made to Hugging Face via the CI.

- Tom Aarsen


tomaarsen:
I'm wary that it'll fill up too quickly, but I'll definitely give this a try!

PR #3349, opened 2025-04-30
feat: expose new model2vec parameters

stephantul:
Hello!

We're releasing a new model2vec version tomorrow. It has some shiny new parameters. This PR updates the spec of the `from_distillation` function to include the new parameters.

One issue is that this means that the `from_distillation` function is no longer backwards compatible with any model2vec version. So for example, if someone uses `model2vec 0.4.1` with the code in this PR, this will no longer work, and they will have to upgrade. One solution could be to actually pin model2vec in the sentence-transformers dependencies (in an extra).

Let me know what you think.

### DIFF
--- a/sentence_transformers/models/StaticEmbedding.py
+++ b/sentence_transformers/models/StaticEmbedding.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import math
 import os
 from pathlib import Path
+from typing import Any
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


tomaarsen:
Heya! This looks awesome. Could you perhaps also import `__version__ as model2vec_version` and then check `if version.parse(model2vec_version) < version.parse("..."):` then throw an error to update `model2vec`.

After all, older versions will break because of this. 

Or perhaps even better, support both the new and older versions of model2vec.

- Tom Aarsen

PR #3360, opened 2025-05-12
Reload all modules when loading the best saved checkpoint, not just the transformer one

### LINKED ISSUE #3344: _load_best_model() only restores auto_model, not the full SentenceTransformer architecture
As discussed, opening this issue to document the current behavior of _load_best_model() in the training script. Right now, it only reloads the underlying auto_model component and keep  the other components of  the full SentenceTransformer model (which may include pooling, dense layers, etc.)  as they were at the end.

This means that after early stopping or selecting the best checkpoint, the final model used for evaluation might differ from the actual best-performing full model, potentially leading to slightly degraded results.

Here's the current logic:
```
def _load_best_model(self) -> None:
        # We want to ensure that this does not fail, and it may change if transformers updates how checkpoints are saved
        # Loading the best model is only supported for `transformers`-based models
        if not isinstance(self.model[0], Transformer):
            logger.info("Could not load best model, as the model is not a `transformers`-based model.")
            return

        try:
            if checkpoint := self.state.best_model_checkpoint:
                step = checkpoint.rsplit("-", 1)[-1]
                self.model.model_card_data.set_best_model_step(int(step))
        except Exception:
            pass

        # Override the model with the `transformers`-based auto_model, and restore the original SentenceTransformers
        # model with the loaded `transformers` model
        full_model = self.model
        self.model = self.model[0].auto_model
        try:
            return super()._load_best_model()
        finally:
            loaded_auto_model = self.model
            self.model = full_model
            self.model[0].auto_model = loaded_auto_model
```

Opening this here as a reference point.

tomaarsen:
Resolves #3344

Hello!

## Pull Request overview
* Reload all modules when loading the best saved checkpoint, not just the transformer one

## Details
I'm now using a dummy model that I load using the checkpoint path, and then I copy the state dict into the original model that we were training with. This is done instead of just overriding the `self.model` with a fresh model, as this might also override user arguments  like `prompts`, `truncate_dim`, etc.

I suspect that this may have some negative consequences as well, e.g. with FSDP, distributed training, or PEFT, considering the `transformers` `_load_best_model` that I now completely override has a lot more contents. 

cc @arthurbr11 @NohTow

- Tom Aarsen

### DIFF
--- a/sentence_transformers/trainer.py
+++ b/sentence_transformers/trainer.py
@@ -23,7 +23,6 @@ from sentence_transformers.evaluation import SentenceEvaluator, SequentialEvalua
 from sentence_transformers.losses.CoSENTLoss import CoSENTLoss
 from sentence_transformers.model_card import SentenceTransformerModelCardCallback
 from sentence_transformers.models import Pooling
-from sentence_transformers.models.Transformer import Transformer
 from sentence_transformers.sampler import (
     DefaultBatchSampler,
     GroupByLabelBatchSampler,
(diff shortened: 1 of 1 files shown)


arthurbr11:
I'm not sure but shouldn't the `CrossEncoder.trainer` function should be updated too as it seems to only load the `auto_model` too  [here](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/cross_encoder/trainer.py#L316-L333)

tomaarsen:
Well spotted. I was thinking about that one too, but I think it's actually fine if the CrossEncoder uses the `_load_best_model` from `transformers`, as CrossEncoder models don't have weights outside of the `transformers` model (whereas SentenceTransformer models can have weights in e.g. a `Dense`).

And the `transformers` `_load_best_model` has some protections for distributed training etc. that my proposal here doesn't have, so I think I prefer using the original `transformers` `_load_best_model` for CrossEncoder models.

- Tom Aarsen

arthurbr11:
Make sense, wasn't aware that `CrossEncoder` don't have weights outside of the `transformer` model.

PR #3362, opened 2025-05-12
HF tokenizer support for word embeddings

talbaumel:
WordEmbeddings class expected WordTokenizer input, added suuport for hf tokenizer as well

### DIFF
--- a/sentence_transformers/models/WordEmbeddings.py
+++ b/sentence_transformers/models/WordEmbeddings.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
+from typing import Union
 
+from transformers import PreTrainedTokenizerFast, PreTrainedTokenizer
 import gzip
 import logging
 import os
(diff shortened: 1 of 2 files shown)


talbaumel:
Hi Tom, thanks for the feedback! I think everything is resoved

tomaarsen:
I believe there's still some issues - I unresolved them. For example, you're still initializing `WordTokenizer.WordTokenizerWrapper` while you did rename the class to `TransformersTokenizerWrapper`.

talbaumel:
I see my last push wasn't push
Sorry!

tomaarsen (commented):
All good! I had a suspicion about that already, but I wasn't sure 😄

One more thing: I think the best class to check for is `PreTrainedTokenizerBase` instead of `PreTrainedTokenizer`. Both `PreTrainedTokenizer` and `PreTrainedTokenizerFast` subclass `PreTrainedTokenizerBase`, so checking `isinstance(tokenizer, PreTrainedTokenizerBase)` should work with either tokenizer type.

talbaumel:
NP, done :)

Issue #3357, opened 2025-05-09
Add a utility to easily find the underlying `PreTrainedModel` subclass


tomaarsen:
Hello!

## Details
It's common to want to inspect or update the underlying `PreTrainedModel` subclass, so let's add a utility to easily find it, and return `None` if it does not exist.

- Tom Aarsen

St4r4x:
Looks like this has already been addressed — `model.transformers_model` was added in commit 963b352c ("Add transformers_model property for easier access", merged 2025-06-27), matching exactly what's described here:

```python
@property
def transformers_model(self) -> PreTrainedModel | None:
    """
    Property to get the underlying transformers PreTrainedModel instance, if it exists.
    """
    for module in self.modules():
        if isinstance(module, Transformer):
            return module.model
        if isinstance(module, PreTrainedModel):
            return module
    return None
```

It's on `main` in `sentence_transformers/base/model.py` and already used internally (e.g. `model.config` delegates to it, ONNX/OpenVINO optimization code uses it too). Might be worth closing this one out?

tomaarsen:
Agreed! This has been added. Thanks for the reminder that this hasn't been closed yet.

- Tom Aarsen

PR #3365, opened 2025-05-18
add a utility function to get the underlying PreTrainedModel subclass (#3357)

CKeibel:
This PR implements a utility function to easily find and return the underlying PreTrainedModel subclass within a SentenceTransformer model, as requested in issue #3357. The utility returns None if no such model is found.

If this approach is considered suitable, I will be happy to add corresponding tests to ensure robustness and maintainability. 😊

### DIFF
--- a/sentence_transformers/util.py
+++ b/sentence_transformers/util.py
@@ -1647,3 +1647,42 @@ def disable_datasets_caching():
     finally:
         if is_originally_enabled:
             enable_caching()
+
+
+def find_pretrained_model_class(model):
+    """
+    Find the class of a pretrained model in a SentenceTransformer model.
+
+    Args:
+        model: The model to search for the pretrained model class.
+
+    Returns:
+        The class of the pretrained model if found, otherwise None.
+    """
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Thank you for this work. I think I'm going to go one step further with it: adding a property on the models themselves. I've incorporated it in https://github.com/arthurbr11/sentence-transformers/pull/44, which will build on top of the v5.0 release PR.  Usage will be like:
```python

model = SentenceTransformer("all-mpnet-base-v2")

# You can now access the underlying transformers model
transformers_model = model.transformers_model
print(type(transformers_model))
# => <class 'transformers.models.mpnet.modeling_mpnet.MPNetModel'>
```

What do you think?

- Tom Aarsen

CKeibel:
Hi! 

That looks like a much better and more logical solution to it. I like it a lot more and it is more intuitive to use it that way :)

tomaarsen:
Then I'll be closing this in favor of that one. I'll work to get it released soon!

- Tom Aarsen

PR #3369, opened 2025-05-22
Add OpenSearch-based semantic search example

zhichao-aws:
Add code examples to do semantic search using OpenSearch engine. 

As dataset, we use the Natural Questions dataset: https://ai.google.com/research/NaturalQuestions/

The answers are indexed to OpenSearch together with their respective sentence embeddings. And we show the comparison between semantic search and BM25 search.


tomaarsen (approved):
I got this working locally, works well. Thank you for this! I have no comments on the code, looks solid.

- Tom Aarsen

PR #3373, opened 2025-05-23
fix: device name in multi-node ddp

sasakiyori:
`torch.distributed.get_rank()` returns the global rank in a DDP training task. It should not be directly mapped to a CUDA device number.
In scenarios such as multi-node setups or k8s with the [NVIDIA device plugin](https://github.com/NVIDIA/k8s-device-plugin), the number of available devices on a node may be smaller than the global rank.
To avoid assigning an invalid device ordinal, it should prioritize `LOCAL_RANK` first, and then compare the number of available GPUs with the global rank.

### DIFF
--- a/sentence_transformers/util.py
+++ b/sentence_transformers/util.py
@@ -1581,10 +1581,12 @@ def get_device_name() -> str:
         str: Device name, like 'cuda:2', 'mps', 'npu', 'hpu', or 'cpu'
     """
     if torch.cuda.is_available():
-        if torch.distributed.is_initialized():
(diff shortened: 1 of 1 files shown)


tomaarsen:
Made some changes to ensure that LOCAL_RANK actually exists to avoid `os.environ.get("LOCAL_RANK")` from failing. Let me know what you think.

- Tom Aarsen

sasakiyori:
LGTM. Thanks for the review :)

tomaarsen (approved):
Seems good to me!

PR #3376, opened 2025-05-27
Minor change to docs, ordering of operations was off by one.

reptillicus:
Off by one on the order of operations.

### DIFF
--- a/docs/sentence_transformer/training_overview.md
+++ b/docs/sentence_transformer/training_overview.md
@@ -435,16 +435,16 @@ The :class:`~sentence_transformers.trainer.SentenceTransformerTrainer` is where 
         )
     )
 
-    # 3. Load a dataset to finetune on
+    # 2. Load a dataset to finetune on
     dataset = load_dataset("sentence-transformers/all-nli", "triplet")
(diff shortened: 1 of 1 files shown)


johneckberg (commented):
The order of operations is correct, there is a [step 2](https://github.com/UKPLab/sentence-transformers/blob/163258f16cc0fa75639fd6e2b58bdc42921178f0/docs/sentence_transformer/training_overview.md?plain=1#L428) listed on line 428 (specifying an `SentenceTransformerModelCardData` object). However, it's inline with step 1, as it's an optional step when initializing the model.

tomaarsen:
Hello!

@johneckberg is right - there is a `2.`, but it's admittedly a bit confusing. 
I think I would like to keep a separate number for the model card (to really show that it's a separate component that's totally optional) and to align it with the numbering in the Documentation section to the side:
image

And I'm also wary to put this on a separate line, the entire script is already pretty long.

In the end, I think it's a matter of choosing between a few evils, and I think keeping it as-is is the best choice for now. So, I'll be closing this, but feel free to let me know if you disagree. 

- Tom Aarsen

Issue #3377, opened 2025-05-28
Mistake in TripletEvaluator documentation


AlexiaALLAL:
In the TripletEvaluator implementation, documentation and the [docstring](https://github.com/UKPLab/sentence-transformers/blob/163258f16cc0fa75639fd6e2b58bdc42921178f0/sentence_transformers/evaluation/TripletEvaluator.py#L27-L28) states:

> Evaluate a model based on a triplet: (sentence, positive_example, negative_example).
Checks if ``similarity(sentence, positive_example) < similarity(sentence, negative_example) + margin``.

However, the actual [code](https://github.com/UKPLab/sentence-transformers/blob/163258f16cc0fa75639fd6e2b58bdc42921178f0/sentence_transformers/evaluation/TripletEvaluator.py#L227) does: 
`accuracy = (positive_scores > negative_scores + self.margin[fn_name]).float().mean().item()`

So I think the doc should be ``similarity(sentence, positive_example) > similarity(sentence, negative_example) + margin`` ?

johneckberg:
Hi @AlexiaALLAL! 
Thank you for noticing this, it appears that this happened when TripletEvaluator was changed to use [similarities instead of distances.](https://github.com/UKPLab/sentence-transformers/commit/b055b5d502bbe297131f13ad04503ea4c92c5abd).

tomaarsen:
Thank you for spotting this and reporting it for me! I've fixed it via @johneckberg 's PR in #3379.

- Tom Aarsen

PR #3379, opened 2025-05-29
Correcting TripletEvaluator.py Docstring

johneckberg:
## Pull Request overview

- Fixes an error in the docstring for the TripletEvaluator class.

## Details:

In issue #3377, @alexiaALLAL noted that in the [TripletEvaluator class](https://github.com/UKPLab/sentence-transformers/blob/163258f16cc0fa75639fd6e2b58bdc42921178f0/sentence_transformers/evaluation/TripletEvaluator.py), the docstring states:
"Checks if similarity(sentence, positive_example) < similarity(sentence, negative_example) + margin."

However, the implementation performs the following:
`accuracy = (positive_scores > negative_scores + self.margin[fn_name])`

It looks like what happened was when TripletEvaluator was changed to use [similarities instead of distances](https://github.com/UKPLab/sentence-transformers/commit/b055b5d502bbe297131f13ad04503ea4c92c5abd), the docstring was changed to replace distance() with similarity(), without changing the less than sign to a greater than sign.

### DIFF
--- a/sentence_transformers/evaluation/TripletEvaluator.py
+++ b/sentence_transformers/evaluation/TripletEvaluator.py
@@ -25,7 +25,7 @@ logger = logging.getLogger(__name__)
 class TripletEvaluator(SentenceEvaluator):
     """
     Evaluate a model based on a triplet: (sentence, positive_example, negative_example).
-    Checks if ``similarity(sentence, positive_example) < similarity(sentence, negative_example) + margin``.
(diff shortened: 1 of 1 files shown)


tomaarsen:
Thank you both, @AlexiaALLAL and @johneckberg, for spotting this and for setting up this PR to resolve it. You're very right indeed, I mistakenly didn't update the operator when we refactored the evaluator. This'll be included in the upcoming release's documentation.

- Tom Aarsen

PR #3385, opened 2025-06-08
Move precision format validation before training loop

mvairav:
Previously, the code checked if the precision was among the supported options only after the embeddings were computed. If an unsupported format was passed, it would raise a ValueError late in the process, potentially losing all prior work. This commit ensures that precision is validated as a string and among the supported options before any embedding computation begins. This also prevents AttributeError from the .endswith call in quantize_embedding if precision is not a string.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -615,6 +615,10 @@ class SentenceTransformer(nn.Sequential, FitMixin, PeftAdapterMixin):
             sentences = [sentences]
             input_was_string = True
 
+        allowed_precisions = {"float32", "int8", "uint8", "binary", "ubinary"}
+        if precision is None or not isinstance(precision, str) or precision not in allowed_precisions:
+            raise ValueError(f"Precision {precision} is not supported")
+
         if prompt is None:
             if prompt_name is not None:
                 try:


sentence_transformers/SentenceTransformer.py
```diff
@@ -615,6 +615,10 @@ def encode(
             sentences = [sentences]
             input_was_string = True
 
+        allowed_precisions = {"float32", "int8", "uint8", "binary", "ubinary"}
+        if precision is None or not isinstance(precision, str) or precision not in allowed_precisions:
+            raise ValueError(f"Precision {precision} is not supported")
```
tomaarsen:
```suggestion
            raise ValueError(f"Precision {precision!r} is not supported")
```
Let's print a repr, so we get `Precision 'fp8' is not supported` instead of `Precision fp8 is not supported`.
I'll fix this, and also in the `quantize_embeddings` function.

tomaarsen:
Hello!

I think this is very fair, it's really frustrating if this pops up right after you've embedded everything. Your fix here is quite clean, so I have no issues introducing this.

- Tom Aarsen

tomaarsen:
I also updated the if-statement to check for `precision` truthy, e.g. don't give errors for `precision=None` or `precision=""` - these have always just not updated the precision, so we don't need to introduce a ValueError for them now.

Please let me know what you think! I feel like we're about ready to merge this.,

- Tom Aarsen

mvairav:
Yes, that looks good. Thank you!

PR #3419, opened 2025-07-02
Update README.md

CharlesCNorton:
- Remove redundant article "a" before "numpy arrays"

- Add missing "for" in comment "predict scores for pairs of texts"

- Pluralize "sparse embedding models" for consistency with other list items


tomaarsen:
Hello!

Thank you so much for finding these, this is very valuable 🤗

- Tom Aarsen

PR #3421, opened 2025-07-03
Adjust label shape handling in MarginMSELoss for single score inputs

tomaarsen:
Hello!

## Pull Request overview
* Adjust label shape handling in MarginMSELoss for single score inputs

## Details
Some recent changes to allow for separate scores for query-positive and query-negative in MarginMSELoss (instead of just query-positive minus query-negative scores) introduced an issue preventing single-score inputs. This resolves that.

In the meantime, please use a list of scores, even if you just have 1 score (e.g. in the triplet case).

cc @arthurbr11 

- Tom Aarsen


tomaarsen:
Test failures related to an outage on one of the datasets, not related to this PR.

- Tom Aarsen

PR #3178, opened 2025-01-17
[`fix`] Use HfArgumentParser-compatible typing for prompts

tomaarsen:
Resolves #3090

Hello!

## Pull Request overview
* Use HfArgumentParser-compatible typing for prompts

## Details
Resolves #3090 by updating the typing from `dict[str, dict[str, str]] | dict[str, str] | str | None` to just `str | None`. I've also added the explanation in the `metadata` to make sure people are aware that there's more valid options.

And I added a simple test case that just ensures there's no crash. This should ensure that this same issue won't happen anymore.

- Tom Aarsen

PR #3427, opened 2025-07-05
Sync `prompts` type with documentation

FremyCompany:
`Union[Dict[str, Dict[str, str]], Dict[str, str], str]` instead of `str`

(no idea if this compiles, just made this change while looking at the github)

### DIFF
--- a/sentence_transformers/training_args.py
+++ b/sentence_transformers/training_args.py
@@ -193,7 +193,7 @@ class SentenceTransformerTrainingArguments(TransformersTrainingArguments):
             with different learning rates.
     """
 
-    prompts: Optional[str] = field(  # noqa: UP007
+    prompts: Optional[Union[Dict[str, Dict[str, str]], Dict[str, str], str]] = field(  # noqa: UP007
         default=None,
         metadata={
             "help": "The prompts to use for each column in the datasets. "


FremyCompany:
The issue was introduced 6 months ago here: @tomaarsen
https://github.com/UKPLab/sentence-transformers/commit/c68bf68299a4435c6a48ea15d789fef596bf1444

tomaarsen:
Hello!

I'm afraid I'll have to keep it as just `str`. The reasoning is described in #3178, but in short: the HF Argument Parser won't recognize complex typings like that. Because of it, some semi-standard argument parsing code would break.

So, I've had to set the typing to just `str` despite more options being possible (and recommended! The same prompt for everything isn't useful - prompts are for allowing the model to distinguish different types of inputs).

- Tom Aarsen

FremyCompany:
Wouldn't it make more sense to just use `any`, then? The reason I came here is that type checking in vscode flags my prompts as invalid type since I upgraded `sentence_transformers`.

image

Of course I can safely ignore but it's annoying.

tomaarsen:
That is indeed frustrating, I'll try and work something out that works both in IDEs and via the HFArgumentParser.

- Tom Aarsen

Samoed:
I think you can try to define typehint with `TYPE_CHEKING` to overcome this issue
```python
if TYPE_CHEKING:
    PromptType = dict[str, dict[str, str]] | dict[str, str] | str  # will work with IDE
else:
    PromptType = str  # will work in runtime
```

tomaarsen:
That sounds like a very solid suggestion, thanks @Samoed! Will experiment with it.

- Tom Aarsen

tomaarsen:
Hello!

I dove a bit further into the `transformers` code. Initially I tried @Samoed's suggestion, but I realized I could get a simpler fix by updating the `_VALID_DICT_FIELDS` in the Training Arguments, which allows `str` inputs. Then, also allow `str` as a type (in the 2nd position) and the CLI HfArgumentParser will also accept it.
I also updated the tests slightly.

- Tom Aarsen

tomaarsen:
Failing test is unrelated, we should be good to go with this. @FremyCompany does this satisfy your IDE requirements?

- Tom Aarsen

tomaarsen (approved):

PR #3442, opened 2025-07-11
[`feat`] Add gather_across_devices parameter to some contrastive losses

tomaarsen:
Hello!

## Pull Request overview
* Add `gather_across_devices` to MNRL and CMNRL, etc.
* Add `contrast_anchors` and `contrast_positives` to GIST and CGIST losses
* Add utility functions for distributed all-gathers

## Details
This is a work-in-progress. The goal of this PR is to introduce gathering across devices, as @NohTow showed that it was still worthwhile despite the same also being possible with GradCache. Also thanks to @NohTow for writing the code that this PR is based on & for helping debugging.

## TODOs:
- [x] Update all other contrastive losses that benefit from in-batch negatives.
	- [x] For Dense
    	- [x] Implement and test for MNRL
    	- [x] Implement and test for CMNRL
    	- [x] Implement and test for MNSRL
    	- [x] Implement and test for CMNSRL
    	- [x] Implement and test for GIST
    	- [x] Implement and test for CGIST
	- [x] ~~For Reranker~~: Rerankers don't benefit from cross-device gathering as they require computing per pair.
	- [x] For Sparse
    	- [x] Implement for SparseMNRL
- [ ] Turn `util.py` into a `util` folder with files for separate components, e.g. `negative_mining.py`, `distributed.py`, `similarity_functions.py`, `huggingface.py`, etc. It's getting too big.

---

- Tom Aarsen

### DIFF
--- a/sentence_transformers/losses/CachedGISTEmbedLoss.py
+++ b/sentence_transformers/losses/CachedGISTEmbedLoss.py
@@ -13,6 +13,7 @@ from transformers import PreTrainedTokenizerBase
 
 from sentence_transformers import SentenceTransformer
 from sentence_transformers.models import StaticEmbedding
+from sentence_transformers.util import all_gather_with_grad
 
 
 class RandContext:
@@ -71,6 +72,9 @@ class CachedGISTEmbedLoss(nn.Module):
         show_progress_bar: bool = False,
         margin_strategy: Literal["absolute", "relative"] = "absolute",
         margin: float = 0.0,
+        contrast_anchors: bool = True,
+        contrast_positives: bool = True,
+        gather_across_devices: bool = False,
     ) -> None:
         """
         This loss is a combination of :class:`GISTEmbedLoss` and :class:`CachedMultipleNegativesRankingLoss`.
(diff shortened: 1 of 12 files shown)


tomaarsen:
I've decided to leave this item for a follow-up PR to avoid expanding the scope of this PR too much:
* Turn `util.py` into a `util` folder with files for separate components, e.g. `negative_mining.py`, `distributed.py`, `similarity_functions.py`, `huggingface.py`, etc. It's getting too big.

With all of these losses tested (at least, checked that they perform better with gathering, ran without crashes, etc.), I'm ready to merge this. The test failures are related to the datasets v4 release, and the issues were fixed in #3445. I'll merge those into this branch first.

- Tom Aarsen

Batwu:
Hi, thank you for the great work on this project.

While reviewing the implementation of `MultipleNegativesRankingLoss` in a distributed setting, I noticed a potential issue with how the `labels` are computed when using `all_gather` on the candidates.

In particular, the code computes labels as:

```python
if torch.distributed.is_initialized():
    rank = torch.distributed.get_rank()
    offset = rank * batch_size
labels = torch.arange(offset, offset + batch_size, device=anchors.device)
```

However, if candidates are constructed using `torch.cat(embeddings[1:])` where `[positives, negatives]` are concatenated along `dim=0`, the actual positions of the positives in the gathered candidates tensor will not necessarily align with `offset + i`. This is because torch.cat first places all positives, then all negatives from each rank, so the true index of the positive may differ from what `offset = rank * batch_size` assumes.
I'm still new to distributed training and PyTorch, so please forgive me if I misunderstood the logic！

NohTow:
Not sure I get you correctly, but essentially since we are only gathering the queries, this offset is meant to align those with their corresponding positive documents, which as you pointed out are always the first n documents (because we concatenate the positive and then the negatives).
Among those n documents, the positives for a query are at `[rank * batch_size: rank* batch_size + 1]` (considering we concatenate following the rank order). It is just meant to skip the first positives that corresponds to the positives of previous rank.

tomaarsen:
I think @Batwu might be on to something though, although I'm not 100% sure yet. If we have e.g. 1 positive and 8 negatives with 4 devices, then the `candidates` that are gathered are structured a bit like `PNNNNNNNNPNNNNNNNNPNNNNNNNNPNNNNNNNN`.

But the offsetting with `labels` only offsets by `batch_size`, resulting in:
```
PNNNNNNNNPNNNNNNNNPNNNNNNNNPNNNNNNNN
1
 2
  3
   4
```
instead of offsetting by `batch_size * (1 + num_negatives)`, resulting in:
```
PNNNNNNNNPNNNNNNNNPNNNNNNNNPNNNNNNNN
1
         2
                  3
                           4
```

I'll have to do some more tests, but I think it's likely that the tests I ran only used an anchor and positive, and no negatives. Otherwise I would have noticed a degradation in performance presumably.

- Tom Aarsen

NohTow:
That is actually totally true, I am very sorry I had the PyLate implementation in mind, where you can see that [we do not concatenate the positives/negatives _before_ gathering the elements](https://github.com/lightonai/pylate/blob/84b1a0c111d28b25f2e00d04782364abc2dbf138/pylate/losses/contrastive.py#L165) (the "concatenation" is done at the score level), which in this case create PPPNNNNNNNN.
I believe you could simply move the concat operation to _after_ the gathering to fix the issue no?

Again, deeply sorry @Batwu, should have checked the code of Tom instead of answering based on the code used as base.

PR #3455, opened 2025-07-22
[`fix`] cast indexing numpy int to Python int

emapco:
In datasets==4.0.0, arrow datasets raise a TypeError (`TypeError: Wrong key type: '5' of type '<class 'numpy.int64'>'. Expected one of int, slice, range, str or Iterable.`) when indexed with a numpy int.

This PR casts the index before indexing `sentences`.

[Traceback](https://github.com/emapco/chem-mrl/actions/runs/16437195797/job/46449425520?pr=5#step:5:187)

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -1013,7 +1013,7 @@ class SentenceTransformer(nn.Sequential, FitMixin, PeftAdapterMixin):
 
         all_embeddings = []
         length_sorted_idx = np.argsort([-self._text_length(sen) for sen in sentences])
-        sentences_sorted = [sentences[idx] for idx in length_sorted_idx]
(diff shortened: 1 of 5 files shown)


tomaarsen:
Hello!

I'm able to reproduce this, thank you for opening. I think the core of the issue is that `dataset["column_name"]` doesn't return a list of strings anymore. This has consequences all over the project. Normally I'd prefer e.g. a PR that updates code to make sure that ST.encode is called with a list of strings, but because the `model.encode(dataset["column_name"])` approach is so prevalent from before datasets v4.0, I think the best approach is indeed this.

- Tom Aarsen

tomaarsen:
The test failures are unrelated - they're caused by "Too many requests" timeouts to Hugging Face. I still have to figure out how to address those.
Thank you for this PR, I think it's ready to be merged. It'll be usable with:
```
```
and fully released in the next version.

- Tom Aarsen

PR #3456, opened 2025-07-22
Add  FLOPS calculation and update metrics in SparseEvaluators

arthurbr11:
Hello!

## Pull Request overview
* This pull request introduces enhancements to evaluation metrics, improves sparsity evaluation. The most significant changes include adding Average FLOPS as a metric and refactoring sparsity statistics handling.

This addition is here to be able to have fair comparison with other splade models that usually report this metric. I'm waiting for opensearch result to make sure that we get same result with the computation of the score.
 
cc @tomaarsen 

* Arthur BRESNU

### DIFF
--- a/sentence_transformers/sparse_encoder/evaluation/SparseInformationRetrievalEvaluator.py
+++ b/sentence_transformers/sparse_encoder/evaluation/SparseInformationRetrievalEvaluator.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING, Any, Callable
 import torch
 
 from sentence_transformers.evaluation import InformationRetrievalEvaluator
-from sentence_transformers.util import append_to_last_row
+from sentence_transformers.util import append_to_last_row, compute_count_vector
 
 if TYPE_CHECKING:
     import numpy as np
@@ -78,11 +78,6 @@ class SparseInformationRetrievalEvaluator(InformationRetrievalEvaluator):
             # For this dataset, we want to concatenate the title and texts for the corpus
             corpus = corpus.map(lambda x: {"text": x["title"] + " " + x["text"]}, remove_columns=["title"])
 
-            # Shrink the corpus size heavily to only the relevant documents + 1,000 random documents
-            required_corpus_ids = set(map(str, relevant_docs_data["corpus-id"]))
-            required_corpus_ids |= set(random.sample(corpus["_id"], k=1000))
-            corpus = corpus.filter(lambda x: x["_id"] in required_corpus_ids)
-
             # Convert the datasets to dictionaries
             corpus = dict(zip(corpus["_id"], corpus["text"]))  # Our corpus (cid => document)
             queries = dict(zip(queries["_id"], queries["text"]))  # Our queries (qid => question)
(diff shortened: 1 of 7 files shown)


sentence_transformers/sparse_encoder/evaluation/SparseInformationRetrievalEvaluator.py
```diff
@@ -106,34 +101,35 @@ class SparseInformationRetrievalEvaluator(InformationRetrievalEvaluator):
 
             # Run evaluation
             results = ir_evaluator(model)
-            '''
+            ```
```
tomaarsen:
```suggestion
            '''
```
Probably best to keep this as a comment in case users want to copy it

sentence_transformers/sparse_encoder/evaluation/SparseInformationRetrievalEvaluator.py
```diff
@@ -106,34 +101,35 @@ class SparseInformationRetrievalEvaluator(InformationRetrievalEvaluator):
...
+            Recall@5: 13.47%
+            Recall@10: 17.01%
+            MRR@10: 0.5706
+            NDCG@10: 0.3530
+            MAP@100: 0.1778
             Model Query Sparsity: Active Dimensions: 40.0, Sparsity Ratio: 0.9987
             Model Corpus Sparsity: Active Dimensions: 206.2, Sparsity Ratio: 0.9932
-            '''
+            Average FLOPS: 4.7
+            ```
```
tomaarsen:
```suggestion
            '''
```

sentence_transformers/sparse_encoder/evaluation/SparseInformationRetrievalEvaluator.py
```diff
@@ -106,34 +101,35 @@ class SparseInformationRetrievalEvaluator(InformationRetrievalEvaluator):
...
             # Print the results
             print(f"Primary metric: {ir_evaluator.primary_metric}")
             # => Primary metric: BeIR-nfcorpus-subset-test_dot_ndcg@10
             print(f"Primary metric value: {results[ir_evaluator.primary_metric]:.4f}")
-            # => Primary metric value: 0.3621
+            # => Primary metric value: 0.Primary metric value: 0.3530
```
tomaarsen:
```suggestion
            # => Primary metric value: 0.3530
```
Copy-paste error I think

sentence_transformers/sparse_encoder/evaluation/SparseInformationRetrievalEvaluator.py
```diff
@@ -185,9 +185,16 @@ def __init__(
 
     def _append_csv_headers(self, similarity_fn_names):
         super()._append_csv_headers(similarity_fn_names)
-        self.csv_headers.extend(
-            ["query_active_dims", "query_sparsity_ratio", "corpus_active_dims", "corpus_sparsity_ratio", "avg_flops"]
-        )
+        if "@" in self.csv_headers[-1]:
```
tomaarsen: @arthurbr11 When is this False? 
Is the problem that it's called twice? Once in init and once if `self.score_functions is None`?

arthurbr11: when we are first initializing the evaluator seems that we doesn't init anything except epochs and steps (because by default `similarity_fn_names = []`, then we would append our custom metrics. 
Then when this function is re call in the evaluate we have the dot-Accuracy, ... added and then we re happen our custom metric and thats what break. 

So TL;DR: want to be sure we add then only if we already instantiate a metric

tomaarsen: Added a comment and updated the condition a bit in [`78a4503` (#3456)](https://github.com/UKPLab/sentence-transformers/pull/3456/commits/78a4503ac60c48430ae7d87b28f7c85def1bc23f)

Issue #3465, opened 2025-07-25
error handling for empty input in cross_encoder.predict


nitin-nsp:
Issue: cross_encoder.predict(arr) throws an error if arr is empty. Currently, there's no check for this.

Fix: Add a condition to handle empty input before calling .predict().

sahibpreetsingh12:
Hi @nitin-nsp is it still open and do you want me to contribute to this ?

tomaarsen:
There's an open PR that should resolve this in #3466, so it's being worked on already 🤗

- Tom Aarsen

PR #3466, opened 2025-07-25
chore: Handle error when predict is called with an empty sentence list

nitin-nsp:
#3465

### DIFF
--- a/sentence_transformers/cross_encoder/CrossEncoder.py
+++ b/sentence_transformers/cross_encoder/CrossEncoder.py
@@ -625,17 +625,23 @@ class CrossEncoder(nn.Module, PushToHubMixin, FitMixin):
                 # => array([0.6912767, 0.4303499], dtype=float32)
         """
         input_was_singular = False
-        if sentences and isinstance(sentences[0], str):  # Cast an individual pair to a list with length 1
+        if isinstance(sentences, tuple) and sentences and isinstance(sentences[0], str):
+            sentences = [sentences]
+            input_was_singular = True
+        elif isinstance(sentences, list) and sentences and isinstance(sentences[0], str):
             sentences = [sentences]
             input_was_singular = True
 
-        if len(sentences) == 0:
+        def _empty_result():
             if convert_to_tensor:
                 return torch.tensor([], device=self.model.device)
-            elif convert_to_numpy:
+            if convert_to_numpy:
                 return np.array([])
-            else:
-                return []
+            return []
+
+        # [] or [[]] or [(), (), ...] that is empty
+        if not sentences or all(not pair for pair in sentences):
+            return _empty_result()
 
         if show_progress_bar is None:
             show_progress_bar = (
--- a/sentence_transformers/cross_encoder/CrossEncoder.py
+++ b/sentence_transformers/cross_encoder/CrossEncoder.py
@@ -632,16 +632,13 @@ class CrossEncoder(nn.Module, PushToHubMixin, FitMixin):
             sentences = [sentences]
             input_was_singular = True
 
-        def _empty_result():
+        # [] or [[]] or [(), (), ...] that is empty
+        if not sentences or all(not pair for pair in sentences):
             if convert_to_tensor:
                 return torch.tensor([], device=self.model.device)
             if convert_to_numpy:
                 return np.array([])
             return []
-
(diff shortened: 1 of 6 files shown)


tomaarsen:
Hello!

Thank you for the PR. I'm considering taking it in a slightly different, direction, e.g. as proposed in the comment below. It will not result in an IndexError (like `master`) or a nice ValueError (like this PR), but instead just return an empty list. What do you think?

```python
from sentence_transformers import CrossEncoder, SentenceTransformer

model = CrossEncoder("cross-encoder/ms-marco-MiniLM-L6-v2")
output = model.predict([])
print(output)
# []
```

- Tom Aarsen

sentence_transformers/cross_encoder/CrossEncoder.py
```diff
@@ -624,6 +624,8 @@ def predict(
                 model.predict(sentences)
                 # => array([0.6912767, 0.4303499], dtype=float32)
         """
+        if not sentences:
+            raise ValueError("The 'sentences' parameter must not be empty. Please provide at least one sentence pair.")
         input_was_singular = False
         if isinstance(sentences[0], str):  # Cast an individual pair to a list with length 1
```
tomaarsen:
```suggestion
        input_was_singular = False
        if sentences and isinstance(sentences[0], str):  # Cast an individual pair to a list with length 1
```

nitin-nsp:
@tomaarsen Thanks for the suggestion! Updated to return empty containers (list/tensor/numpy) based on flags this keeps behavior consistent while avoiding errors. Appreciate the review!

tomaarsen:
Your changes indeed best reflect the other parameters (`convert_to_tensor`, `convert_to_numpy`). I'd like to see what the outputs for SentenceTransformer and SparseEncoder are when they get empty inputs, I'd like everything aligned here. Sorry for the delay.

- Tom Aarsen

nitin-nsp:
Thanks for the reply, @tomaarsen !

In the this commit,
- I’ve added test cases and patched the function to handle nested empty input errors gracefully.

- I’ve also attached the test results confirming correct behavior across different configurations (convert_to_tensor, convert_to_numpy, etc.).

Let me know if anything else needs alignment!

Thanks again ,
Nitin

```
pytest -s -q -k  test_predict_empty
input:  [] result:  tensor([], device='mps:0')
.input:  [] result:  tensor([], device='mps:0')
.input:  [[]] result:  tensor([], device='mps:0')
.input:  [[]] result:  tensor([], device='mps:0')
.input:  [(), ()] result:  tensor([], device='mps:0')
.input:  [(), ()] result:  tensor([], device='mps:0')
.input:  [] result:  tensor([], device='mps:0')
.input:  [] result:  tensor([], device='mps:0')
.input:  [[]] result:  tensor([], device='mps:0')
.input:  [[]] result:  tensor([], device='mps:0')
.input:  [(), ()] result:  tensor([], device='mps:0')
.input:  [(), ()] result:  tensor([], device='mps:0')
.input:  [] result:  []
.input:  [] result:  []
.input:  [[]] result:  []
.input:  [[]] result:  []
.input:  [(), ()] result:  []
.input:  [(), ()] result:  []
.input:  [] result:  []
.input:  [] result:  []
.input:  [[]] result:  []
.input:  [[]] result:  []
.input:  [(), ()] result:  []
.input:  [(), ()] result:  []
.
===================================================================== warnings summary =====================================================================

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
24 passed, 1 skipped, 1358 deselected, 1 warning in 47.11s```

tomaarsen:
Hello!

Apologies for the delay, I had some time to dive into this more. I made a handful of changes, as you can see in the commits.
In short, I had a few concerns:
1. We should extend this to SparseEncoder/SentenceTransformer, as it's quite a good idea to have good tests for this.
2. I sped up the tests by using a pytest fixture model.
3. I updated the CrossEncoder implementation so it doesn't return early. Having multiple return statements is risky, as perhaps I'll accidentally only update one of the two in the future, and we get a discrepancy. Beyond that, `[[]]` and `[(), ()]` have non-0 length, so then it might be unexpected to get an output of length 0.

I think it's very nice that this had also led to some improvements on SentenceTransformer/SparseEncoder. What do you think of my changes?

- Tom Aarsen

PR #3467, opened 2025-07-25
[`feat`] Update the trackio default project if not already defined

tomaarsen:
Hello!

## Pull Request overview
* Update the trackio default project if not already defined

## Details
Works identically to with W&B, nice to separate these runs from other HF-related runs that users might do.

- Tom Aarsen


tomaarsen:
Unrelated test failures due to Rate Limit issues that I've been having with huggingface.co

- Tom Aarsen

Issue #3472, opened 2025-07-28
SPLADE-Index Python Library: An ultra-fast search index for SPLADE sparse retrieval models


rasyosef:
`splade-index`: https://github.com/rasyosef/splade-index

# SPLADE-Index⚡

<i>
SPLADE-Index is an ultrafast index for SPLADE sparse retrieval models implemented in pure Python and powered by Scipy sparse matrices. It is built on top of the BM25s library.
</i>
<br/><br/>

SPLADE is a neural retrieval model which learns query/document sparse expansion. Sparse representations benefit from several advantages compared to dense approaches: efficient use of inverted index, explicit lexical match, interpretability... They also seem to be better at generalizing on out-of-domain data (BEIR benchmark).

For more information about SPLADE models, please refer to the following. 
 - [SPLADE: Sparse Lexical and Expansion Model for First Stage Ranking](https://arxiv.org/abs/2107.05720)
 - [List of Pretrained Sparse Encoder (Sparse Embeddings) Models](https://sbert.net/docs/sparse_encoder/pretrained_models.html)
 - [Training and Finetuning Sparse Embedding Models with Sentence Transformers v5](https://huggingface.co/blog/train-sparse-encoder).

## Installation

You can install `splade-index` with pip:

```bash
pip install splade-index
```

Recommended (but optional) dependencies:

```bash
# To speed up the top-k selection process, you can install `jax`
pip install "jax[cpu]"
```

## Quickstart

Here is a simple example of how to use `splade-index`:

```python
from sentence_transformers import SparseEncoder
from splade_index import SPLADE

# Download a SPLADE model from the 🤗 Hub
model = SparseEncoder("rasyosef/splade-tiny")

# Create your corpus here
corpus = [
    "a cat is a feline and likes to purr",
    "a dog is the human's best friend and loves to play",
    "a bird is a beautiful animal that can fly",
    "a fish is a creature that lives in water and swims",
]

# Create the SPLADE retriever and index the corpus
retriever = SPLADE()
retriever.index(model=model, documents=corpus)

# Query the corpus
queries = ["does the fish purr like a cat?"]

# Get top-k results as a tuple of (doc ids, documents, scores). All three are arrays of shape (n_queries, k).
results = retriever.retrieve(queries, k=2)
doc_ids, result_docs, scores = results.doc_ids, results.documents, results.scores

for i in range(doc_ids.shape[1]):
[...]

PR #3473, opened 2025-07-29
[`docs`] Add splade_index semantic search example

tomaarsen:
Hello!

## Pull Request overview
* Add semantic search example using [splade_index](https://github.com/rasyosef/splade-index) 

## Details
As a follow-up of #3472, this PR adds a documentation example of how to use the splade-index project for efficient retrieval with SPLADE. Unlike the other integrations in this documentation page, splade-index is so nicely integrated that there's no need for a `semantic_search_splade_index` function. Instead, the usage script is extremely straight-forward, and it's extremely efficient as well. I used 100k docs with naver/splade-v3 as an experiment, and my average latency for query inference + top 5 search only takes 0.05 seconds on my local GPU.

cc @xhluca @rasyosef

P.s. would it be interesting to upstream this into bm25s?

- Tom Aarsen

### DIFF
--- a/docs/sparse_encoder/pretrained_models.md
+++ b/docs/sparse_encoder/pretrained_models.md
@@ -55,8 +55,8 @@ Note that all the numbers of below are extracted information from different pape
 | [naver/splade-v3-lexical](https://huggingface.co/naver/splade-v3-lexical)                                                                                 | 40.0            | 49.1                | 109M       |
 | [prithivida/Splade_PP_en_v1](https://huggingface.co/prithivida/Splade_PP_en_v1)                                                                           | 37.2            | 48.7                | 109M       |
 | [naver/splade_v2_max](https://huggingface.co/naver/splade_v2_max)                                                                                         | 34.0            | 46.4                | 67M        |
-| [rasyosef/splade-mini](https://huggingface.co/rasyosef/splade-mini)                                                                                       | 32.8            | 42.9                | 11M        |
-| [rasyosef/splade-tiny](https://huggingface.co/rasyosef/splade-tiny)                                                                                       | 30.8            | 40.5                | 4M         |
+| [rasyosef/splade-mini](https://huggingface.co/rasyosef/splade-mini)                                                                                       | 33.2            | 42.5                | 11M        |
+| [rasyosef/splade-tiny](https://huggingface.co/rasyosef/splade-tiny)                                                                                       | 30.9            | 40.6                | 4M         |
 | BM25 (Baseline)                                                                                                                                           | 18.4            | 45.6                | NA         |
 
 ## Inference-Free SPLADE Models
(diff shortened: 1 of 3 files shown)


xhluca:
That's an interesting new addition to sentence-transformers! Regarding upstreaming to bm25s, I think it would result in (1) more dependencies for bm25s (and its users), and (2) lower visibility for the original repository, i.e. `splade-index` repo. I think it would make sense to keep them separate for now and maybe combine them into a single "retrieval" library eventually, e.g. https://pypi.org/project/retrieval/?

tomaarsen:
Very understandable. I'll defer to your judgement there. 

- Tom Aarsen

rasyosef:
Hi @tomaarsen, 

The tiny & mini splade models on huggingface were updated recently with improved versions and the MRR@10 scores on MSMARCO Dev have increased as well. Can you please update the new scores in your table? Sorry for the inconvenience.

### New Scores

The splade models were evaluated on 55 thousand queries and 8.84 million documents from the [MSMARCO](https://huggingface.co/datasets/microsoft/ms_marco) dataset.

||Size (# Params)|MRR@10 (MS MARCO dev)|
|:---|:----|:-------------------|
|`BM25`|-|18.0|-|-|
|`rasyosef/splade-tiny`|4.4M|30.9|
|`rasyosef/splade-mini`|11.2M|33.2|
|`naver/splade-v3-distilbert`|67.0M|38.7|

tomaarsen:
Sounds good, will update once I've recomputed scores for BEIR-13 (i.e. BEIR without CQADupstack and MSMARCO).

- Tom Aarsen

rasyosef:
Hi @tomaarsen,

Do you have a python script for computing BEIR-13 scores for sparse embedding models? If so, can you please share it with me? I had trained a `29M` param SPLADE model ([splade-small](https://huggingface.co/rasyosef/splade-small)), that achieved an **MRR@10 of 35.2** on MSMARCO, and I wanted to evaluate it on the BEIR benchmark as well.

I also had a question about the sparsity vs benchmark score tradeoff for sparse embedding models. Would it be worth it to sacrifice   1 point in MRR@10 for 40% fewer active dimensions, since this translates to faster search and lower memory usage when indexing and querying documents.

tomaarsen:
I used https://github.com/tomaarsen/mteb/blob/simple_splade_eval/demo_splade_os.py, from that branch of MTEB. I just pushed some changes there that I needed to easily evaluate SparseEncoder models with MTEB. See the full commit here: https://github.com/tomaarsen/mteb/commit/13cb9f2524ea5a4d091d3154bd282e4c5a8b5df6

You can install it with:
```bash
pip install git+https://github.com/tomaarsen/mteb@simple_splade_eval
```
and the run the `demo_splade_os.py` script. That should print out some `results` files.

> I also had a question about the sparsity vs benchmark score tradeoff for sparse embedding models. Would it be worth it to sacrifice 1 point in MRR@10 for 40% fewer active dimensions, since this translates to faster search and lower memory usage when indexing and querying documents.

This remains a tricky question, there's more tradeoffs with SparseEncoder models. Everyone has different preferences. I don't have great advice here. 

- Tom Aarsen

rasyosef:
Cool, Thanks!

Issue #3399, opened 2025-06-19
Feature Request: Surfacing Token Level Attention Scores inside the `forward` pass of `Transformer` class.


pramodith:
I've been playing around with the idea of using attention scores of tokens to come up with a new pooling mechanism for a given embedding model. Right now the `forward` pass of the `Transformer`'s class doesn't surface the token level attention scores, even if we send the kwarg `{"output_attentions": True}`.

https://github.com/UKPLab/sentence-transformers/blob/910ed144dfc0a08f31517b0d01580302015fa408/sentence_transformers/models/Transformer.py#L449-L451

I think it'd be helpful to have this optional feature returned if the user does want the attention scores. I'm happy to work on these changes and get a PR out.

whybe-choi:
Hi @pramodith, this is a great feature suggestion! I'm also very interested in being able to use attention scores for creating custom pooling mechanisms.

I saw you mentioned you'd be happy to create a PR for this. I was just wondering if you've had a chance to start working on it or if you still plan to?

No pressure at all, but if your priorities have shifted or you're short on time, I would be happy to pick this up and submit a PR. Please let me know what you think!

pramodith:
Hey @whybe-choi! I didn't get to this because none of the maintainers responded, I can get a PR out for this by Monday, since you've shown in an interest.

whybe-choi:
Thanks @pramodith ! Looking forward to your PR.

pramodith:
@whybe-choi have a PR out!

whybe-choi:
Awesome, thanks @pramodith! I'll check out the PR and give it a test run as soon as possible.

PR #3478, opened 2025-08-01
Allow the Transformer Module to Return Attention Scores

pramodith:
This PR addresses #3399.

Attention Scores can be useful for creating custom poolers that use the attention scores to weight and combine the embeddings of different tokens into a single document embedding.

In this PR we update the `Transformer` class to check if the `AutoModel` returns attention scores and returns them as a part of the output in the `features` dict.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -1119,7 +1119,7 @@ class SentenceTransformer(nn.Sequential, FitMixin, PeftAdapterMixin):
     def forward(self, input: dict[str, Tensor], **kwargs) -> dict[str, Tensor]:
         for module_name, module in self.named_children():
             module_kwargs = {}
-            if isinstance(module, Router):
+            if isinstance(module, Router) or isinstance(module, Transformer):
                 module_kwargs = kwargs
             else:
                 module_kwarg_keys = []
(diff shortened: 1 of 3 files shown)


sentence_transformers/SentenceTransformer.py
```diff
@@ -1119,7 +1119,7 @@ def encode(
     def forward(self, input: dict[str, Tensor], **kwargs) -> dict[str, Tensor]:
         for module_name, module in self.named_children():
             module_kwargs = {}
-            if isinstance(module, Router):
+            if isinstance(module, Router) or isinstance(module, Transformer):
```
pramodith, line 1122: I'm not entirely sure how safe this change is but it seemed like the easiest way to make sure that the SentenceTransformer class passes the `output_attentions` flag to the `Transformer` model. Open to suggestions if this can lead to bugs.

tomaarsen:
Hello!

I think this is very cool, but I'm not sure if implementing this directly in Sentence Transformers is the best approach. After all, it requires some custom pooling (subclass) as well before it can be used, then it's not that strange to also require a custom Transformer module (subclass).

Beyond that, for a custom Transformer module, you should be able to implement this more neatly with:
1. No changes in `SentenceTransformer.forward`.
2. Load the `SentenceTransformer` with `config_kwargs={"output_attentions": True}`.
3. Update `Transformer.forward` with something like `if "attentions" in outputs: features["attention_scores"] = outputs.attentions`.

Definitely keep me in the loop if you're able to develop some new pooling method with this that consistently does well. If we can update both the Transformer and the Pooling, then that would make more sense to add to Sentence Transformers directly, I think.

- Tom Aarsen

pramodith:
Gotcha! Thanks for the feedback Tom!. Closing this PR for now.

Issue #3462, opened 2025-07-23
Image Search example raises a mypy error


anakin87:
👋 This seems like a minor issue, but I'm flagging it in case anyone wants to take a look or address it.

```python
# image_search.py
# (example from https://sbert.net/examples/sentence_transformer/applications/image-search/README.html#usage)

from sentence_transformers import SentenceTransformer
from PIL import Image

# Load CLIP model
model = SentenceTransformer("clip-ViT-B-32")

# Encode an image:
img_emb = model.encode(Image.open("two_dogs_in_snow.jpg"))

# Encode text descriptions
text_emb = model.encode(
    ["Two dogs in the snow", "A cat on a table", "A picture of London at night"]
)

# Compute similarities
similarity_scores = model.similarity(img_emb, text_emb)
print(similarity_scores)
```

`mypy image_search.py` raises:
```
error: No overload variant of "encode" of "SentenceTransformer" matches argument type "ImageFile"  [call-overload]
note: Possible overload variants: ...
```

This likely happens because none of the overload variants accept images.

tomaarsen:
Hello!

Thank you for opening. This is indeed a missed case, and it's something I'd love to address properly when we heavily improve the library's support for different modalities. It's on the agenda, but we're investigating how we can move in that direction without backwards incompatibilities.

- Tom Aarsen

sahibpreetsingh12:
hi @tomaarsen 
I have read the issue and yes this can be a great addition to sentence transformers
If you want me I can get started?
and Yes what should be first few things on my target

sahibpreetsingh12:
Hi!  @tomaarsen  I would like to work on improving the type hint for `encode()` to support image inputs and fix the `mypy` error. 
If that's fine shall i start working on it

anakin87:
This was fixed in https://github.com/huggingface/sentence-transformers/releases/tag/v5.4.0

Thank you Tom!

PR #3485, opened 2025-08-08
Fix: Add Image.Image support to encode() overloads to resolve mypy error (#3462)

sahibpreetsingh12:
Hi @tomaarsen 

In this PR I have tried #3462 by updating the `encode()` method's type hints (including all relevant overload variants) to support `PIL.Image.Image` and `list[PIL.Image.Image]`.

Summary of Changes:
- Imported `Image` from `PIL`
- Updated all `encode()` overloads to include `Image.Image` and `list[Image.Image]`
- Also updated the main `encode()` function signature accordingly

This ensures that tools like `mypy` no longer raise type errors when calling `encode()` with image inputs. As it is working correctly at inference time — the issue was limited to type checking.

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -22,6 +22,7 @@ from typing import Any, Callable, Literal, overload
 
 import numpy as np
 import numpy.typing as npt
+from PIL import Image
 import torch
 import torch.multiprocessing as mp
 import transformers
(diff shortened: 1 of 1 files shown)


sahibpreetsingh12:
@tomaarsen how to make sure i pass next time I make submission

sahibpreetsingh12:
Hi @tomaarsen what can i do to move this PR further?

tomaarsen:
Hello @sahibpreetsingh12,

Apologies for the delay. As I mentioned in https://github.com/UKPLab/sentence-transformers/issues/3462#issuecomment-3112535889, I'm unsure whether I want to add this behaviour already. I think my preference is to include it in a large refactor that brings full multimodality support.

- Tom Aarsen

tomaarsen:
I've pushed the multimodality release, so this has been resolved now.

- Tom Aarsen

Issue #3368, opened 2025-05-20
Mistake in CoSENTLoss documentation


pashok3d:
In the CoSENTLoss implementation, documentation and the docstring states:
> `loss = logsum(1 + exp(s(i,j) - s(k,l)) + …)`, where `(i,j)` and `(k,l)` are any of the input pairs in the batch such that the expected similarity of `(i,j)` is greater than `(k,l)`.

However, the actual code masks on
https://github.com/UKPLab/sentence-transformers/blob/dd8a8d7d4d5660e7b2b69e8a1ece8202417c0069/sentence_transformers/losses/CoSENTLoss.py#L89
meaning it keeps pair-differences only when the true label of the first pair is less than the second. And computes
https://github.com/UKPLab/sentence-transformers/blob/dd8a8d7d4d5660e7b2b69e8a1ece8202417c0069/sentence_transformers/losses/CoSENTLoss.py#L86
(i.e. always s_i - s_j). As a result, the loss sums exp(s_i - s_j) over all (i,j) where y_i < y_j, not y_i > y_j as the docstring claims.

So it should be:
`loss = logsum(1 + exp(s(i,j) - s(k,l)) + …)`, where `(i,j)` and `(k,l)` are any of the input pairs in the batch such that the expected similarity of `(i,j)` is **smaller** than `(k,l)`.

pashok3d:
@tomaarsen Hi! Yes, indeed, it is related to that comment. IMO there was a confusion in https://github.com/UKPLab/sentence-transformers/issues/2437#issuecomment-2304623160 and #3230 PR actually changed correct description to incorrect one. 
The confusion arises because the image in https://github.com/UKPLab/sentence-transformers/issues/2437#issuecomment-2304623160 shows AnglE loss function that uses $s(u,v)$ for similarity and $\Delta\theta_{ij}$ for angle **difference**, whereas the formulation of loss function in documentation uses only **similarities**. Similarity (cosine) is **inversely** related to the angle difference, therefore it should be:

For CoSENT it should be:
``loss = logsum(1+exp(s(k,l)-s(i,j))+exp...)``, where ``(i,j)`` and ``(k,l)`` are any of the input pairs in the batch such that the expected similarity of ``(i,j)`` is greater than ``(k,l)``.

But for AnglE it should be:
[...]

PR #3496, opened 2025-08-22
Docs Patch for AnglE and CoSENT Losses

johneckberg:
# Pull Request overview

This PR addresses #3368 and ensures that the documentation for both CoSENT and AnglE loss align with the formulas in their respective reference papers.

### CoSENT Loss

- The CoSENT loss documentation has been reverted back to what is was before PR #3230 to match the formula that's in both the [2022 blog post](https://kexue.fm/archives/8847) and [2024 paper](https://penghao-bdsc.github.io/papers/CoSENT_TASLP2024.pdf) on CoSENT loss. 
- The doc string for CoSENT loss was also updated to reference the 2024 paper published by the IEEE, not the 2022 blog post.
- Changed the acronym expansion to match the title of the 2024 paper (Cosine Sentence to Consistent Sentence)

### AnglE Loss

- The AnglE Loss documentation was updated to match the formula described in the [2024 AoE paper](https://aclanthology.org/2024.acl-long.101/). 
- The AnglE doc string was also updated to point to the version published by the ACL, instead of the [pre-print](https://arxiv.org/abs/2309.12871v1) on Arxiv it currently references.

### Other notes

- This PR also updates the citation attribute for both loss functions to match their respective reference papers.
- The AnglE documentation states, "This is the same as CoSENTLoss, with a different similarity function." While this is true for the actual implementations , it may lead to some confusion as the published formulas for the two loss functions are slightly different. Maybe it would be best to leave issue #3368 open, or leave a note in the documentation of AnglE loss? I'm not sure what the best approach would be.

### DIFF
--- a/sentence_transformers/losses/CoSENTLoss.py
+++ b/sentence_transformers/losses/CoSENTLoss.py
@@ -120,8 +120,8 @@ class CoSENTLoss(nn.Module):
         return """
 @article{10531646,
     author={Huang, Xiang and Peng, Hao and Zou, Dongcheng and Liu, Zhiwei and Li, Jianxin and Liu, Kay and Wu, Jia and Su, Jianlin and Yu, Philip S.},
-    journal={IEEE/ACM Transactions on Audio, Speech, and Language Processing}, 
-    title={CoSENT: Consistent Sentence Embedding via Similarity Ranking}, 
+    journal={IEEE/ACM Transactions on Audio, Speech, and Language Processing},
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Thank you for opening. Feel free to ignore the failing CI, those are unrelated. 
I can leave #3368 open, users will be able to find this PR with some extra elaborations as well. I appreciate you for updating the links as well.

- Tom Aarsen

PR #3497, opened 2025-08-25
fix: add batch size parameter to model prediction in CrossEncoderRerankingEvaluator

emapco:
Update the call to `model.predict` to include `batch_size=self.batch_size`. Previously, the batch_size initialization argument was unused; this commit ensures that the user-provided value is utilized.


tomaarsen:
Hello!

Thanks a bunch for catching this, it was a total oversight. I appreciate it!

- Tom Aarsen

PR #3509, opened 2025-09-01
always pass `input_ids`, `attention_mask`, `token_type_ids`, `inputs_embeds` to forward

Samoed:
I think that using `inspect.signature` can be a breaking change, because custom model implementaion could have in  forward with `*args, **kwargs`?
```python
import inspect

def func(*args, **kwargs):
    pass
print(list(inspect.signature(func).parameters))
# ['args', 'kwargs']
```

E. g. `mme5` have just `features: Dict[str, torch.Tensor], **kwargs` in their implementaion and this can break it. 

https://huggingface.co/intfloat/mmE5-mllama-11b-instruct/blob/main/custom_st.py#L42

I think you can always pass `["input_ids", "attention_mask", "token_type_ids", "inputs_embeds"]`

### DIFF
--- a/sentence_transformers/models/Transformer.py
+++ b/sentence_transformers/models/Transformer.py
@@ -228,6 +228,8 @@ class Transformer(InputModule):
         """Returns token_embeddings, cls_token"""
         # Get the signature of the auto_model's forward method to pass only the expected arguments from `features`
         model_forward_params = list(inspect.signature(self.auto_model.forward).parameters)
+        # always pass ["input_ids", "attention_mask", "token_type_ids", "inputs_embeds"] to the forward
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Very fair, I think it makes sense to combine the signature approach for custom model architectures with some common defaults. Thank you for opening this so quickly, I appreciate it.

- Tom Aarsen

tomaarsen:
I made some minor changes to prevent having to inspect the signature on every forward pass. What do you think?

- Tom Aarsen

Samoed:
Great! This is better place for them

tomaarsen:
Perfect, thanks! Apologies for the failing CI, they're related to rate limits on huggingface, and I've not been able to resolve them yet. The tests pass locally though, so this should be all set.

- Tom Aarsen

PR #3516, opened 2025-09-14
fix: add makedirs to informationretrievalevaluator

stephantul:
The Information Retrieval Evaluator crashes if a csv path has been passed for which the directory does not yet exists. This is fine, but it only crashes _after_ an experiment is done and the csv writer tries to write to the path.

This PR fixes that by creating the directories before writing it. An alternative could be to check if it exists before starting an experiment and then crashing before starting the experiment. Also fixes a typo.


tomaarsen:
Hello!

Thanks a bunch for this fix. I'm not a big fan of the evaluators right now, mostly because there's so much duplication that should be moved to the superclass, but this is a solid fix for now.
Also thanks for the disc -> disk fix. I know disc is totally fine in British English, but I feel like disk is much more common nowadays. I've applied it all over.

- Tom Aarsen

PR #3521, opened 2025-09-21
Fix:Import SentenceTransformer class explicitly in losses module

altescy:
## What

Fix import statements in losses module:
`from sentence_transformers import SentenceTransformer` → `from sentence_transformers.SentenceTransformer import SentenceTransformer`

## Why

I found that some loss modules do not import `SentencdeTransformer` class:

```python
>>> import importlib
>>> mod = importlib.import_module("sentence_transformers.losses.MatryoshkaLoss")
>>> mod.SentenceTransformer   #  sentence_transformers/losses/MatryoshkaLoss.py imports SentenceTransformer via `from sentence_transformers import SentenceTransformer`
<module 'sentence_transformers.SentenceTransformer' from '.../sentence_transformers/SentenceTransformer.py'>
```

In this case, `SentencdeTransformer` **module** is imported, not a class.

This can cause an error in some situations such as when utilizing type hints:

```python
>>> typing.get_type_hints(MatryoshkaLoss.__init__)
  ...
TypeError: Forward references must evaluate to types. Got <module 'sentence_transformers.SentenceTransformer'  from '.../sentence_transformers/SentenceTransformer.py'>
```

It would be nice to import the class explicitly.

### DIFF
--- a/sentence_transformers/losses/AnglELoss.py
+++ b/sentence_transformers/losses/AnglELoss.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from sentence_transformers import SentenceTransformer, losses, util
+from sentence_transformers import losses, util
+from sentence_transformers.SentenceTransformer import SentenceTransformer
 
 
 class AnglELoss(losses.CoSENTLoss):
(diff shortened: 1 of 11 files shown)


tomaarsen:
I agree completely, thank you! It's a shame the files are named the same as the class names instead of proper snake-cased variants, but refactors in their names are a bit risky and hacky.

I'll gladly merge this once I've verified the tests.

P.s. the test failures are likely unrelated, I've been having issues with timeouts when downloading artifacts from Hugging Face.

- Tom Aarsen

PR #3525, opened 2025-09-25
bug: fix state bug in trainer

stephantul:
This PR fixes a state bug in the trainer. It was impossible to train on CPU because of this bug. I think it also could have caused a bunch of other (hidden) issues, because the whole global state was actually always the default state.

It's a bit complicated, but this is the gist: creating a `SentenceTransformerTrainingArguments` calls the `post_init` of its super class, `TransformerTrainingArguments`. This, in turn, sets some flags in a global mutable state object, `Accelerator.state.PartialState`. 

In the `SentenceTransformersTrainer`, a dummy `SentenceTransformerTrainingArguments` is created to figure out the default arguments that would have come out of an object without any specific arguments. This is done to list the arguments that are different from the default. But the creation of this `SentenceTransformersTrainingArguments` dummy resets the global state. As the default in the `TransformersTrainingArguments` is to take the best device, e.g., `mps` if it is available, it was impossible to actually train on `cpu`. 

The solution is very simple: make the dummy call use the already configured state instead of overwriting it.

### DIFF
--- a/sentence_transformers/trainer.py
+++ b/sentence_transformers/trainer.py
@@ -185,7 +185,9 @@ class SentenceTransformerTrainer(Trainer):
 
         # Get a dictionary of the default training arguments, so we can determine which arguments have been changed
         # for the model card
-        default_args_dict = SentenceTransformerTrainingArguments(output_dir="unused", accelerator_config={"use_configured_state": True}).to_dict()
+        default_args_dict = SentenceTransformerTrainingArguments(
+            output_dir="unused", accelerator_config={"use_configured_state": True}
+        ).to_dict()
 
         # If the model ID is set via the SentenceTransformerTrainingArguments, but not via the SentenceTransformerModelCardData,
         # then we can set it here for the model card regardless
--- a/sentence_transformers/trainer.py
+++ b/sentence_transformers/trainer.py
@@ -185,7 +185,7 @@ class SentenceTransformerTrainer(Trainer):
 
         # Get a dictionary of the default training arguments, so we can determine which arguments have been changed
         # for the model card
-        default_args_dict = SentenceTransformerTrainingArguments(output_dir="unused").to_dict()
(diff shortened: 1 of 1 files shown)


stephantul:
Here's a simple script you can run:

```python

from typing import cast

from datasets import Dataset, load_dataset
from transformers import AutoTokenizer

    SentenceTransformer,
    SentenceTransformerTrainer,
)
from sentence_transformers.losses import MatryoshkaLoss, MultipleNegativesRankingLoss
from sentence_transformers.models.StaticEmbedding import StaticEmbedding

if __name__ == "__main__":
    model_dim = 32
    static_embedding = StaticEmbedding(AutoTokenizer.from_pretrained("google-bert/bert-base-uncased"), embedding_dim=model_dim)
    model = SentenceTransformer(
        modules=[static_embedding],
    )
    print(model.device)
    # prints mps:0

    pubmedqa_dataset = cast(Dataset, load_dataset("sentence-transformers/pubmedqa", "triplet-20", split="train"))
    pubmedqa_dataset_dict = pubmedqa_dataset.train_test_split(test_size=100, seed=12)
    train_dataset: Dataset = pubmedqa_dataset_dict["train"]
    eval_dataset: Dataset = pubmedqa_dataset_dict["test"]

    dims = [model_dim]
    loss = MatryoshkaLoss(model, loss, matryoshka_dims=dims)

    run_name = "zzz`"
        # Required parameter:
        use_cpu=True,
    )

        model=model,
        args=args,
        train_dataset=train_dataset,
        eval_dataset=eval_dataset,
        loss=loss,
    )
    trainer.train()
    print(model.device)
    # prints cpu
```

tomaarsen:
Hello!

Works like a dream, thanks a ton for chasing this down. I know it took a long time, and I think it would have taken me an age as well. Much appreciated.

- Tom Aarsen

PR #3535, opened 2025-10-01
fix: allow MSE to be used with matryoshka

stephantul:
The MSELoss is currently not usable with Matryoshka because the target embeddings are not truncated together with the model embeddings. This PR changes the MSELoss to automatically truncate the target to be the size of the model when the model size < the target size.

This allows the model to be used safely with Matryoshka.
We do it automatically, but this is safe, because we only truncate the target to the model dimension if the model dimension is smaller. This will never lead to the model not learning specific dimensions.

The worst that could happen is that a user creates a mismatch in, e.g., the teacher embeddings and student embeddings, and that this passes silently. To prevent this, we could set some kind of flag to disallow this behavior when the model is not used in a Matryoshka environment (?). Not sure if this is overengineering, however.

I also added tests this time.

I also fixed the typing of the MSELoss, which was incorrect.

### DIFF
--- a/sentence_transformers/losses/MSELoss.py
+++ b/sentence_transformers/losses/MSELoss.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from collections.abc import Iterable
+from collections.abc import Sequence
 
 import torch
 from torch import Tensor, nn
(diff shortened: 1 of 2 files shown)


tomaarsen:
I'm tempted to instead go in a different direction and update MatryoshkaLoss rather than MSELoss. I'm actually somewhat a fan of giving an error when MSELoss is used with different dimensionalities. Then the user can make the conscious decision to truncate, or find some other solution.

See #3538

- Tom Aarsen

stephantul:
Makes sense, your change also makes it compatible with 2DMatryoshka out of the box, while this maybe doesn't.

PR #3537, opened 2025-10-02
Add support for Intel 'xpu' device in environment.py

domschl:


tomaarsen:
Hello!

I don't have access to an XPU decide to test this. Do you?
Is this required for loading a model with XPU devices conveniently? I imagine it means that you can just use this now:
```

model = SentenceTransformer("google/embeddinggemma-300m")
```

And it'll move it straight to XPU?

- Tom Aarsen

domschl:
I've tested this on a system with Intel Core Ultra 2.

- Without the patch, 'cpu' is given to torch as device for operations, and hence torch uses CPU, even if an XPU device is available with all versions of torch (including latest nightly)
- With the patch (and torch nightly 2.10.0.dev), one can select 'xpu' with SentenceTransformer, 'xpu' as device is given to torch, and calculations (I tested with embeddings, encode() API) correctly move from cpu to graphics xpu device, resulting in 2.5 x performance increase.

Current release version of torch 2.8 (at least with my hardware) always returns `False` on the `.is_available()` query.

I don't know what the plans for torch are to handle automatically using xpu if available at a later time (as 'mps' currently does).

So for now (according to my tests with the only hardware I have available) the patch and nightly torch are required to use xpu with SentenceTransformer.

tomaarsen:
I've updated the base branch from `v5.1-release` to `master`, as I make new changes on `master` and then merge them into the release branches when I make a release. I also ran some formatting changes, and had to force-push to place your commit on top of `master` instead of `v5.1-release`.

Thank you for answering my question re. the (now) automated use of XPU if it's available. I think this works quite smoothly now, and I'm glad to include it in the next release. Also, apologies for the delayed response, I tend to revisit promising PRs when I'm nearing another release (which is now). Thank you for your work!

- Tom Aarsen

PR #3543, opened 2025-10-09
[`fix`] correct dataset link in training scripts

thomasht86:
This link: https://huggingface.co/datasets/sentence-transformers/msmarco-hard-negatives/resolve/main/cross-encoder-ms-marco-MiniLM-L6-v2-scores.pkl.gz does not exist, resulting in an error when trying to run affected training scripts. 

Looking at https://huggingface.co/datasets/sentence-transformers/msmarco-hard-negatives/tree/main I see that https://huggingface.co/datasets/sentence-transformers/msmarco-hard-negatives/resolve/main/cross-encoder-ms-marco-MiniLM-L-6-v2-scores.pkl.gz does exist. So assuming that's the correct one. The training runs when correcting these.

### DIFF
--- a/examples/sentence_transformer/training/ms_marco/train_bi-encoder_margin-mse.py
+++ b/examples/sentence_transformer/training/ms_marco/train_bi-encoder_margin-mse.py
@@ -129,7 +129,7 @@ ce_scores_file = os.path.join(data_folder, "cross-encoder-ms-marco-MiniLM-L6-v2-
 if not os.path.exists(ce_scores_file):
     logging.info("Download cross-encoder scores file")
     util.http_get(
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Thanks a bunch, you're very right! I once renamed the https://huggingface.co/cross-encoder/ms-marco-MiniLM-L-6-v2 model to https://huggingface.co/cross-encoder/ms-marco-MiniLM-L6-v2 to align with the more common format to denote the number of layers. I also applied this renaming in the repository, but I never renamed/changed up that filename, while I did update this link in a mass replace.

The link that you're proposing here is correct.

- Tom Aarsen

PR #3550, opened 2025-10-16
`MegaBatchMarginLoss` use mini batches for positives too

benHeid:
This fixes that for very large batch sizes the encoding of the positives can cause OOMs. This is fixed by using the mini-batches also for embedding the positives instead of embedding all positives at once.


tomaarsen:
Hello!

Apologies for the delay. I think it makes a lot of sense to make this change, but I do think there's a few issues. Specifically, the `range(0, len(positive), ...)` doesn't work as `len(positive)` is e.g. ~3 because `positive` is a dictionary. Beyond that, we don't need to step in and out of eval/train mode and no_grad mode for every minibatch. I'll push a proposal.

- Tom Aarsen

PR #3565, opened 2025-11-07
ArXiv -> HF Papers

qgallouedec:


tomaarsen:
Great idea! 7 of the new links weren't indexed yet, so I've indexed those as well.
I'll double-check some edge cases, e.g. bibtex, but otherwise this is good to go!

- Tom Aarsen

PR #3567, opened 2025-11-11
Document broken LexRank pip implementation

stevenae:
Added a note about the broken pip implementation of LexRank and provided a link to a working implementation.


tomaarsen:
Hello!

This is a great idea, thanks for the proposal. I've made some slight changes, linking to `master` instead of a specific commit, and moving the comment up a bit because otherwise the linter complains. I'll merge this now!

- Tom Aarsen

PR #3589, opened 2025-12-08
feat/deps: make Pillow an optional dependency

akx:
Pillow is only used for a type check, so it shouldn't be pulled in as a required dependency.

This uses `getpixel` as an indicator of the given object being an image; all PIL images have that method.

### DIFF
--- a/tests/cross_encoder/test_train_stsb.py
+++ b/tests/cross_encoder/test_train_stsb.py
@@ -55,6 +55,7 @@ def evaluate_stsb_test(
     assert score > expected_score or abs(score - expected_score) < 0.1
 
 
+@pytest.mark.skipif("CI" in os.environ, reason="This test triggers rate limits too often in the CI")
 def test_pretrained_stsb(sts_resource: tuple[list[InputExample], list[InputExample]]):
     _, sts_test_samples = sts_resource
     model = CrossEncoder("cross-encoder/stsb-distilroberta-base")
@@ -76,6 +77,7 @@ def test_train_stsb_slow(
     evaluate_stsb_test(model, 75, sts_test_samples)
 
 
+@pytest.mark.skipif("CI" in os.environ, reason="This test triggers rate limits too often in the CI")
 def test_train_stsb(
     distilroberta_base_ce_model: CrossEncoder, sts_resource: tuple[list[InputExample], list[InputExample]]
 ) -> None:
--- a/tests/sparse_encoder/test_train_stsb.py
+++ b/tests/sparse_encoder/test_train_stsb.py
@@ -123,6 +123,7 @@ def test_train_stsb_slow(
     evaluate_stsb_test(model, 10, sts_test_samples)  # Lower expected score for a short training
 
 
+@pytest.mark.skipif("CI" in os.environ, reason="This test triggers rate limits too often in the CI")
 def test_train_stsb(
     dummy_sparse_encoder_model: SparseEncoder, sts_resource: tuple[list[InputExample], list[InputExample]]
 ) -> None:


tomaarsen:
Hello!

I think this is a nice direction, we should indeed try to remove this dependency. However, sadly I don't think I can justify this without a major release. After all, if someone installs `sentence-transformers<6.0`, and they expect it to also install Pillow, suddenly their pipelines will break. In short: I don't think I can remove dependencies without a major release, so instead I've added a comment to remove it for a v6.0.0. 

I've also updated the type check. I think the `getpixel` approach is a bit confusing, and this try-except import might be simpler. 

What do you think?

- Tom Aarsen

sentence_transformers/models/CLIPModel.py
```diff
@@ -7,10 +7,14 @@
...
 import transformers
-from PIL import Image
 
 from sentence_transformers.models.Router import InputModule
 
+try:
+    from PIL.Image import Image
+except Exception:
+    Image = None
+
```
Samoed: Wouldn't be better to import `PIL` inside `CLIP` and raise error/warning (FutureWarning)?

akx:
@tomaarsen Thanks for checking this out.

The `getpixel` check was deliberately done as a `hasattr` smell check to avoid having to even import PIL (which takes its own time [before we have lazy imports](https://peps.python.org/pep-0810/)) just to have the type available.

While I kind of technically agree that removing the dep would be a semver-major change, I would hope that people aren't just relying on an implicit dependency on PIL via this package's deps in production. (I also find it unlikely, if not impossible, that someone could be passing in PIL images to sentence-transformers without having used any other PIL APIs...)

Beyond that, if those users weren't locking their deps in some other way, even https://github.com/huggingface/sentence-transformers/commit/f9724ffe771a1c2a6107d646ccb504866794946b would possibly break their workflows since they could suddenly be pulling in transformers >= 5.0 when it wasn't previously possible due to the upper bound.

sentence_transformers/models/CLIPModel.py
tomaarsen: Great suggestion, applied it in 23718debb92a13ac989cdb719efa54f8f56ac8de

tomaarsen:
@akx Thanks for your views. Regarding the lazy nature of the import; I've now moved the import to the `tokenize` method based on @Samoed 's suggestion.

For the rest: I agree. I wouldn't expect people to only rely on `sentence-transformers` to provide their pillow if they're using it directly, and I think this change would only break for very few people, while reducing the install size for almost everyone. As for your argument re. transformers v5: I feel like that's trickier. Semver doesn't aim to force every package using X to release a major version just to upgrade X itself.
Beyond that, I'm loosening restrictions, rather than e.g. enforcing `transformers>=5`. The latter would more reasonably break people's workflows. 

I think very strictly, semantic versioning is just based on functionality and the public API of the project, so all of these changes are fine in a minor release. I still have to make up my mind on this, but I may revert the `pyproject.toml` changes I made in 615672af1a024bf59b896c63772b52e8e95fe720.

- Tom Aarsen

sentence_transformers/models/CLIPModel.py
```diff
@@ -68,12 +71,21 @@ def forward(self, features: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
         return features
 
     def tokenize(self, texts, padding: str | bool = True) -> dict[str, torch.Tensor]:
+        try:
+            from PIL.Image import Image
+        except Exception:
+            logger.warning(
+                "PIL.Image could not be imported. Image inputs will not be supported. "
+                "Please install the pillow package to enable image support: pip install pillow"
```
Samoed:
```suggestion
                "Please install the pillow package to enable image support: pip install sentence-transformers[image] or pip install pillow"
```

sentence_transformers/models/CLIPModel.py
```diff
@@ -68,12 +71,21 @@ def forward(self, features: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
...
+            )
+            Image = None
```
Samoed: Should be an error risen then because no images would be processed?

tomaarsen: Images have to be provided as a `pillow` `Image`, which is only possible if `pillow` is installed, so people would notice the lack of `pillow` install earlier. 
Just having a warning allows to still process texts with the CLIPModel, although that does seem slightly pointless.

Samoed: Maybe then use `warning.warn`, because to see logger messages you'd need to activate logging?

tomaarsen:
Good news, we should be able to simplify. `transformers==4.41.0` all the way up to v5 throws the following error when loading a CLIPImageProcessor:
```
ImportError:
CLIPImageProcessor requires the PIL library but it was not found in your environment. You can install it with pip:
`pip install pillow`. Please note that you may need to restart your runtime after installation.
```

In short, it's not possible to initialize a CLIPModel without a nice warning about `pillow` not being installed, courtesy of `transformers`. Then, this PR only needs to make the import lazy by moving it to `tokenize`, no errors, warnings or try-excepts needed.

I've also decided that I'm okay with including this in v5.2.

- Tom Aarsen

sentence_transformers/models/CLIPModel.py
tomaarsen: Just tried to test this, and noticed that `transformers` imposes a hard requirement for `pillow` to initialize the CLIP model, so that makes things a lot easier for us.

PR #3603, opened 2025-12-28
removed unused parameter in compute_loss_from_embeddings

artemguivan:
Removed labels parameter from compute_loss_from_embeddings method because we don't use it.


tomaarsen:
Hello!

Thanks for opening this. I've kept this parameter to align with the ~8 other cases of this method existing on other losses, so that a call to `loss.compute_loss_from_embeddings` has expected behaviour, even when you don't exactly know which loss it is, only that it has the `compute_loss_from_embeddings` method.

I don't think we should change this behaviour.

- Tom Aarsen

artemguivan:
Thank you for your reply!  I understand, I think your decision is the right one!

PR #3623, opened 2026-01-28
[Fix] Add padding for odd embedding dimensions in tensors (sparse encoders)

jadermcs:
Fix pairwise_angle_sim to handle odd embedding dimensions

The function uses `torch.chunk(x, 2, dim=1)` to split embeddings into real and imaginary components for complex angle calculations. This fails when the embedding dimension is odd since it cannot be evenly divided by 2.
```python
# y.size() == (32, 30503) (batch, vocabulary size, e.g., splade)

a, b = torch.chunk(x, 2, dim=1)
c, d = torch.chunk(y, 2, dim=1)
# c.size() == (32, 15252)
# d.size() == (32, 15251) ^^^ mismatch
z = torch.sum(c**2 + d**2, dim=1, keepdim=True) # size of c and d mismatches returning ERROR
```

I added a check for odd embedding dimensions and applied zero-padding (`torch.nn.functional.pad`) to both x and y tensors before chunking. The padding adds a single zero to the right side of the embedding dimension, which has minimal impact on the angle similarity computation while ensuring `torch.chunk` can split the tensors evenly.

### DIFF
--- a/sentence_transformers/util/similarity.py
+++ b/sentence_transformers/util/similarity.py
@@ -213,12 +213,11 @@ def pairwise_angle_sim(x: Tensor, y: Tensor) -> Tensor:
 
     x = _convert_to_tensor(x)
     y = _convert_to_tensor(y)
-    
-    # Pad tensors if the embedding dimension is odd
+
+    # Pad tensors if the embedding dimension is odd, as torch.chunk requires even dimensions
     if x.shape[1] % 2 != 0:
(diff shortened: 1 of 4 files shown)


tomaarsen:
Hello!

Nice work, I totally agree that we should patch this. I added a test and ran `pre-commit` so the quality checks are happy.
I'll merge this once the tests go green. Thanks!

- Tom Aarsen

tomaarsen:
Looks like there's some unrelated timeouts. No worries.

- Tom Aarsen

Issue #3621, opened 2026-01-27
Update example scripts relying on `model.fit` to use `SentenceTransformerTrainer` instead


tomaarsen:
Hello!

## Feature Request overview
* Many example scripts use `model.fit`, while I softly deprecated that method in v3.0 in favor of `SentenceTransformerTrainer`

## Details
Many example scripts use `model.fit`, while I softly deprecated that method in v3.0 in favor of `SentenceTransformerTrainer`. I wrote a migration guide here: https://sbert.net/docs/migration_guide.html#migrating-from-v2-x-to-v3-x

But never carried out the upgrade myself on all evaluation scripts. I did a bunch myself in https://github.com/huggingface/sentence-transformers/pull/2622, but there's some cases that I never tackled. For some of those it might not be possible to get a 1-1 perfect match (e.g. I believe one MS MARCO script uses `model.fit` because then it can fully regenerate the dataset for each batch), but it should be possible to get close.

Here's some that aren't done yet: https://github.com/search?q=repo%3Ahuggingface%2Fsentence-transformers+model.fit+language%3APython+path%3A%2F%5Eexamples%5C%2Fsentence_transformer%5C%2F%2F&type=code

If you're interested in working on this, feel free to take on just one case; I'm not expecting anyone to update all of them.

- Tom Aarsen

itsdevandy:
Update: Looks like someone else got to this first, thanks!

> Hi @tomaarsen, I’d like to pick this up and start working on it. 
> 
> I'll work on migrating these two from the listed examples:
> 
> 1. [train_stsb_simcse.py](https://github.com/huggingface/sentence-transformers/blob/1b4e222b1ef1798b418575b8be23ba530d9743d5/examples/sentence_transformer/unsupervised_learning/SimCSE/train_stsb_simcse.py#L90)
> 2. [train_bi-encoder_margin-mse.py](https://github.com/huggingface/sentence-transformers/blob/1b4e222b1ef1798b418575b8be23ba530d9743d5/examples/sentence_transformer/training/ms_marco/train_bi-encoder_margin-mse.py#L243)

harshitsharma496:
Hi @tomaarsen 

Picking up : 

1.examples/sentence_transformer/unsupervised_learning/CT_In-Batch_Negatives/train_ct-improved_from_file.py
2. examples/sentence_transformer/unsupervised_learning/CT/train_askubuntu_ct.py

[...]

PR #3625, opened 2026-01-28
migrate `train_ct_from_file.py` from v2 to v3

omkar-334:
Note - 
I replaced the `ContrastiveTensionDataLoader` with the logic written in `sentence_transformers/losses/ContrastiveTensionLoss.py`  
#3621

Previous  
..
Screenshot 2026-01-28 at 8 02 47 PM

New  
..
Screenshot 2026-01-28 at 8 03 06 PM


omkar-334:
hey @tomaarsen is there a clean way to replace `ContrastiveTensionDataLoader` instead of just copying over the logic?
or does this work?  
If the `ContrastiveTensionDataLoader` and other dataloaders are not needed after v3.x (since v3 now accepts HF dataset instead of pytorch dataloaders), do you think after migrating examples, the code could be removed?

omkar-334:
Also,I think this script will not get a perfect match because `ContrastiveTensionDataLoader` randomizes the order?

tomaarsen:
Hello!

Thanks for this!
If you'd like, you can take this PR a bit bigger:
1. Update the `ContrastiveTensionLoss` docstring & the `loss_overview.md` so that it accepts:
```
        Inputs:
            +------------------------+--------+
            | Texts                  | Labels |
            +========================+========+
            | (sentence_A, sentence_B) pairs | none   |
            +------------------------+--------+
```
2. Update the `ContrastiveTensionLoss` so that if `labels` is None, then it sets `labels` based on whether the pair is identical (label=1) or different (label=0). If the `labels` is already provided, you can just use the provided one.
3. Instead of using

```python
# Generate sentence pairs for ContrastiveTensionLoss
# For CT, we need pairs: positive pairs (same sentence) and negative pairs (different sentences)
# The ratio is controlled by pos_neg_ratio
def generate_ct_pairs(sentences, pos_neg_ratio):
    """Generate sentence pairs for ContrastiveTensionLoss with the specified pos_neg_ratio.

    This exactly replicates the logic of ContrastiveTensionDataLoader.__iter__():
    - Uses len(batch) % pos_neg_ratio to determine if pair is positive (0) or negative (>0)
    - For positive: (s1, s1) with label=1
    - For negative: (s1, s2) with label=0 where s2 is next sentence
    - Increments sentence_idx after each pair
    """
    pairs = []
    random.shuffle(sentences)
    sentence_idx = 0

    while sentence_idx + 1 < len(sentences):
        s1 = sentences[sentence_idx]
        if len(pairs) % pos_neg_ratio > 0:  # Negative (different) pair
            sentence_idx += 1
            if sentence_idx < len(sentences):
                s2 = sentences[sentence_idx]
                label = 0
            else:
                break
        else:  # Positive (identical pair)
            s2 = sentences[sentence_idx]
            label = 1

        sentence_idx += 1
        pairs.append({"sentence1": s1, "sentence2": s2, "label": label})

    return pairs


logging.info("Generating training pairs...")
train_pairs = generate_ct_pairs(train_sentences, pos_neg_ratio)
train_dataset = Dataset.from_list(train_pairs)
```
You can first make the `Dataset` with just texts and then `map` them to pairs with some `pos_neg_ratio`, e.g.:

```python
train_dataset = Dataset.from_dict({"text1": train_sentences})

def to_ct_pairs(sample, pos_neg_ratio=0.3):
    sample["text2"] = sample["text1"] if random.random() < pos_neg_ratio else random.choice(train_sentences)
    return sample

train_dataset = train_dataset.map(to_ct_pairs, fn_kwargs={"pos_neg_ratio": pos_neg_ratio})
```
(I didn't run this to test it, just wrote it in GitHub).
4. Update the remainder of the `ContrastiveTensionLoss` to remove `model.fit` in the example / remove that you must use `ContrastiveTensionDataLoader`. 
5. Also, could you please add some saving code like here: https://github.com/huggingface/sentence-transformers/pull/3631#discussion_r2740927619

If I'm correct, this should give near-identical results (apart from having different randomization), and still backwards compatible (if someone still provides labels), but it's more intuitive as having to provide `labels` is a bit weird if they're just exactly to denote whether the texts are the same.

Similarly, changes can be made to `ContrastiveTensionLossInBatchNegatives`: I feel like it should not be documented with `model.fit`, no longer mention the `ContrastiveTensionDataLoader`, and that it should accept `single sentences` with `none` labels. It should also only use `sentence_features[0]` for `self.model1` and `self.model2`. 
But perhaps that can be done in one of the `ct_improved` PRs, if you're interested!

- Tom Aarsen

omkar-334:
> You can first make the `Dataset` with just texts and then `map` them to pairs with some `pos_neg_ratio`, e.g.:
> 
> ```python
> train_dataset = Dataset.from_dict({"text1": train_sentences})
> 
> def to_ct_pairs(sample, pos_neg_ratio=0.3):
>     sample["text2"] = sample["text1"] if random.random() < pos_neg_ratio else random.choice(train_sentences)
>     return sample
> 
> train_dataset = train_dataset.map(to_ct_pairs, fn_kwargs={"pos_neg_ratio": pos_neg_ratio})
> ```
> 
> (I didn't run this to test it, just wrote it in GitHub).

Given that `pos_neg_ratio` is int, and it should be batch_size should be divisible by it, i think we should keep it that way. 

```python
train_dataset = load_dataset("sentence-transformers/askubuntu-questions", split="train").filter(
    lambda x: x["text"] not in dev_or_test_questions
)
sentences = train_dataset['text']


def to_ct_pairs(sample, pos_neg_ratio=8):
    pos_neg_ratio = 1 / pos_neg_ratio
    sample["text2"] = sample["text"] if random.random() < pos_neg_ratio else random.choice(sentences)
    sample["label"] = 1.0 if sample["text"] == sample["text2"] else 0.0
    return sample


train_dataset = train_dataset.map(to_ct_pairs, fn_kwargs={"pos_neg_ratio": pos_neg_ratio}).rename_column("text", "text1")
```  
 
I tried this out and this works....   
Although this approach nearly doubles the training data :)
Here's the logic behind it - 
1. Original DataLoader: Consumes 15 sentences to create 8 pairs (1 positive uses 1 sentence, 7 negatives use 14 sentences), so ~53% efficiency.
2. Map approach: Every sentence becomes a pair (each sentence = text1), so 100% efficiency → nearly 2x more pairs from the same data.

tomaarsen:
I think we're fine to go with the `map` approach, I missed the difference in efficiency. Also, thanks for fixing my snippet's issues (e.g. using a ratio, using a Dataset in the random.choice, etc.). I'll put the snippet in #3639 as well, and it should then be ready. The `label` portion should not be needed, though (although it does work to provide it).

- Tom Aarsen

omkar-334:
this is ready for review @tomaarsen

tomaarsen (approved):

PR #3626, opened 2026-01-28
migrate `train_stsb_ct.py` from v2 to v3

omkar-334:
#3621

Again, same issue as #3625 regarding the dataloader logic. It is the same dataloader, and the code is redundant here. 

Note - I'll attach the results after the wikipedia dataset gets downloaded

### DIFF
--- a/examples/sentence_transformer/unsupervised_learning/CT/train_stsb_ct.py
+++ b/examples/sentence_transformer/unsupervised_learning/CT/train_stsb_ct.py
@@ -3,9 +3,9 @@ import random
 import traceback
 from datetime import datetime
 
-from datasets import Dataset, load_dataset
+from datasets import load_dataset
 
-from sentence_transformers import InputExample, LoggingHandler, SentenceTransformer, losses, models
+from sentence_transformers import LoggingHandler, SentenceTransformer, losses, models
 from sentence_transformers.evaluation import EmbeddingSimilarityEvaluator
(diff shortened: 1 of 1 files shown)


tomaarsen:
I've made some comments here: https://github.com/huggingface/sentence-transformers/pull/3625#issuecomment-3817649758, which also apply to this PR! Feel free to let me know what you think.

- Tom Aarsen

omkar-334:
this is ready for review @tomaarsen. Do we need to adjust anything in `to_ct_pairs` regarding the label?

omkar-334:
@tomaarsen which do you recommend - `test_evaluator(model)` or `model.evaluate(test_evaluator)`?  
You've said something about the model evaluations also getting saved when we save the model... which method above enables this?  
I've seen discrepancies over the usage of this in the examples....

tomaarsen:
They are identical, `model.evaluate(evaluator)` simply calls `evaluator(model)` internally. I think I personally prefer `evaluator(model)`, but neither is strictly wrong.

- Tom Aarsen

omkar-334:
thanks for this! @tomaarsen... looks like i need to check the other open PRs and resolve them before you review

tomaarsen:
All good! I usually do a final push of my own if there's some tiny changes here and there that would finish a PR. It's usually just some style preference stuff.

- Tom Aarsen

PR #3627, opened 2026-01-28
migrate `train_stsb_ct_improved` from v2 to v3

omkar-334:
#3621


tomaarsen:
If you'd like, then some changes can be made to `ContrastiveTensionLossInBatchNegatives` to make it a bit stronger:
1. I feel like it should not be documented with `model.fit`, no longer mention the `ContrastiveTensionDataLoader`, 
2. it should accept `single sentences` with `none` labels. It should then use `sentence_features[0]` for `self.model1` and `self.model2`, it's a bit weird that it asks for the same text twice.

As for the PR as it stands now: 
1. I think it's a tad messy still due to the `EmbeddingSimilarityEvaluator` initializations being a tad unusual with the list comprehensions. That should become much much easier when we're using https://huggingface.co/datasets/sentence-transformers/stsb.
2. Could we use ratios for the training argument steps? Nowadays you can use e.g. 0.1 for `warmup_steps` and it's quite nice. I usually do 0.1 for warmup_ratio, 0.1 for save_steps, 0.1 for eval_steps, and 0.01 for logging_steps.
3. Could you add some saving/push_to_hub code, e.g. https://github.com/huggingface/sentence-transformers/pull/3631#discussion_r2740927619

Thanks a lot for your hard work!

- Tom Aarsen

omkar-334:
> If you'd like, then some changes can be made to `ContrastiveTensionLossInBatchNegatives` to make it a bit stronger:
> 
> 1. I feel like it should not be documented with `model.fit`, no longer mention the `ContrastiveTensionDataLoader`,
> 2. it should accept `single sentences` with `none` labels. It should then use `sentence_features[0]` for `self.model1` and `self.model2`, it's a bit weird that it asks for the same text twice.

hey, I've created a PR for this at #3639, with a few questions over there..

omkar-334:
> 1. I think it's a tad messy still due to the `EmbeddingSimilarityEvaluator` initializations being a tad unusual with the list comprehensions. That should become much much easier when we're using https://huggingface.co/datasets/sentence-transformers/stsb.

I've tried to simplify this logic.   
Do you think including `InputExample` is needed? I've removed the `InputExample` and just passed in lists. I'm not too sure about `train_dataset`, `dev_evaluator` and `test_evaluator`. I'll need to run it once and check.   

This PR will need #3639 to be merged first.

examples/sentence_transformer/unsupervised_learning/CT_In-Batch_Negatives/train_stsb_ct-improved.py
```diff
@@ -24,58 +28,77 @@
...
+)
 
-dev_evaluator = EmbeddingSimilarityEvaluator.from_input_examples(dev_samples, name="sts-dev")
-test_evaluator = EmbeddingSimilarityEvaluator.from_input_examples(test_samples, name="sts-test")
+test_evaluator = EmbeddingSimilarityEvaluator(
+    test["sentence1"],
+    test["sentence2"],
+    test["score"],
+    name="sts-test",
+)
```
tomaarsen, line 53: Very clean 👏 
We indeed don't need `InputExample`, that's also a relic from v2 (although it still works with some evaluators/`model.fit`.

examples/sentence_transformer/unsupervised_learning/CT_In-Batch_Negatives/train_stsb_ct-improved.py
```diff
@@ -24,58 +28,77 @@
...
 
-################# Initialize an SBERT model #################
+# Initialize an SBERT model
 word_embedding_model = models.Transformer(model_name, max_seq_length=max_seq_length)
 pooling_model = models.Pooling(word_embedding_model.get_word_embedding_dimension())
 model = SentenceTransformer(modules=[word_embedding_model, pooling_model])
 
-
```
omkar-334, line 58: @tomaarsen Why didn't this example have a `ContrastiveTensionDataloader` if it uses the CT loss?  
Should we add our `to_ct_pairs` function here?

tomaarsen: `ContrastiveTensionLossInBatchNegatives` never needed it, it uses in-batch samples as the negatives automatically.

PR #3628, opened 2026-01-28
migrate `train_askubuntu_ct-improved.py` from v2 to v3

omkar-334:
#3621

Previous  
..
Screenshot 2026-01-28 at 9 03 49 PM

Now  
..
(will attach soon)


tomaarsen:
My comments from https://github.com/huggingface/sentence-transformers/pull/3627#issuecomment-3817696102 also apply here, mostly, but this is already a big step in the right direction! Especially if the data loading is also simplified later.

- Tom Aarsen

omkar-334:
> My comments from [#3627 (comment)](https://github.com/huggingface/sentence-transformers/pull/3627#issuecomment-3817696102) also apply here, mostly, but this is already a big step in the right direction! Especially if the data loading is also simplified later.

I've applied step ratios and model saving... The `ContrastiveTensionLossInBatchNegatives` and duplicate sentences fix is dependant on #3639, so will change that later.  
The dataset migration is dependant on #3638, will update this soon.

tomaarsen (approved):

PR #3629, opened 2026-01-28
migrate `2_programming_train_bi-encoder.py` from v2 to v3

omkar-334:
Here we replace `NoDuplicatesDataLoader` functionality with `batch_sampler=BatchSamplers.NO_DUPLICATES` argument in `SentenceTransformerTrainingArguments`  

#3621

### DIFF
--- a/sentence_transformers/cross_encoder/losses/MultipleNegativesRankingLoss.py
+++ b/sentence_transformers/cross_encoder/losses/MultipleNegativesRankingLoss.py
@@ -138,11 +138,16 @@ class MultipleNegativesRankingLoss(nn.Module):
         # Create a mask for each anchor to each candidate index, where the matching positive
         # and hard negatives are masked out.
         mask = ~torch.eye(batch_size, dtype=torch.bool).repeat(1, num_columns)
-        if self.num_negatives is not None and self.num_negatives < len(candidates_flattened):
+
+        # Valid negatives for each anchor are all candidates except the ones that are the positive or hard negatives
+        # And num_columns = 1(positive) + num_hard_negatives
+        valid_negative_num = len(candidates_flattened) - num_columns
+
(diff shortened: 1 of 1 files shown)


examples/sentence_transformer/unsupervised_learning/query_generation/2_programming_train_bi-encoder.py
```diff
@@ -39,13 +39,31 @@
...
+args = SentenceTransformerTrainingArguments(
+    output_dir="output/programming-model",
+    num_train_epochs=num_epochs,
+    per_device_train_batch_size=batch_size,
     warmup_steps=warmup_steps,
-    show_progress_bar=True,
+    learning_rate=2e-5,
+    save_strategy="no",
+    logging_steps=100,
+    batch_sampler=BatchSamplers.NO_DUPLICATES,
```
tomaarsen, line 53: Nice, smart

examples/sentence_transformer/unsupervised_learning/query_generation/2_programming_train_bi-encoder.py
```diff
@@ -39,13 +39,31 @@
 
 # Tune the model
 num_epochs = 3
-warmup_steps = int(len(train_dataloader) * num_epochs * 0.1)
-model.fit(
-    train_objectives=[(train_dataloader, train_loss)],
-    epochs=num_epochs,
+batch_size = 64
+steps_per_epoch = len(train_dataset) // batch_size
+warmup_steps = int(steps_per_epoch * num_epochs * 0.1)
```
tomaarsen: We can pass a `warmup_steps=0.1` to the args.

tomaarsen:
I just found out that `warmup_steps=0.1` only works in `transformers` v5+, so it's best to use `warmup_ratio=0.1` instead still. Apologies for that. I think this is good to go, though! 

- Tom Aarsen

PR #3638, opened 2026-01-29
Replace `http_get` with `load_dataset` -`askubuntu` and `all-nli`

omkar-334:
~~Still have to replace `askubuntu` train - draft for now...~~
This Pull Request replaces legacy datasets - `askubuntu` and `all-nli` with the newer versions hosted on huggingface.

Issue #3264, opened 2025-03-10
SentenceTransformerTrainer with Streaming dataset


mrpeerat:
Hi,

I used the streaming dataset with SentenceTransformerTrainer but found this error.

**Code**
```
def main():
    model_name = "XXXX"
    batch_size = 32
    
    model = SentenceTransformer(
        model_name
    )

    train_dataset = load_dataset("aisingapore/sea-lion-pile", split="train", streaming=True)
    loss = MultipleNegativesRankingLoss(model)

    args = SentenceTransformerTrainingArguments(
        # Required parameter:
        output_dir=f"models/xlmr-{batch_size}-{temperature}",
        # Optional training parameters:
        num_train_epochs=1,
        per_device_train_batch_size=16,
        per_device_eval_batch_size=16,
        learning_rate=1e-5,
        warmup_ratio=0.1,
        fp16=False,  # Set to False if you get an error that your GPU can't run on FP16
        bf16=True,  # Set to True if you have a GPU that supports BF16
        batch_sampler=BatchSamplers.NO_DUPLICATES,  # MultipleNegativesRankingLoss benefits from no duplicate samples in a batch
        # Optional tracking/debugging parameters:
        eval_strategy="no",
        save_strategy="steps",
        save_steps=100,
        save_total_limit=2,
        logging_steps=100,
    )

    trainer = SentenceTransformerTrainer(
        model=model,
        args=args,
        train_dataset=train_dataset,
        # eval_dataset=eval_dataset,
        loss=loss,
        # evaluator=dev_evaluator,
    )
    trainer.train()

    # model.save_pretrained("models/mpnet-base-all-nli-triplet/final")

if __name__ == "__main__":
    main()
```

**Error**
```
[rank6]: Traceback (most recent call last):
[rank6]:   File "XXXX/sentence-transformers/simcse_new_trainer.py", line 150, in <module>
[rank6]:     main()
[rank6]:   File "XXXX/sentence-transformers/simcse_new_trainer.py", line 136, in main
[rank6]:     trainer = SentenceTransformerTrainer(
[rank6]:               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank6]:   File "XXXX/sentence-transformers/sentence_transformers/trainer.py", line 240, in __init__
[rank6]:     super().__init__(**super_kwargs)
[rank6]:   File "XXXX/envs/seateb/lib/python3.11/site-packages/transformers/trainer.py", line 620, in __init__
[...]

PR #3639, opened 2026-01-29
Update `ContrastiveTensionLoss` and `ContrastiveTensionLossInBatchNegatives`

omkar-334:
#3627

See [comment](https://github.com/huggingface/sentence-transformers/pull/3627/#issuecomment-3817696102) 

> If you'd like, then some changes can be made to ContrastiveTensionLossInBatchNegatives to make it a bit stronger:
>  1. I feel like it should not be documented with model.fit, no longer mention the ContrastiveTensionDataLoader,
>  2. it should accept single sentences with none labels. It should then use sentence_features[0] for self.model1 and self.model2, it's a bit weird that it asks for the same text twice.

Created a new PR for this since the scope felt a bit too large for a follow-up commit


sentence_transformers/losses/ContrastiveTensionLoss.py
omkar-334, line 122: Should we also edit the docstring for ContrastiveTensionLoss? I think this will need the PR for `ContrastiveTensionDataLoader`-alternative to be merged first

sentence_transformers/losses/ContrastiveTensionLoss.py
```diff
@@ -151,20 +150,34 @@ def __init__(self, model: SentenceTransformer, scale: float = 20.0, similarity_f
...
-                    InputExample(texts=['This is the 2nd sentence', 'This is the 2nd sentence']),
-                    InputExample(texts=['This is the 3rd sentence', 'This is the 3rd sentence']),
-                    InputExample(texts=['This is the 4th sentence', 'This is the 4th sentence']),
-                    InputExample(texts=['This is the 5th sentence', 'This is the 5th sentence']),
+                    InputExample(texts=['This is the 1st sentence']),
+                    InputExample(texts=['This is the 2nd sentence']),
+                    InputExample(texts=['This is the 3rd sentence']),
+                    InputExample(texts=['This is the 4th sentence']),
+                    InputExample(texts=['This is the 5th sentence']),
                 ]
```
omkar-334: Why do we have `train_examples` twice here? Should i remove the first variable?

sentence_transformers/losses/ContrastiveTensionLoss.py
```diff
@@ -174,9 +187,9 @@ def __init__(self, model: SentenceTransformer, scale: float = 20.0, similarity_f
         self.logit_scale = nn.Parameter(torch.ones([]) * np.log(scale))
 
     def forward(self, sentence_features: Iterable[dict[str, Tensor]], labels: Tensor) -> Tensor:
-        sentence_features1, sentence_features2 = tuple(sentence_features)
-        embeddings_a = self.model1(sentence_features1)["sentence_embedding"]  # (bsz, hdim)
-        embeddings_b = self.model2(sentence_features2)["sentence_embedding"]
+        sentence_features = sentence_features[0]
+        embeddings_a = self.model1(sentence_features)["sentence_embedding"]  # (bsz, hdim)
+        embeddings_b = self.model2(sentence_features)["sentence_embedding"]
```
tomaarsen, line 258: Perfect!

sentence_transformers/losses/ContrastiveTensionLoss.py
tomaarsen: Yeah I think the first one was always wrong (because you were meant to provide two identical texts).

And I think the cleanest is to immediately make a Dataset, e.g. like:
```python
                train_dataset = Dataset.from_dict({
                    "sentence": [
                       "It's nice weather outside today.",
                       "He drove to work.",
                       "It's so sunny.",
                    ]
                })
```
Then we don't need the InputExample (which is also a bit dated and I'd love to slowly stop using)

sentence_transformers/losses/ContrastiveTensionLoss.py
```diff
@@ -114,8 +114,7 @@ def __init__(self, model: SentenceTransformer, scale: float = 20.0, similarity_f
         are sampled from the batch. Using in-batch negative sampling gives a stronger training signal than the original :class:`ContrastiveTensionLoss`.
         The performance usually increases with increasing batch sizes.
 
-        Note that you should not use the `ContrastiveTensionDataLoader` for this loss, but just a normal DataLoader with `InputExample` instances.
-        The two texts of each `InputExample` instance should be identical.
+        Each `InputExample` instance must contain a single text.
```
tomaarsen: I think we can mention instead e.g. a Dataset with one text column:

```suggestion
        The training :class:`datasets.Dataset` must contain one text column.
```
P.s. the 
```
:class:`datasets.Dataset`
```
is from Sphinx autodoc and will automatically link to the Datasets documentation.

sentence_transformers/losses/ContrastiveTensionLoss.py
tomaarsen: I think it would be nice to upgrade that class too, like mentioned in https://github.com/huggingface/sentence-transformers/pull/3625#issuecomment-3817649758, but that can be implemented in this PR, the PR where I posted the comment, or a new PR, in case you want to work on that as well. Whatever you prefer 🤗

sentence_transformers/losses/ContrastiveTensionLoss.py
```diff
@@ -49,28 +49,39 @@ class ContrastiveTensionLoss(nn.Module):
...
+                    "It's so sunny.",
+                ]
+                "sentence_B": [
+                    "It's nice weather outside today.",
+                    "He drove to work.",
+                    "It's so sunny.",
+                ]
+                })
+
+            train_dataloader = ContrastiveTensionDataLoader(train_dataset, batch_size=3, pos_neg_ratio=3)
```
omkar-334: The only thing left is this... Should we update this with https://github.com/huggingface/sentence-transformers/pull/3625#issuecomment-3828489460
or let this be?

omkar-334: except this, I think this PR along with #3264 and #3638 are ready to be merged!

tomaarsen: Yes! I've updated it in 458cedd8d8b55e5fce3fd164f6d0ca0fde6ff806, alongside some other changes that I noticed. For example, the `label` approach required identical padding length across  the two inputs, but that wasn't always the case.

omkar-334: hmm, got it. It does look better now.

omkar-334:
> I think this is ready now! What do you think?

Yes!

PR #3641, opened 2026-01-30
Fix: select valid negatives as in-batch negatives

### LINKED ISSUE #3640: Masked elements can be selected as in-batch negatives in `MultipleNegativesRankingLoss`
Hello!

In `MultipleNegativesRankingLoss.get_in_batch_negatives()`, there is a bug where masked elements (which should be excluded) can still be selected as in-batch negatives under certain conditions. I think this is because `torch.multinomial()` can select the index which has zero value. I think this is weird but, PyTorch has not fixed yet. https://github.com/pytorch/pytorch/issues/48273 

## Reproduction
I added print statements in `get_in_batch_negatives()` to find the bug.
```python
def get_in_batch_negatives(
        self, anchors: list[str], candidates: list[list[str]]
    ) -> Generator[list[str], None, None]:
        batch_size = len(anchors)
        num_columns = len(candidates)

        # Given N anchors, we want to select num_negatives negatives for each anchor
        candidates_flattened = [candidate for sublist in candidates for candidate in sublist]

        # Create a mask for each anchor to each candidate index, where the matching positive
        # and hard negatives are masked out.
        mask = ~torch.eye(batch_size, dtype=torch.bool).repeat(1, num_columns)
        print("mask\n", mask)
        if self.num_negatives is not None and self.num_negatives < len(candidates_flattened):
            # From the remaining options, we randomly select num_negatives indices.
            negative_indices = torch.multinomial(mask.float(), self.num_negatives)
            print("negative_indices\n", negative_indices)
       # existing code ...
```
And trained the model.
```python

from datasets import Dataset

from sentence_transformers.cross_encoder import CrossEncoder, CrossEncoderTrainer, losses

model = CrossEncoder("microsoft/mpnet-base")
train_dataset = Dataset.from_dict(
    {
        "query": ["query1", "query2", "query3"],
        "positive": ["positive1", "positive2", "positive3"],
        "negative": ["negative1", "negative2", "negative3"],
    }
)
loss = losses.MultipleNegativesRankingLoss(model, num_negatives=5)

trainer = CrossEncoderTrainer(model=model, train_dataset=train_dataset, loss=loss)
trainer.train()
```
Here is the output.
```
mask
 tensor([[False,  True,  True, False,  True,  True],
        [ True, False,  True,  True, False,  True],
        [ True,  True, False,  True,  True, False]])
negative_indices
 tensor([[2, 1, 4, 5, 3],
        [2, 3, 5, 0, 1],
        [3, 1, 4, 0, 2]])
...
```
If my understanding is correct, last indexes of each list in negative_indices are positive or hard negative.

### Environment
- Python: 3.12.11
- torch: 1.11.0+cu113
- OS: ubuntu

## Suggested Fix
Calculate the number of valid negatives and replace `len(candidates_flattened)` with it.
```python
# Valid negatives for each anchor are all candidates except the ones that are the positive or hard negatives
# And num_columns = 1(positive) + num_hard_negatives
valid_negative_num = len(candidates_flattened) - num_columns

if self.num_negatives is not None and self.num_negatives < valid_negative_num:
    # From the remaining options, we randomly select num_negatives indices.
    negative_indices = torch.multinomial(mask.float(), self.num_negatives)
```
If this solution looks correct to you, I would be happy to open a Pull Request to fix it. Please let me know!

fuutot:
Hello @tomaarsen !

This PR fixes #3640.
As we discussed in the issue, simply checking `self.num_negatives < len(candidates_flattened)` is not enough.

I have replaced `len(candicates_flattened) ` with the number of valid negatives.

Let me know if you need any changes!

### DIFF
--- a/sentence_transformers/cross_encoder/losses/MultipleNegativesRankingLoss.py
+++ b/sentence_transformers/cross_encoder/losses/MultipleNegativesRankingLoss.py
@@ -138,11 +138,16 @@ class MultipleNegativesRankingLoss(nn.Module):
         # Create a mask for each anchor to each candidate index, where the matching positive
         # and hard negatives are masked out.
         mask = ~torch.eye(batch_size, dtype=torch.bool).repeat(1, num_columns)
-        if self.num_negatives is not None and self.num_negatives < len(candidates_flattened):
+
+        # Valid negatives for each anchor are all candidates except the ones that are the positive or hard negatives
+        # And num_columns = 1(positive) + num_hard_negatives
+        valid_negative_num = len(candidates_flattened) - num_columns
+
+        if self.num_negatives is not None and self.num_negatives < valid_negative_num:
             # From the remaining options, we randomly select num_negatives indices.
             negative_indices = torch.multinomial(mask.float(), self.num_negatives)
         else:
-            # If num_negatives is None or larger than the number of candidates, we select all negatives
+            # If num_negatives is None or larger than the number of valid negatives, we select all negatives
             # by using the mask as a slicer to get the indices of the negative candidates
             all_indices = torch.arange(batch_size * num_columns).repeat(batch_size, 1)
             negative_indices = all_indices[mask].reshape(batch_size, -1)


tomaarsen (approved):
Looks very nice, thank you for chasing down these issues in the CrossEncoder MNRL!

- Tom Aarsen

PR #3647, opened 2026-02-02
migrate `train_askubuntu_ct.py` from v2 to v3

omkar-334:
Part of #3621


omkar-334:
ready for review @tomaarsen

PR #3651, opened 2026-02-02
[feat]: add embedding Gemma spreadout loss

omkar-334:
TODO - 
1. add equation, requirements, example in Dosctring

Doubt - 
1. This loss is different from the GOR loss. As mentioned in EmbeddingGemma - 
```
It uses only the “second moment” term of the regularizer's original definition, as it finds this also sufficiently pushes the “mean” term towards its target value.
```
Do we want to include both Gemma spreadout loss and GOR?

cc @tomaarsen @AymenKallala


omkar-334:
The GOR loss is implemented at https://github.com/ColumbiaDVMM/Spread-out_Local_Feature_Descriptor/blob/master/tensorflow/patch_cnn_triplet.py

Issue #3650, opened 2026-02-02
Update docstring for `DenoisingAutoEncoderLoss.py`


omkar-334:
I think this is the last loss file for which the docstring has to be updated...   
https://github.com/huggingface/sentence-transformers/blob/0588ea0fe7fb95bb5d4ca1ee92630ac0dc58b62a/sentence_transformers/losses/DenoisingAutoEncoderLoss.py#L59

The implementation for the `noise_transform` is already present in `TSDAE` examples - https://github.com/huggingface/sentence-transformers/blob/main/examples/sentence_transformer/unsupervised_learning/TSDAE/train_stsb_tsdae.py

@tomaarsen Should I just replace the`DenoisingAutoEncoderDataset` with the implemented example's logic?

tomaarsen:
That would be quite nice indeed! I'd love to get rid of the old `model.fit` in all docs. Recently, I also updated some of the TSDAE docs in https://github.com/huggingface/sentence-transformers/pull/3619, this would nicely wrap up that TSDAE work.

- Tom Aarsen

omkar-334:
I noticed that `model.fit` still needs to be replaced in the MS-Marco examples (open PRs), this loss file, and the tests (`tests/test_train_stsb.py` and `tests/cross_encoder/test_train_stsb.py`). I think that would complete this work.

If that sounds good, I’m happy to take this on. I could also help with replacing `InputExample` or any other remaining pieces if needed.

tomaarsen:
Feel free! I'd be glad to move away from them fully.

- Tom Aarsen

PR #3653, opened 2026-02-03
Replace `model.fit` in test files

omkar-334:
Test file 1 -  
Screenshot 2026-02-03 at 6 20 40 PM  

Test file 2 -   
Screenshot 2026-02-03 at 6 14 37 PM   

> I noticed that `model.fit` still needs to be replaced in the MS-Marco examples (open PRs), this loss file, and the tests (`tests/test_train_stsb.py` and `tests/cross_encoder/test_train_stsb.py`). I think that would complete this work.
> 
> If that sounds good, I’m happy to take this on. I could also help with replacing `InputExample` or any other remaining pieces if needed. 

 _Originally posted by @omkar-334 in [#3650](https://github.com/huggingface/sentence-transformers/issues/3650#issuecomment-3837458557)_
 
Part of #3621

### DIFF
--- a/tests/cross_encoder/test_train_stsb.py
+++ b/tests/cross_encoder/test_train_stsb.py
@@ -1,15 +1,17 @@
 from __future__ import annotations
 
 import os
+import tempfile
 from collections.abc import Generator
 
 import pytest
-from datasets import load_dataset
-from torch.utils.data import DataLoader
+from datasets import Dataset, load_dataset
 
 from sentence_transformers import CrossEncoder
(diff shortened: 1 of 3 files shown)


omkar-334:
ready for review and merge

tomaarsen:
This is excellent, much cleaner. I've also extended it to the SparseEncoder, which already didn't use model.fit, but did use the old InputExample style still. This should be good to go! Thank you.

- Tom Aarsen

PR #3656, opened 2026-02-04
Add EmbeddingCache utility to avoid recomputing embeddings

iamgautamraj:
## Description

Adds a new [EmbeddingCache](cci:2://file:///d:/sentence-transformers/sentence_transformers/util/embedding_cache.py:19:0-280:9) utility that caches sentence embeddings to disk, avoiding recomputation for previously encoded sentences.

**Motivation:** While building a semantic search application, I noticed that my document corpus was being re-embedded every time I restarted my server. For large corpora, this wastes significant time and compute resources.

## Usage

```python
from sentence_transformers.util import EmbeddingCache

cache = EmbeddingCache("./embeddings_cache")

# First call - computes and caches
embeddings = cache.encode(model, ["Hello world", "How are you"])

# Second call - loads from cache (instant!)
embeddings = cache.encode(model, ["Hello world", "How are you"])
```

## Features
- Disk-based caching - Persists across restarts
- Model-aware - Different models use separate cache directories
- Partial cache hits - Only computes embeddings for new sentences
- Cache management - `clear()` and `stats()` methods included

## Changes
- Added `sentence_transformers/util/embedding_cache.py`
- Updated `sentence_transformers/util/__init__.py` to export EmbeddingCache
- Added `tests/util/test_embedding_cache.py` with 11 test cases

## Testing
All tests pass locally: `pytest tests/util/test_embedding_cache.py -v` # 11 passed

### DIFF
--- a/sentence_transformers/util/__init__.py
+++ b/sentence_transformers/util/__init__.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from .decorators import save_to_hub_args_decorator
 from .distributed import all_gather, all_gather_with_grad
+from .embedding_cache import EmbeddingCache
 from .environment import (
     check_package_availability,
     get_device_name,
@@ -49,6 +50,8 @@ __all__ = [
     # From distributed.py
     "all_gather",
     "all_gather_with_grad",
+    # From embedding_cache.py
+    "EmbeddingCache",
     # From environment.py
     "get_device_name",
     "check_package_availability",
--- a/sentence_transformers/util/embedding_cache.py
+++ b/sentence_transformers/util/embedding_cache.py
@@ -1,0 +1,281 @@
+"""Embedding cache utility for avoiding recomputation of sentence embeddings."""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import logging
+import os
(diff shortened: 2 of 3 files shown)


tomaarsen:
Hello!

Thank you for the proposal. I appreciate your effort on this, but I'm wary of including this in the core repository. Firstly, Sentence Transformers currently focuses on the model and training level, not so much the application level. Caching like this is scaffolding that users can (and do) build themselves or use dedicated libraries for. 

There's some extra concerns as well, e.g. having different `encode` parameters in the cache key, no option for `encode_query`/`encode_document`, cache invalidations, size limits/eviction policies, alternative storage backends, semantic caching, etc. If this gets added, it will likely result in a lot of extra feature requests (TTL, compression, etc.), and I'll end up doing a lot of maintenance work on that.

I think the best solution for now is to let more mature libraries specific for caching handle this behaviour, and let users find their own best working solution. 

- Tom Aarsen

iamgautamraj:
Thanks for the thoughtful feedback! You points are fair enough.
I appreciate you taking the time to explain the reasoning. I'll close this PR.

PR #3658, opened 2026-02-11
[`perf`] Speed up NoDuplicatesBatchSampler iteration (NO_DUPLICATES and NO_DUPLICATES_HASHED)

hotchpotch:
Hello!

## Pull request overview
- On very large datasets, I wanted to improve runtime and memory behavior, so this PR focuses on performance improvements in the sampler.
- In my measurements, this change gives a modest speedup for `NO_DUPLICATES` and around a 20% speedup for `NO_DUPLICATES_HASHED`.
- It targets the hot loop in `NoDuplicatesBatchSampler`.
- It aims to keep sampling behavior compatible with the historical implementation.
- It adds tests to check compatibility and dtype behavior.

## Details
The old implementation kept remaining indices in a Python `dict` and deleted accepted indices batch by batch.
That works, but it carries a lot of Python object overhead on large datasets.

This PR switches that part to:
- `randperm(..., dtype=int32|int64)` as a NumPy-backed index array
- a singly linked list over positions (`next_positions`) for O(1) removals

My intent here is to improve performance without changing semantics.
- Conflicting samples are still deferred and retried later.
- `drop_last` behavior is unchanged.
- Seeded determinism is unchanged.

About dtypes:
- Most datasets fit in `int32` row indices, so index/position NumPy arrays use `int32` in that case to reduce memory usage.
- They automatically switch to `int64` only when row count exceeds the `int32` range.
- Precomputed hashes stay `int64`.

## Benchmark
Setup:
- Script: `examples/sentence_transformer/evaluation/evaluation_no_dup_batch_sampler_speed.py`
- Dataset: `sentence-transformers/msmarco-co-condenser-margin-mse-sym-mnrl-mean-v1`
- Subset/split: `triplet-hard` / `train`
- Args: `--batch-size 8192 --no-progress-bar --precompute-num-proc 8`
- Baseline: `d0df0ccc`
- Current: this PR branch (`no_dup_hashed_improve`)
- Note: `NO_DUPLICATES` is reported with 3 runs because each run is long; `NO_DUPLICATES_HASHED` also has an extra 10-run check for stability.

### `precompute_hashes=False` (`NO_DUPLICATES`) (3-run mean)
| Metric | Baseline | Current | Delta |
| --- | ---: | ---: | ---: |
| Sampler time (s) | 175.216 | 171.275 | **-3.941s (-2.25%)** |

Runs:
| Run | Baseline (s) | Current (s) |
| --- | ---: | ---: |
| 1 | 174.716 | 170.041 |
| 2 | 176.797 | 172.140 |
| 3 | 174.136 | 171.644 |

### `precompute_hashes=True` (`NO_DUPLICATES_HASHED`) (3-run mean)
| Metric | Baseline | Current | Delta |
| --- | ---: | ---: | ---: |
| Sampler time (s) | 22.819 | 17.768 | **-5.051s (-22.13%)** |
| Hash build time (s) | 4.545 | 4.590 | +0.045s (+0.99%) |
| Total time (s) | 27.364 | 22.358 | **-5.006s (-18.29%)** |
| hash_uss current_delta (MiB) | 306.567 | 303.043 | **-3.523 MiB (-1.15%)** |
| hash_uss peak_delta (MiB) | 6891.520 | 6802.773 | **-88.747 MiB (-1.29%)** |

Runs:
| Run | Baseline sampler (s) | Baseline build (s) | Baseline total (s) | Current sampler (s) | Current build (s) | Current total (s) |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 23.079 | 4.582 | 27.661 | 17.740 | 4.590 | 22.330 |
| 2 | 22.884 | 4.582 | 27.466 | 17.622 | 4.618 | 22.240 |
| 3 | 22.493 | 4.472 | 26.965 | 17.942 | 4.563 | 22.505 |

### Extra check (`NO_DUPLICATES_HASHED`, 10 runs)
- Sampler time: `22.768s -> 17.842s` (**-21.64%**)

In short:
| Mode | Summary |
| --- | --- |
| `NO_DUPLICATES` | `-2.25%` (3-run mean) |
| `NO_DUPLICATES_HASHED` | `-22.13%` (3-run mean), `-21.64%` (10-run check) |

Memory note:
- For `NO_DUPLICATES_HASHED`, memory stays roughly flat in these runs, with a slight average decrease in both `hash_uss current_delta` and `hash_uss peak_delta`.

## Testing
- `uv run pytest tests/samplers/test_no_duplicates_batch_sampler.py -q`
- `uv run pytest tests/test_trainer.py -k "get_batch_sampler" -q`

Added tests:
- Output equality against a reference implementation of the historical dict-based algorithm (`drop_last` and `precompute_hashes` combinations)
- `precompute_hashes=True` stores precomputed hashes as `np.int64`
- Index/position arrays switch to `int64` when `int32` range is exceeded

## Related PR
- https://github.com/huggingface/sentence-transformers/pull/3611

---

This implementation was developed in collaboration with Codex, and all code has been reviewed by `@hotchpotch`.
If you have any questions or see anything that should be improved, I would really appreciate your feedback.

### DIFF
--- a/examples/sentence_transformer/training/other/training_batch_hard_trec.py
+++ b/examples/sentence_transformer/training/other/training_batch_hard_trec.py
@@ -174,5 +174,5 @@ except Exception:
     logging.error(
         f"Error uploading model to the Hugging Face Hub:\n{traceback.format_exc()}To upload it manually, you can run "
         f"`huggingface-cli login`, followed by loading the model using `model = SentenceTransformer({final_output_dir!r})` "
-        f"and saving it using `model.push_to_hub('{model_name}-multi-task')`."
+        f"and saving it using `model.push_to_hub('{model_name}-trec')`."
     )
--- a/sentence_transformers/evaluation/InformationRetrievalEvaluator.py
+++ b/sentence_transformers/evaluation/InformationRetrievalEvaluator.py
@@ -55,21 +55,25 @@ class InformationRetrievalEvaluator(SentenceEvaluator):
     Example:
         ::
 
+            import logging
             import random
+
+            from datasets import load_dataset
             from sentence_transformers import SentenceTransformer
             from sentence_transformers.evaluation import InformationRetrievalEvaluator
-            from datasets import load_dataset
+
+            logging.basicConfig(format="%(asctime)s - %(message)s", datefmt="%Y-%m-%d %H:%M:%S", level=logging.INFO)
 
             # Load a model
-            model = SentenceTransformer('all-MiniLM-L6-v2')
+            model = SentenceTransformer("all-MiniLM-L6-v2")
 
-            # Load the Touche-2020 IR dataset (https://huggingface.co/datasets/BeIR/webis-touche2020, https://huggingface.co/datasets/BeIR/webis-touche2020-qrels)
-            corpus = load_dataset("BeIR/webis-touche2020", "corpus", split="corpus")
-            queries = load_dataset("BeIR/webis-touche2020", "queries", split="queries")
-            relevant_docs_data = load_dataset("BeIR/webis-touche2020-qrels", split="test")
+            # Load the Touche-2020 IR dataset (https://huggingface.co/datasets/mteb/webis-touche2020-v3)
+            corpus = load_dataset("mteb/webis-touche2020-v3", "corpus", split="corpus")
+            queries = load_dataset("mteb/webis-touche2020-v3", "queries", split="train")
+            relevant_docs_data = load_dataset("mteb/webis-touche2020-v3", "default", split="test")
 
             # For this dataset, we want to concatenate the title and texts for the corpus
-            corpus = corpus.map(lambda x: {'text': x['title'] + " " + x['text']}, remove_columns=['title'])
+            corpus = corpus.map(lambda x: {"text": (x["title"] + " " + x["text"]).strip()}, remove_columns=["title"])
 
             # Shrink the corpus size heavily to only the relevant documents + 30,000 random documents
             required_corpus_ids = set(map(str, relevant_docs_data["corpus-id"]))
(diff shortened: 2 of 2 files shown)


hotchpotch:
Note: I successfully trained on multiple datasets (100k–50M each, ~800M rows in total) with `NO_DUPLICATES_HASHED` using this implementation.

tomaarsen:
Hello!

This is a bit of a trade-off between readability/simplicity and performance, but I'm okay with making that trade. However, we should really avoid the type of AI comments that refer to the "previous implementation" etc. Those comments are useful when you're reviewing generated changes in the moment, but not useful for readers later on. 

This is notably faster than before for me, and also faster than some alternatives that I just experimented with.

I think it might be ready if you're okay with my small edits!

- Tom Aarsen

tomaarsen:
I also sped up the tests a lot, so now they're not going to slow down the overall testing suite.

- Tom Aarsen

hotchpotch:
Hello!

No concerns from my side. I’d appreciate it if you could merge this when convenient.

> However, we should really avoid the type of AI comments that refer to the "previous implementation" etc. Those comments are useful when you're reviewing generated changes in the moment, but not useful for readers later on.

Your point about avoiding AI-generated comments that refer to “previous implementations” is absolutely fair. I left them in because I thought they might help readability, but I agree they become noise for future readers. I’ll be more careful going forward so the code stays easy to maintain for ongoing development.

Also, thank you for proactively removing those comments and for fixing the slow test issue as well.

I often develop with coding agents, and I recognize that if the resulting PR quality is not high enough, it can increase the workload for OSS maintainers, contributors, and reviewers. I want to keep that burden as low as possible by submitting changes that are easier to review and maintain. If my process ever adds review overhead, or if there are things I should handle myself before requesting review, please feel free to tell me directly.


Thanks again for the thoughtful review.

tomaarsen:
Thanks for your candor! I have no concerns about coding agents as long as they're used by people already experienced with software development and/or retrieval systems. And overall, comments about how something is implemented are solid, but agents sometimes add comments that clarify why a change was made, instead of why the new code is the way it is.

For example:
> # Keep this import outside the try-block below so ImportError is not
> # mistaken for hash-shape/data issues.

This explains why the agent moved the import out of the try-except, but for future readers, the import being inside of the try-except wasn't even a thought.

Or
> # Keep the same random order semantics as before, but avoid Python dict/list overhead.
> # This array stores only row indices, so int32 is sufficient for <=2^31-1 rows and
> # cuts index-memory roughly in half for very large datasets.

The "as before" is a bit confusing to future readers: they probably won't know what was before.

Or from another PR:
> The key insight is that caching happens at the SpladeLoss (wrapper) level, not at the base loss level.

In this case, a reader wouldn't think about caching happening at the base loss level, this comment likely only exists because caching in the base loss level is a possible alternative, just not the one implemented.

I really appreciate your PRs, and your general work on retrieval and beyond is also very valuable. For example, I used your https://github.com/hotchpotch/yasem project as inspiration for the SparseEncoder API.

- Tom Aarsen

hotchpotch:
Thank you for merging this PR, and thank you as well for the concrete examples in your feedback.

Your comments made me realize that when I use coding agents, I need to review both implementation and inline　comments much more carefully. I’ll take extra care to avoid the same kind of mistakes going forward.

I’m also very happy to hear that my project was helpful as inspiration!

PR #3663, opened 2026-02-17
Improve MatryoshkaLoss warning guidance for matryoshka_dims settings

hotchpotch:
Hello!

## Pull Request overview
- This PR adds clearer guidance in `MatryoshkaLoss` for `matryoshka_dims` settings that are not aligned with the recommended consistent halving pattern described in the Matryoshka paper (Section 3).
- The goal is to make this setup issue easier to notice during training configuration while keeping such configurations allowed.

## Details
- Kept the existing warning when the model’s full embedding dimension is not included in `matryoshka_dims`.
- Added a separate warning when `matryoshka_dims` do not follow the recommended consistent halving schedule after sorting.
- Added a direct reference to the MRL paper Section 3 in the warning message:
  `https://arxiv.org/html/2205.13147v4#S3`
- Added/updated tests to validate:
  - warning emission for non-halving settings
  - warning behavior when both warning conditions are met
  - warning behavior when only one condition is met
  - non-halving behavior for two-dimension settings

- As written in Section 3 of the [Matryoshka Representation Learning paper](https://arxiv.org/html/2205.13147v4#S3):

> The usual set ℳ consists of consistent halving until the representation size hits a low information bottleneck.

## Testing
- `uv run pytest tests/losses/test_MatryoshkaLoss.py -k "warn or warning or halving"`

## Note
- In local training runs, MRL performance was slightly worse without consistent halving, but improved after switching to a consistent halving schedule.

### DIFF
--- a/sentence_transformers/losses/MatryoshkaLoss.py
+++ b/sentence_transformers/losses/MatryoshkaLoss.py
@@ -211,6 +211,18 @@ class MatryoshkaLoss(nn.Module):
                     "when using the model without specifying a lower truncation dimension. It is strongly recommended to include "
                     f"{model_embedding_dim} in matryoshka_dims."
                 )
+        sorted_dims = sorted(matryoshka_dims, reverse=True)
+        if len(sorted_dims) >= 2:
+            # Warn when dims do not follow a strict halving schedule, e.g. d, d/2, d/4, ...
+            # This matches the "consistent halving" guidance from the Matryoshka paper (2205.13147).
+            non_halving = any(sorted_dims[i] != 2 * sorted_dims[i + 1] for i in range(len(sorted_dims) - 1))
+            if non_halving:
+                logger.warning(
+                    "matryoshka_dims do not follow the recommended consistent halving schedule after sorting: %s. "
+                    "This configuration is allowed, but a typical schedule is [d, d/2, d/4, ...]. "
+                    "Refer to the MRL paper, Section 3: https://arxiv.org/html/2205.13147v4#S3",
+                    sorted_dims,
+                )
 
         # Sort the dimensions and weights in descending order
         dims_weights = zip(matryoshka_dims, matryoshka_weights)
--- a/tests/losses/test_MatryoshkaLoss.py
+++ b/tests/losses/test_MatryoshkaLoss.py
@@ -56,11 +56,14 @@ def test_dimension_exceeds_model_dimension_raises_error(static_retrieval_mrl_en_
         )
 
 
-def test_model_dimension_not_in_dims_warns(static_retrieval_mrl_en_v1_model, mock_loss, caplog):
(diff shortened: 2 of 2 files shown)


hotchpotch:
With the current implementation in this PR, the following common `matryoshka_dims` pattern will emit a warning:
`[768, 512, 256, 128, 64]` (because `768 -> 512` is not strict halving).

I noticed this pattern is also used in README/example snippets:
- `examples/sentence_transformer/training/matryoshka/matryoshka_nli.py`
- `examples/sentence_transformer/training/matryoshka/matryoshka_sts.py`
- `examples/sentence_transformer/training/matryoshka/README.md`
- `examples/sentence_transformer/training/adaptive_layer/README.md`

If you prefer, I can update these docs/examples in this PR for consistency.
Otherwise, I can keep this PR scoped to warning behavior only.

If helpful, I can also close this PR entirely. Please don’t hesitate to say so if this warning feels unnecessary. I’m completely fine with closing it and moving in a different direction.

tomaarsen:
I ran some very very simple tests on matryoshka_nli.py, and couldn't really get any notable difference between e.g. 768, 512, 256, 128 and "exact halving" with 768, 384, 192, 96, 48. 

image

I evaluated the exact halving model on the exact halving and non-exact halving dimensions, but there's no big differences, so I'm not necessarily sure that exact halving is really important for good performance. Could you share some details from your tests perhaps?

- Tom Aarsen

hotchpotch:
Hello!

Thank you for the review and for taking the time to validate this carefully.

I want to share a follow-up from my side. My original assumption came from a private 384-dim experiment where I observed slightly worse full-dimension performance with a non-halving schedule, and I concluded that strict halving was necessary.

After your feedback, I reran this in a reproducible setup with public resources.

Base model: `hotchpotch/mMiniLMv2-L6-H384`

Setup summary:
- Training data: GooAQ (`3,000,000` samples)
- Batch size: `256`
- Learning rate: `5e-5`
- Objective: MNRL + MRL
- Evaluation: NanoBEIR full, measured at each truncate dimension

Results:

| MRL dims            | Truncate Dim | Score  |
|---------------------|--------------|--------|
| [384, 192, 96, 48]  | 384          | 0.3043 |
| [384, 192, 96, 48]  | 192          | 0.3041 |
| [384, 192, 96, 48]  | 96           | 0.2995 |
| [384, 192, 96, 48]  | 48           | 0.2939 |
| [384, 256, 128, 64] | 384          | 0.3118 |
| [384, 256, 128, 64] | 256          | 0.3138 |
| [384, 256, 128, 64] | 128          | 0.3084 |
| [384, 256, 128, 64] | 64           | 0.3051 |

In this reproducible run, `[384, 256, 128, 64]` performed better than `[384, 192, 96, 48]`, which is consistent with your point that MRL can work well without strict halving.

Given this, I will close this PR.

Thank you again for your guidance, and sorry for the extra review cycle.

PR #3668, opened 2026-02-19
[`fix`] `GroupByLabelBatchSampler` to guarantee multi-class batches for triplet losses

### LINKED ISSUE #2920: Difference between `SentenceLabelDataset` and `GroupByLabelBatchSampler`
Hi @tomaarsen 
First of all - kudos to you maintaining such a pragmatic library.

I am facing some difficulty on using `GROUP_BY_LABEL` batch sampler in v3.0 and want to highlight the issues to check if there is any way to mitigate those.

I went through the issues and found this: https://github.com/UKPLab/sentence-transformers/issues/2698#issuecomment-2144534374
You have mentioned here that the idea is to replace `SentenceLabelDataset` by `GroupByLabelBatchSampler` but I think there is very drastic differences between the two and we haven't retained the same functionality of `SentenceLabelDataset` while opting for `GroupByLabelBatchSampler` as a replacement.

I am taking an detailed example to explain the differences:

Let's take a simple example with a list of integers representing classes, and we'll use it to illustrate how the two approaches handle homogeneity in batch construction.

### Example Data:
- **Classes:** `[1, 1, 1, 1, 2, 2, 2, 3, 3, 4]`
- **Number of Samples per Label:**
  - Class 1: 4 samples
  - Class 2: 3 samples
  - Class 3: 2 samples
  - Class 4: 1 sample
- **Batch Size:** 4
- **Samples per Label in the Second Method:** 2

#### **Step 1: Initialization**
- **Grouping by Labels:**
  - Class 1: `[0, 1, 2, 3]`
  - Class 2: `[4, 5, 6]`
  - Class 3: `[7, 8]`
  - Class 4: `[9]`

- **Truncate to Even Number:**
  - Class 1: `[0, 1, 2, 3]` (all 4 samples)
  - Class 2: `[4, 5]` (only 2 samples, 1 sample is discarded)
  - Class 3: `[7, 8]` (both samples)
  - Class 4: `[9]` (removed because it doesn't have at least 2 samples)

- **Resulting Groups:**
  - Class 1: `[0, 1, 2, 3]`
  - Class 2: `[4, 5]`
  - Class 3: `[7, 8]`

#### **Step 2: Batch Construction**
- **Batch 1:**
  - Randomly selects Class 1 and takes all samples: `[0, 1, 2, 3]`
  - **Result:** `[0, 1, 2, 3]` (4 samples from Class 1, fully homogeneous)

- **Batch 2:**
  - Randomly selects Class 2 and takes both samples: `[4, 5]`
  - Randomly selects Class 3 and takes both samples: `[7, 8]`
  - **Result:** `[4, 5, 7, 8]` (2 samples from Class 2, 2 samples from Class 3)

#### **Step 1: Initialization**
- **Label Range:** `[1, 2, 3, 4]`
- **Already Seen Dictionary:** Empty

#### **Step 2: Batch Construction**
- **Batch 1:**
  - Starts with Class 1, takes 2 samples (e.g., `[0, 1]`)
  - **Result:** `[0, 1]` (2 samples from Class 1)
  - Moves to Class 2, takes 2 samples (e.g., `[4, 5]`)
  - **Result:** `[0, 1, 4, 5]` (2 samples from Class 1, 2 samples from Class 2)

- **Batch 2:**
  - Continues with Class 1, takes remaining 2 samples (e.g., `[2, 3]`)
  - **Result:** `[2, 3]` (2 samples from Class 1)
  - Continues with Class 2 but only 1 sample is left, so skips it and moves to Class 3
  - Takes both samples from Class 3 (e.g., `[7, 8]`)
  - **Result:** `[2, 3, 7, 8]` (2 samples from Class 1, 2 samples from Class 3)

- **Batch 3:**
  - If continuing, it would reset, shuffle, and start over, potentially including Class 4 depending on whether `with_replacement` is `True` or `False`.

### **Comparison of Homogeneity:**

- **`GroupByLabelBatchSampler`:**
  - **Batch 1:** Fully homogeneous batch from Class 1: `[0, 1, 2, 3]`.
  - **Batch 2:** Mixes samples from two classes (Class 2 and Class 3) because neither can fully fill a batch on their own: `[4, 5, 7, 8]`.
  - **Overall:** Prioritizes keeping batches homogeneous when possible, especially when a class has enough samples to fill an entire batch. It avoids mixing classes unless necessary.

- **`SentenceLabelDataset`:**
  - **Batch 1:** Mixed from the start, taking 2 samples from Class 1 and 2 from Class 2: `[0, 1, 4, 5]`.
  - **Batch 2:** Also mixed, with samples from Class 1 and Class 3: `[2, 3, 7, 8]`.
  - **Overall:** More likely to mix classes in every batch. It enforces a fixed number of samples per label and sequentially processes labels, leading to less homogeneous batches overall.

# TL;DR:
I am trying to fine-tune sentence transformers models using the dataset with this label distribution:
```
Class 1: 5000 Samples
Class 2: 5000 Samples
Class 3: 3000 Samples
Class 5 to 50: Less than 50 samples each
```
In the new `GroupByLabelBatchSampler` the batching logic is yielding most of the batches as homogeneous and there is not much improvement observed after fine-tuning. 
IMO this type of data could have been easily used with `SentenceLabelDataset` as it ensures there is at max `N` samples from each label in a batch. Intuitively, ST models should benefit from having in-batch negatives and more heterogeneous batches.

Can you help me in veryfying if my understanding is correct and if yes, is there any way to opt for the older logic?

MrLoh:
Fixes #2920

### Problem

`GroupByLabelBatchSampler` produced ~99% single-class batches, causing **zero gradients** with all four `Batch*TripletLoss` classes it was recommended for. The old implementation simply concatenated all samples for each label and chunked them into batches, so the only multi-class batches occurred at label boundaries.

### Solution

Replace `GroupByLabelBatchSampler` in-place with PK sampling + dynamic P scheduling. No new class or enum -- `BatchSamplers.GROUP_BY_LABEL` now does the right thing.

Each batch draws **K samples** (samples per label, K >= 2) from each of **P distinct labels** (labels per batch, P >= 2):

- `P = min(num_labels, batch_size // 2)` -- maximize label diversity
- `K = batch_size // P` -- fill each label's slot equally
- A greedy schedule pre-computed in `__init__` assigns labels to batches

**Dynamic P shrinking**: as minority labels are exhausted, the scheduler reduces P (down to 2) so that majority-label samples aren't discarded. This addresses the sample efficiency concern raised in review -- e.g. TREC coarse labels go from 9.4% to 99.3% sample utilization.

**Minority label inclusion**: labels with at least 2 samples but fewer than K are still scheduled into a proper multi-class batch (contributing all their available samples). This ensures they participate in in-batch triplet mining and actually influence the loss, rather than being relegated to a remainder batch where they may be the only label present.

### What changed

- **`sampler.py`**: Replaced `GroupByLabelBatchSampler.__init__`, `__iter__`, added `__len__` with PK sampling and dynamic P scheduling. Minority labels (>= 2 samples, < K) get at least 1 scheduled draw. Deleted the separately-added `BalancedLabelBatchSampler`.
- **`training_args.py`**: Updated `GROUP_BY_LABEL` docstring to describe the new behavior. Removed `BALANCED_LABELS` enum.
- **`test_group_by_label_batch_sampler.py`**: Rewrote tests for PK behavior -- multi-class guarantees, min samples per label, batch size consistency, drop_last, sample coverage, `__len__` accuracy, error cases, minority label inclusion in PK batches, and two efficiency tests for imbalanced data (TREC coarse-like and fine-like distributions).
- **Loss docstrings** (4 files): Updated recommendation text to accurately describe the multi-label guarantee.
- **Docs** (`datasets.md`, `migration_guide.md`): Updated descriptions of `GROUP_BY_LABEL`.
- **Example** (`training_batch_hard_trec.py`): Updated comment.

### Note on pre-existing UP038 lint warnings

There are pre-existing `UP038` ruff warnings in `sampler.py` (using tuple syntax in `isinstance` calls). This rule was [deprecated from ruff's recommended set](https://github.com/astral-sh/ruff/issues/7871) because the `X | Y` alternative has runtime overhead and edge-case issues. Consider adding `"UP038"` to the ignore list in `pyproject.toml`.

### DIFF
--- a/sentence_transformers/sampler.py
+++ b/sentence_transformers/sampler.py
@@ -194,7 +194,7 @@ class GroupByLabelBatchSampler(DefaultBatchSampler):
 
     def __len__(self) -> int:
         n = self._stream_length // self.batch_size
-        if not self.drop_last and self._stream_length % self.batch_size > 0:
+        if not self.drop_last and self._stream_length % self.batch_size >= 4:
             n += 1
         return n
 
--- a/tests/samplers/test_group_by_label_batch_sampler.py
+++ b/tests/samplers/test_group_by_label_batch_sampler.py
@@ -96,6 +96,18 @@ def test_len_matches_iteration(balanced_dataset: Dataset) -> None:
         assert len(sampler) == len(batches), f"drop_last={drop_last}: __len__={len(sampler)} != actual={len(batches)}"
 
 
+def test_len_matches_iteration_remainder_of_two() -> None:
+    """When stream_length % batch_size == 2, __len__ and __iter__ must agree.
+
+    2 labels with 6 and 4 samples -> stream_length=8, batch_size=6 -> remainder=2.
+    """
+    labels = [0] * 6 + [1] * 4
+    ds = Dataset.from_dict({"data": list(range(10)), "label": labels})
+    sampler = GroupByLabelBatchSampler(dataset=ds, batch_size=6, drop_last=False, valid_label_columns=["label"])
+    batches = list(sampler)
+    assert len(sampler) == len(batches), f"__len__={len(sampler)} but __iter__ yielded {len(batches)} batches"
+
+
 def test_raises_on_single_label() -> None:
     data = {"data": list(range(20)), "label": [0] * 20}
     ds = Dataset.from_dict(data)


tomaarsen:
Hello!

First of all, thanks a lot for the detailed findings/experiments here! I checked my `examples` directory and my uploaded models on HF, and I saw that I indeed have very few experiments with these Batch...TripletLosses. Back when I refactored the training in Sentence Transformers for v3, I softly deprecated the old ...Dataset classes with these samplers which were supposed to behave identically. However, it does seem like I messed up the GroupByLabelBatchSampler.

I ran some tests using the freshly updated https://github.com/huggingface/sentence-transformers/blob/main/examples/sentence_transformer/training/other/training_batch_hard_trec.py (thanks @omkar-334!)
* https://huggingface.co/tomaarsen/all-distilroberta-v1-trec-balanced-sampler#training-logs (From this PR): 0.9434 triplet accuracy, 122 steps, gradual loss from 4.8492 -> 4.0453
* https://huggingface.co/tomaarsen/all-distilroberta-v1-trec-group-by-label#training-logs (Original GroupByLabel): 0.9313 triplet accuracy, 155 steps, jumpy loss ranging from 0.0 to 1.8914
* https://huggingface.co/tomaarsen/all-distilroberta-v1-trec-batch-sampler#training-logs (The default BatchSampler): 0.9313 triplet accuracy, 155 steps, gradual loss from 4.8467 -> 3.9052

In short: I think this is a great sign, but does show one concern:
```
> scheduled_samples
3872
> total_samples
4951
```
I think this is because there's probably 1 class with a lot of samples, but it's still only used twice in each batch, and then eventually there's no samples from other labels to place into the same batch (we need P >= 2 after all). 

I had Claude write up this script:

<details><summary>Click to expand</summary>

```python
"""
Debug script to diagnose label distribution and batch sampler efficiency.
"""
from collections import defaultdict, Counter

# Load TREC dataset
dataset = load_dataset("omkar334/trec", split="train")
print(f"Total samples: {len(dataset)}")
print(f"Total unique labels: {len(set(dataset['fine_label']))}")

# Count samples per label
label_counts = Counter(dataset['fine_label'])
print(f"\nLabel distribution:")
print(f"  Min samples per label: {min(label_counts.values())}")
print(f"  Max samples per label: {max(label_counts.values())}")
print(f"  Mean samples per label: {sum(label_counts.values()) / len(label_counts):.1f}")

# Show top 10 labels by count
print(f"\nTop 10 labels by sample count:")
for label, count in label_counts.most_common(10):
    print(f"  Label {label}: {count} samples")

# Show bottom 10 labels by count
print(f"\nBottom 10 labels by sample count:")
for label, count in sorted(label_counts.items(), key=lambda x: x[1])[:10]:
    print(f"  Label {label}: {count} samples")

# Simulate the sampler's P/K computation
batch_size = 32

# Step 1: Filter labels with at least 2 samples
filtered = {lab: count for lab, count in label_counts.items() if count >= 2}
print(f"\nAfter filtering (>= 2 samples): {len(filtered)} labels kept")

# Step 2: Iteratively compute P/K
iteration = 0
while True:
    iteration += 1
    num_labels = len(filtered)
    p = min(num_labels, batch_size // 2)
    k = batch_size // p
    print(f"  Iteration {iteration}: P={p}, K={k}, num_labels={num_labels}")
    
    new_filtered = {lab: count for lab, count in filtered.items() if count >= k}
    if len(new_filtered) == len(filtered):
        print(f"  → Converged!")
        break
    print(f"  → {len(filtered)} labels → {len(new_filtered)} labels")
    filtered = new_filtered

final_p = p
final_k = k
final_labels = len(filtered)
print(f"\nFinal P/K scheme: P={final_p}, K={final_k}, {final_labels} labels")

# Step 3: Simulate greedy packing
draws = {lab: count // final_k for lab, count in filtered.items()}
schedule = []
while True:
    available = sorted(
        [(lab, d) for lab, d in draws.items() if d > 0],
        key=lambda x: -x[1],
    )
    if len(available) < final_p:
        break
    batch_labels = [lab for lab, _ in available[:final_p]]
    schedule.append(batch_labels)
    for lab in batch_labels:
        draws[lab] -= 1

scheduled_samples = len(schedule) * final_p * final_k
total_samples_in_groups = sum(label_counts[lab] for lab in filtered.keys())
unused_samples = total_samples_in_groups - scheduled_samples
efficiency = scheduled_samples / total_samples_in_groups * 100

print(f"\nScheduling results:")
print(f"  Batches scheduled: {len(schedule)}")
print(f"  Scheduled samples: {scheduled_samples}")
print(f"  Total samples in groups: {total_samples_in_groups}")
print(f"  Unused samples: {unused_samples} ({100 - efficiency:.1f}%)")
print(f"  Efficiency: {efficiency:.1f}%")

# Show remaining draws per label (leftover capacity)
print(f"\nRemaining draws per label (after greedy packing):")
remaining_draws = [(lab, d) for lab, d in draws.items()]
remaining_draws.sort(key=lambda x: -x[1])
for lab, d in remaining_draws:
    if d > 0:
        print(f"  Label {lab}: {d} full K-draws left ({d * final_k} samples)")
```

</details>

```
Total samples: 5452
Total unique labels: 50

Label distribution:
  Min samples per label: 4
  Max samples per label: 962
  Mean samples per label: 109.0

Top 10 labels by sample count:
  Label 29: 962 samples
  Label 35: 464 samples
  Label 24: 421 samples
  Label 38: 363 samples
  Label 26: 276 samples
  Label 25: 274 samples
  Label 39: 218 samples
  Label 13: 217 samples
  Label 5: 207 samples
  Label 27: 191 samples

Bottom 10 labels by sample count:
  Label 16: 4 samples
  Label 6: 4 samples
  Label 42: 6 samples
  Label 47: 8 samples
  Label 12: 9 samples
  Label 46: 9 samples
  Label 37: 9 samples
  Label 10: 10 samples
  Label 19: 11 samples
  Label 49: 11 samples

After filtering (>= 2 samples): 50 labels kept
  Iteration 1: P=16, K=2, num_labels=50
  → Converged!

Final P/K scheme: P=16, K=2, 50 labels

Scheduling results:
  Batches scheduled: 133
  Scheduled samples: 4256
  Total samples in groups: 5452
  Unused samples: 1196 (21.9%)
  Efficiency: 78.1%

Remaining draws per label (after greedy packing):
  Label 29: 348 full K-draws left (696 samples)
  Label 35: 99 full K-draws left (198 samples)
  Label 24: 77 full K-draws left (154 samples)
  Label 38: 48 full K-draws left (96 samples)
  Label 26: 5 full K-draws left (10 samples)
  Label 25: 4 full K-draws left (8 samples)
  Label 42: 1 full K-draws left (2 samples)
  Label 22: 1 full K-draws left (2 samples)
  Label 49: 1 full K-draws left (2 samples)
  Label 6: 1 full K-draws left (2 samples)
```

The issue is even bigger for `coarse_labels` instead of `fine_labels`:
```
Total samples: 5452
Total unique labels: 6

Label distribution:
  Min samples per label: 86
  Max samples per label: 1250
  Mean samples per label: 908.7

Top 10 labels by sample count:
  Label 1: 1250 samples
  Label 3: 1223 samples
  Label 2: 1162 samples
  Label 5: 896 samples
  Label 4: 835 samples
  Label 0: 86 samples

Bottom 10 labels by sample count:
  Label 0: 86 samples
  Label 4: 835 samples
  Label 5: 896 samples
  Label 2: 1162 samples
  Label 3: 1223 samples
  Label 1: 1250 samples

After filtering (>= 2 samples): 6 labels kept
  Iteration 1: P=6, K=5, num_labels=6
  → Converged!

Final P/K scheme: P=6, K=5, 6 labels

Scheduling results:
  Batches scheduled: 17
  Scheduled samples: 510
  Total samples in groups: 5452
  Unused samples: 4942 (90.6%)
  Efficiency: 9.4%

Remaining draws per label (after greedy packing):
  Label 1: 233 full K-draws left (1165 samples)
  Label 3: 227 full K-draws left (1135 samples)
  Label 2: 215 full K-draws left (1075 samples)
  Label 5: 162 full K-draws left (810 samples)
  Label 4: 150 full K-draws left (750 samples)
```

So, perhaps we'll need a weighted sampling to make sure that we're not throwing away too many samples? However, more samples from a common label does mean less unique labels per batch, which may again result in worse performance. Or perhaps the same algorithm, but reducing `P` by 1 until 2 so that it can continue drawing batches, just with less samples per batch. For example in the coarse_label setting: it can still draw 150 batches with 5 unique labels, and then another 12 with 4 unique labels, and then another 53 with 3 unique labels, and lastly another 12 batches with 2 unique labels.

Perhaps:

<details><summary>Click for a potential GroupByLabelBatchSampler implementation with P-shrinking</summary>

```python
class GroupByLabelBatchSampler(DefaultBatchSampler):
    """
    Batch sampler that groups samples by label while ensuring each batch contains samples from
    multiple label classes, with at least 2 samples per class. Uses PK sampling: each batch draws
    ``K`` samples from each of ``P`` distinct labels, where ``P >= 2`` and ``K >= 2``.

    This sampler is the correct choice for losses that perform in-batch triplet mining, which
    require at least 2 distinct classes per batch with at least 2 examples each.

    Recommended for:
        - :class:`~sentence_transformers.losses.BatchAllTripletLoss`
        - :class:`~sentence_transformers.losses.BatchHardSoftMarginTripletLoss`
        - :class:`~sentence_transformers.losses.BatchHardTripletLoss`
        - :class:`~sentence_transformers.losses.BatchSemiHardTripletLoss`

    The name "GroupByLabel" refers to its fundamental operation of grouping samples by their
    label values while simultaneously enforcing label balance across batches.

    Args:
        dataset (Dataset): The dataset to sample from.
        batch_size (int): Number of samples per batch. Must be divisible by 2.
            The effective batch size is ``P * K`` where ``P = min(num_labels, batch_size // 2)``
            and ``K = batch_size // P``. This may be up to ``P - 1`` less than ``batch_size``
            when ``batch_size`` is not evenly divisible by ``P``.
        drop_last (bool): If True, drop the last incomplete batch if the dataset size
            is not divisible by the batch size.
        valid_label_columns (List[str], optional): List of column names to check for labels.
            The first column name from ``valid_label_columns`` found in the dataset will
            be used as the label column.
        generator (torch.Generator, optional): Optional random number generator for shuffling
            the indices.
        seed (int): Seed for the random number generator to ensure reproducibility. Defaults to 0.
    """

    def __init__(
        self,
        dataset: Dataset,
        batch_size: int,
        drop_last: bool,
        valid_label_columns: list[str] | None = None,
        generator: torch.Generator | None = None,
        seed: int = 0,
    ) -> None:
        super().__init__(
            dataset,
            batch_size=batch_size,
            drop_last=drop_last,
            valid_label_columns=valid_label_columns,
            generator=generator,
            seed=seed,
        )
        self.dataset = dataset

        if self.batch_size % 2 == 1:
            raise ValueError("The batch size for `GroupByLabelBatchSampler` must be divisible by 2.")

        labels = self._determine_labels_to_use(dataset, self.valid_label_columns)
        groups: dict[Any, list[int]] = defaultdict(list)
        for sample_idx, label in enumerate(labels):
            groups[label].append(sample_idx)

        # Iteratively compute P/K and filter labels that are too small.
        # Removing labels can increase K, which may disqualify more labels.
        filtered: dict[Any, list[int]] = {lab: idx for lab, idx in groups.items() if len(idx) >= 2}
        while True:
            num_labels = len(filtered)
            if num_labels < 2:
                raise ValueError(
                    "GroupByLabelBatchSampler requires at least 2 distinct labels with sufficient samples, "
                    f"but only {num_labels} label(s) qualified."
                )
            p = min(num_labels, self.batch_size // 2)
            k = self.batch_size // p
            new_filtered = {lab: idx for lab, idx in filtered.items() if len(idx) >= k}
            if len(new_filtered) == len(filtered):
                break
            filtered = new_filtered

        self.groups = dict(filtered)
        self.labels_per_batch = p
        self.samples_per_label = k

        # Pre-compute batch schedule via greedy packing: always pick the labels with the most
        # remaining draws. Uses dynamic P adjustment: continues scheduling even when fewer than P
        # labels remain (as long as at least 2 are available) to maximize sample usage.
        draws = {lab: len(idx) // k for lab, idx in self.groups.items()}
        self._schedule: list[list[Any]] = []
        while True:
            available = sorted(
                [(lab, d) for lab, d in draws.items() if d > 0],
                key=lambda x: -x[1],
            )
            if len(available) < 2:  # Can't form a valid batch with P >= 2
                break
            # Dynamic P: use min(ideal_p, available_labels)
            current_p = min(p, len(available))
            batch_labels = [lab for lab, _ in available[:current_p]]
            self._schedule.append(batch_labels)
            for lab in batch_labels:
                draws[lab] -= 1

        # Calculate scheduled samples accounting for dynamic P per batch
        scheduled_samples = sum(len(batch) * k for batch in self._schedule)
        total_samples = sum(len(idx) for idx in self.groups.values())
        self._has_remainder = scheduled_samples < total_samples

    @staticmethod
    def _determine_labels_to_use(dataset: Dataset, valid_label_columns: list[str] | None) -> list[Any]:
        for column_name in valid_label_columns or []:
            if column_name in dataset.column_names:
                return dataset[column_name]
        raise ValueError(
            f"None of the valid_label_columns {valid_label_columns} are in the dataset, "
            f"which only has these columns: {dataset.column_names}."
        )

    def __iter__(self) -> Iterator[list[int]]:
        if self.generator and self.seed is not None:
            self.generator.manual_seed(self.seed + self.epoch)

        labels = list(self.groups.keys())
        k = self.samples_per_label

        # Shuffle samples within each label
        shuffled: dict[Any, list[int]] = {}
        for label in labels:
            indices = self.groups[label]
            perm = torch.randperm(len(indices), generator=self.generator)
            shuffled[label] = [indices[i] for i in perm]

        # Shuffle batch order (not the label assignments within each batch)
        batch_order = torch.randperm(len(self._schedule), generator=self.generator)

        positions: dict[Any, int] = {label: 0 for label in labels}

        for idx in batch_order:
            batch_labels = self._schedule[idx]
            batch: list[int] = []
            for label in batch_labels:
                pos = positions[label]
                batch.extend(shuffled[label][pos : pos + k])
                positions[label] = pos + k
            yield batch

        if not self.drop_last:
            batch = []
            for label in labels:
                pos = positions[label]
                if pos < len(shuffled[label]):
                    batch.extend(shuffled[label][pos:])
            if batch:
                yield batch

    def __len__(self) -> int:
        n = len(self._schedule)
        if not self.drop_last and self._has_remainder:
            n += 1
        return n
```

</details>

* https://huggingface.co/tomaarsen/all-distilroberta-v1-trec-balanced-sampler-dynamic-p#training-logs (From this PR): 0.9354 triplet accuracy, 214 steps, gradual loss from 4.8486 -> 3.8720

Either way: I'd be happiest to fully throw away the existing `GroupByLabelBatchSampler` code and replace it with a balanced implementation that ensures P >= 2. I think we can get away with the backwards incompatibility (old scripts will have different performance) by recognizing that this is more of a notable bug fix than a new feature. Also because the old `GroupByLabelBatchSampler` doesn't seem to have any use.

- Tom Aarsen

MrLoh:
Thanks for taking a look and for running those experiments! I've reworked the PR based on your feedback:

**Replaced `GroupByLabelBatchSampler` in-place** -- deleted `BalancedLabelBatchSampler` and the `BALANCED_LABELS` enum entirely. `GROUP_BY_LABEL` now does PK sampling directly, so this is a clean bug fix with no new API surface.

**Implemented dynamic P shrinking** as you suggested. The greedy scheduler now reduces P as minority labels run out (down to P=2), which dramatically improves sample utilization:

| Dataset | Fixed P efficiency | Dynamic P efficiency |
|---------|-------------------|---------------------|
| TREC coarse (6 labels) | 9.4% | 99.3% |
| TREC fine (50 labels) | 52.0% | 88.5% |

**Minority labels are not left behind**: labels with >= 2 samples but fewer than K now get at least 1 scheduled draw into a proper multi-class batch. Their samples participate in triplet mining and influence the loss, rather than being silently dropped (with `drop_last=True`) or relegated to an unstructured remainder batch where they'd be the only label.

**Other changes in this update:**
- `batch_size < 4` now raises `ValueError` (previously `batch_size=2` would give P=1, producing single-class batches)
- Added efficiency regression tests using TREC-like distributions
- All tests and docs updated to reflect the in-place replacement

The implementation follows your suggested code closely -- same P/K iteration, same greedy schedule with dynamic P, same `__iter__` structure.

sentence_transformers/sampler.py
```diff
@@ -195,15 +197,14 @@ def __init__(
...
-            min(n * k, len(self.groups[lab])) for lab, n in label_draws_used.items()
-        )
+        scheduled_samples = sum(min(n * k, len(self.groups[lab])) for lab, n in label_draws_used.items())
         total_samples = sum(len(idx) for idx in self.groups.values())
         self._has_remainder = scheduled_samples < total_samples
 
     @staticmethod
     def _determine_labels_to_use(dataset: Dataset, valid_label_columns: list[str] | None) -> list[Any]:
-        for column_name in valid_label_columns or []:
+        valid_label_columns = valid_label_columns or ["label", "labels"]
```
tomaarsen: Can we keep the default as `[]`? It shouldn't be an issue really, the `valid_label_columns` are grabbed from the data_collator, which defaults to `["label", "labels", "score", "scores"]` and although it can technically be overridden, it's essentially never touched.

examples/sentence_transformer/training/other/training_batch_hard_trec.py
```diff
@@ -133,7 +133,7 @@ def triplets_from_labeled_dataset(dataset):
     num_train_epochs=num_epochs,
     per_device_train_batch_size=train_batch_size,
     warmup_ratio=0.1,
-    # Use GROUP_BY_LABEL batch sampler for triplet losses that require multiple samples per label
+    # GROUP_BY_LABEL ensures each batch has P distinct labels with K samples each (P >= 2, K >= 2)
```
tomaarsen:
```suggestion
    # GROUP_BY_LABEL ensures each batch has 2+ distinct labels with 2+ samples each
```
A bit simpler to understand for people

Or perhaps

```suggestion
    # GROUP_BY_LABEL ensures each batch has at least 2 distinct labels with at least 2 samples per label
```

sentence_transformers/sampler.py
```diff
@@ -78,10 +78,26 @@ def __init__(
...
+    To maximise sample usage with imbalanced label distributions, the scheduler dynamically
+    reduces ``P`` as minority labels are exhausted. As long as at least 2 labels still have
+    unused samples, batches keep being produced -- just with fewer labels and therefore a
+    smaller effective batch size. This avoids discarding samples from majority labels that
+    would otherwise be left behind once all minority labels are consumed.
+
+    Labels with at least 2 samples but fewer than ``K`` are still scheduled into a proper
+    multi-class batch (contributing all their available samples). This ensures minority
+    labels participate in in-batch triplet mining and actually influence the loss, rather
+    than being relegated to a remainder batch where they may be the only label present.
```
tomaarsen:
```suggestion
    Batch sampler that ensures each batch contains samples from multiple distinct labels,
    suitable for losses that perform in-batch triplet mining.

    Each batch is constructed by drawing ``K`` samples (``K >= 2``, "samples per label") from
    each of ``P`` distinct labels (``P >= 2``, "labels per batch"), giving an effective batch
    size of ``P * K``. The initial values are computed as:

    - ``P = min(num_labels, batch_size // 2)``, use as many labels as possible
    - ``K = batch_size // P``, fill each label's slot equally

    To maximise sample usage with imbalanced label distributions, the scheduler dynamically
    reduces ``P`` as minority labels are exhausted. As long as at least 2 labels still have
    unused samples, batches keep being produced, just with fewer labels and therefore a
    smaller effective batch size. This avoids discarding samples from majority labels that
    would otherwise be left behind once all minority labels are consumed.

    Labels with at least 2 samples but fewer than ``K`` are still scheduled into a proper
    multi-class batch (contributing all their available samples). This ensures minority
    labels participate in in-batch triplet mining and actually influence the loss, rather
    than being relegated to a remainder batch where they may be the only label present.
```

tomaarsen:
I ran some tests by adding a `breakpoint()` in `BatchAllTripletLoss`, and I saw that some batches are only e.g. 4 items samples with 4 unique labels, i.e. still broken.

I tried some options, but because I want to ensure that each batch is still `batch_size` (except the last one if `drop_last` is False), I think we should consider simplifying again. I came up with this solution with pairs that get added in a round-robin fashion. I already pushed it (although we can always revert if needed).

It seems to work pretty nicely, the model does outperform all previous:
* https://huggingface.co/tomaarsen/all-distilroberta-v1-trec-group-by-label-b3d9cd1#training-logs (From this PR): 0.9495 triplet accuracy, 140 steps, gradual loss from 4.7867 -> 3.7847

What do you think? 

- Tom Aarsen

MrLoh:
Love it when an improvement actually manages to simplify the code! This looks much better then previous approach.

sentence_transformers/sampler.py
```diff
@@ -129,8 +117,6 @@ def __init__(
         generator: torch.Generator | None = None,
         seed: int = 0,
     ) -> None:
-        if generator is None:
-            generator = torch.Generator()
```
MrLoh: This was added to address this bugbot concern:

> `GroupByLabelBatchSampler.__iter__` only applies seed when `self.generator` is set, but still uses randomness (`torch.randperm`) even when `self.generator` is None. This makes the `seed` argument ineffective in the common case where no generator is passed, reducing reproducibility and potentially changing training behavior between runs.

@tomaarsen your call, but the concern seems legit to me

sentence_transformers/sampler.py
```diff
@@ -153,7 +153,7 @@ def __init__(
 
     @staticmethod
     def _determine_labels_to_use(dataset: Dataset, valid_label_columns: list[str] | None) -> list[Any]:
-        valid_label_columns = valid_label_columns or ["label", "labels"]
+        valid_label_columns = valid_label_columns or []
```
MrLoh: This had bee introduced to address this bugbot concern:

> `GroupByLabelBatchSampler._determine_labels_to_use` raises when `valid_label_columns` is `None`/empty, even if the dataset contains a label column. If `SentenceTransformerTrainer` passes `valid_label_columns=None` (or forgets to pass it), triplet finetuning can fail at sampler construction with a `ValueError`.

@tomaarsen your call, but the concern seems legit to me as evidenced by the test you needed to fix after removing this

sentence_transformers/sampler.py
MrLoh: I think you were right to remove this, this since the batch sampler is only constructed by a trainer typically

sentence_transformers/sampler.py
MrLoh: I think you were right to remove this, this since the batch sampler is only constructed by a trainer typically

MrLoh:
There was one other small issue that I pushed a fix for (formatting on that commit should be fixed now)

sentence_transformers/sampler.py
tomaarsen: Yeah, the sampler is never used standalone, so it's mostly fine. But formally, it would be nicer to have every batch sampler create this generator if it's None, but then it's best to add it for all batch samplers, and then it might be confusing to include it in this PR.

sentence_transformers/sampler.py
```diff
@@ -194,7 +194,7 @@ def __iter__(self) -> Iterator[list[int]]:
 
     def __len__(self) -> int:
         n = self._stream_length // self.batch_size
-        if not self.drop_last and self._stream_length % self.batch_size > 0:
+        if not self.drop_last and self._stream_length % self.batch_size >= 4:
```
tomaarsen, line 197: Very well spotted I think, I had to add the >= 4 to make sure that I'm not yielding a partial batch with 1 label (some losses might hypothetically crash on the final step then), but didn't realize that this should also be used here.

tomaarsen: Reproduced the previous issue, and this solves it. Great job

tomaarsen:
Thanks a bunch for helping with this! It'll be included in the next release.

- Tom Aarsen

PR #3673, opened 2026-02-24
[`feat`] Add generic NanoEvaluator abstractions with NanoBEIR backward compatibility

hotchpotch:
Hello!

## Summary

This PR introduces a generic `NanoEvaluator` abstraction for sampled information-retrieval evaluation, while preserving backward compatibility for existing NanoBEIR evaluators.

`NanoBEIREvaluator` is very useful for large IR benchmarks because it evaluates sampled subsets with much lower cost. In the same spirit, this PR extends the approach beyond NanoBEIR so that other datasets with the same `corpus/queries/qrels` structure can be evaluated with the same evaluator family.

For example, this enables evaluation on datasets such as:
- https://huggingface.co/datasets/hotchpotch/NanoMLDR
- https://huggingface.co/datasets/hotchpotch/NanoMIRACL
- https://huggingface.co/datasets/hotchpotch/NanoCodeSearchNet

## Details

- Add `NanoEvaluator` as a generic parent evaluator for dense retrieval.
- Refactor `NanoBEIREvaluator` to subclass `NanoEvaluator` while keeping its public API and output key conventions.
- Add `CrossEncoderNanoEvaluator` and `SparseNanoEvaluator` as generic parents for cross-encoder and sparse settings.
- Keep `CrossEncoderNanoBEIREvaluator` and `SparseNanoBEIREvaluator` as NanoBEIR-specific wrappers with backward-compatible behavior.
- Add/expand examples for dense, sparse, and cross-encoder usage.

## Backward Compatibility

- Existing `NanoBEIREvaluator` usage remains valid.
- Existing `CrossEncoderNanoBEIREvaluator` and `SparseNanoBEIREvaluator` usage remains valid.
- Metric naming and expected primary metrics are preserved for NanoBEIR evaluators.
- I also ran the sample code in NanoBEIREvaluator.py and confirmed that the resulting values remained unchanged from previous behavior.

---

This is an initial implementation to make the idea concrete. I would greatly appreciate any feedback!

### DIFF
--- a/sentence_transformers/cross_encoder/evaluation/nano_beir.py
+++ b/sentence_transformers/cross_encoder/evaluation/nano_beir.py
@@ -41,42 +41,149 @@ DATASET_NAME_TO_HUMAN_READABLE: dict[str, str] = {
 
 
 class CrossEncoderNanoBEIREvaluator(CrossEncoderNanoEvaluator):
-    """Evaluate a CrossEncoder model on NanoBEIR datasets.
+    """
+    This class evaluates a CrossEncoder model on the NanoBEIR collection of Information Retrieval datasets.
 
-    This evaluator is designed for reranking over NanoBEIR datasets. Rather
-    than reranking the full corpus, it reranks only ``rerank_k`` first-stage
-    candidates (``bm25`` by default), and reports MAP / MRR@k / nDCG@k.
+    The collection is a set of datasets based on the BEIR collection, but with a significantly smaller size, so it can
+    be used for quickly evaluating the retrieval performance of a model before committing to a full evaluation.
+    The datasets are available on Hugging Face in the `NanoBEIR collection <https://huggingface.co/collections/sentence-transformers/nanobeir-datasets>`_.
+    This evaluator will return the same metrics as the :class:`~sentence_transformers.cross_encoder.evaluation.CrossEncoderRerankingEvaluator`
+    (i.e., MRR@k, nDCG@k, MAP), for each dataset and on average.
 
-    This class preserves the NanoBEIR short-name API while delegating shared
-    reranking dataset mechanics to
+    Rather than reranking all documents for each query, the evaluator will only rerank the ``rerank_k`` documents from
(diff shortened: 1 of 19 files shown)


tests/evaluation/test_nano_evaluator.py
```diff
@@ -0,0 +1,232 @@
+from __future__ import annotations
```
hotchpotch, line 1: These tests were added to cover the existing NanoBEIR behavior first, so that NanoBEIR could be refactored without breaking that behavior.

sentence_transformers/cross_encoder/evaluation/nano_beir.py
```diff
@@ -104,7 +98,7 @@ class CrossEncoderNanoBEIREvaluator(SentenceEvaluator):
             the positives that are already in the documents list. Always set to True if your ``samples`` contain ``negative``
             instead of ``documents``. When using ``documents``, setting this to True will result in a more useful evaluation
             signal, but setting it to False will result in a more realistic evaluation. Defaults to True.
-        batch_size (int): Batch size to compute sentence embeddings. Defaults to 64.
+        batch_size (int): Batch size to compute sentence embeddings. Defaults to 32.
```
hotchpotch: The default batch size has long been 32 in the implementation, but the documentation still said 64, so I corrected the docstring to match the  actual behavior.

hotchpotch:
After reviewing the changes again, I realized that although the new files are mostly copies of existing implementations, the diff shows them as pure additions.

That makes it very hard to distinguish what was already in the existing implementation from what is actually new and worth focusing on in review, so the review cost feels quite high.

I’m going to move this PR back to draft for now and try to restructure the changes so the diffs are easier to read and the review cost is lower.

hotchpotch:
Hi!

After taking another look at this PR, I realized that this NanoEvaluator implementation has become too large and too difficult to follow in its current form to be a good candidate for merging.

Because of that, I’m going to close this draft PR.

I’m sorry if you had already spent time looking through the code. In that case, I may have taken up some of your time with a PR that was not yet in a review-friendly state.

Thank you for your time, and sorry again for the noise.

tomaarsen:
All good, I appreciate your time on this!

PR #3678, opened 2026-02-27
[`fix`] Patch AdaptiveLayerLoss / Matryoshka2dLoss with DDP

tomaarsen:
Hello!

## Pull Request overview
* Fix `AdaptiveLayerLoss` (and by extension `Matryoshka2dLoss`) compatibility with DDP training

## Details
When training with DDP, the `SentenceTransformer` model gets wrapped by `DistributedDataParallel`, which doesn't support `[0]` indexing. This caused `AdaptiveLayerLoss.forward` to crash when patching and restoring the transformer's forward function. The fix unwraps the DDP wrapper via `.module` before indexing: `base_model = self.model.module if hasattr(self.model, "module") else self.model`.

This patch was proposed by @NohTow and I've not had a chance to test it yet.

- Tom Aarsen


tomaarsen:
I can't reproduce this issue, I'm able to train with Matryoska2d just fine on multi-GPU.

- Tom Aarsen

Issue #3682, opened 2026-03-05
IndexError: list index out of range


zkailinzhang:
query_embeddings = model_emd.encode([response])
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/zhangkailin/.conda/envs/easyr1/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/zhangkailin/.conda/envs/easyr1/lib/python3.12/site-packages/sentence_transformers/SentenceTransformer.py", line 1062, in encode
    features = self.tokenize(sentences_batch, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/zhangkailin/.conda/envs/easyr1/lib/python3.12/site-packages/sentence_transformers/SentenceTransformer.py", line 1621, in tokenize
    return self[0].tokenize(texts, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/zhangkailin/.conda/envs/easyr1/lib/python3.12/site-packages/sentence_transformers/models/Transformer.py", line 309, in tokenize
    batch2.append(text_tuple[1])
                  ~~~~~~~~~~^^^
IndexError: list index out of range

Samoed:
Can you share full code to reproduce an error?

tomaarsen:
Specifically, is `response` a string, or a list/tuple? I feel like it might be a list of length 1, and then you're accidentally passing a nested list to `encode` instead of just a string or a single list.

Consider
```python
query_embeddings = model_emd.encode(response)
```
or if you know it's always a query:
```python
query_embeddings = model_emd.encode_query(response)
```
This method just calls `encode`, but uses a query prompt if it's defined in the model configuration.

- Tom Aarsen

PR #3685, opened 2026-03-11
fix: handle empty tuples in tokenize method

Jah-yee:
## Summary

Fixes IndexError when `texts` contains tuples with less than 2 elements in the `tokenize` method of `Transformer.py`.

## Problem

When calling `model.encode()` with input containing empty tuples or single-element tuples, the code at line 378-379 attempts to access `text_tuple[1]` without checking if it exists:

```python
for text_tuple in texts:
    batch1.append(text_tuple[0])
    batch2.append(text_tuple[1])  # IndexError here
```

This causes:
```
batch2.append(text_tuple[1])
IndexError: list index out of range
```

## Solution

Added a length check before accessing `text_tuple[1]`:

```python
for text_tuple in texts:
    batch1.append(text_tuple[0])
    if len(text_tuple) > 1:
        batch2.append(text_tuple[1])
```

## Testing

The fix has been verified to handle:
- Empty tuples `()`
- Single-element tuples `(str,)`
- Normal two-element tuples `(str, str)` (existing behavior preserved)

## Related Issue

Addresses: #3682

### DIFF
--- a/sentence_transformers/models/Transformer.py
+++ b/sentence_transformers/models/Transformer.py
@@ -376,7 +376,8 @@ class Transformer(InputModule):
             batch1, batch2 = [], []
             for text_tuple in texts:
                 batch1.append(text_tuple[0])
-                batch2.append(text_tuple[1])
+                if len(text_tuple) > 1:
+                    batch2.append(text_tuple[1])
             to_tokenize = [batch1, batch2]
 
         # strip


tomaarsen:
Hello!

Thank you for your PR! I'm afraid I'm likely not going to make this change now, as I'm wary of changes to `Transformer` at the moment. It will be completely overhauled in #3554, which is planned to be the core of the next minor release. In that PR, the `tokenize` method will be fully replaced with a `preprocess` method, and none of the prior `tokenize` code will make a return.

- Tom Aarsen

PR #3692, opened 2026-03-19
fix: use caller's class name as default model_type for subclasses

### LINKED ISSUE #3536: Wrong defaults used when loading older non-mean-pooled models via subclass
causes silently wrong inference

problem is only in v5. here's an example w/ bge, which [outputs the CLS token](https://huggingface.co/BAAI/bge-small-en-v1.5/blob/main/1_Pooling/config.json)

to repro:

```python


class MySentenceTransformer(SentenceTransformer): ...


model = MySentenceTransformer("BAAI/bge-small-en-v1.5")
#  v5 output: No sentence-transformers model found with name BAAI/bge-small-en-v1.5. Creating a new one with mean pooling.
# <v5 output: (nothing)

print(model.encode("test")[0])
#  v5 output: -0.30347106
# <v5 output: -0.030682037

for module in model:
    if isinstance(module, Pooling):
        print(module)
#  v5 output: Pooling({'word_embedding_dimension': 384, 'pooling_mode_cls_token': False, 'pooling_mode_mean_tokens': True, 'pooling_mode_max_tokens': False, 'pooling_mode_mean_sqrt_len_tokens': False, 'pooling_mode_weightedmean_tokens': False, 'pooling_mode_lasttoken': False, 'include_prompt': True})
# <v5 output: Pooling({'word_embedding_dimension': 384, 'pooling_mode_cls_token': False, 'pooling_mode_mean_tokens': True, 'pooling_mode_max_tokens': False, 'pooling_mode_mean_sqrt_len_tokens': False, 'pooling_mode_weightedmean_tokens': False, 'pooling_mode_lasttoken': False, 'include_prompt': True})
```

luckily we had a correctness test in CI, so this didn't cause problems. otherwise it would've silently failed

idk yet what the best fix is, but while debugging i think the problem comes from the default introduced in commit 227a369

AtharvaJaiswal005:
## Summary
- Use `self.__class__.__name__` instead of hardcoded `"SentenceTransformer"` as the default `model_type` in `_get_model_type()`
- Fixes subclassed `SentenceTransformer` loading older models (without `model_type` in config) with wrong pooling defaults

## Root Cause
`_get_model_type()` hardcoded the fallback to `"SentenceTransformer"` in two places:
1. When no `config_sentence_transformers.json` exists
2. When the config exists but has no `model_type` key

Meanwhile, `_model_config["model_type"]` is set to `self.__class__.__name__` (e.g. `"MySentenceTransformer"`). This mismatch caused the comparison at line 325 to fail, falling through to `_load_auto_model` instead of `_load_sbert_model`, resulting in wrong pooling defaults (mean pooling instead of CLS token).

## Reproduction
```python

class MySentenceTransformer(SentenceTransformer): ...

model = MySentenceTransformer("BAAI/bge-small-en-v1.5")
# Before fix: "No sentence-transformers model found... Creating a new one with mean pooling."
# After fix: loads correctly with CLS token pooling
```

## Changes
- `sentence_transformers/SentenceTransformer.py`: Changed default from `"SentenceTransformer"` to `self.__class__.__name__` in both fallback paths of `_get_model_type()`

Fixes #3536

### DIFF
--- a/sentence_transformers/SentenceTransformer.py
+++ b/sentence_transformers/SentenceTransformer.py
@@ -2525,8 +2525,8 @@ print(similarities)
         )
 
         if config_sentence_transformers_json_path is None:
-            return "SentenceTransformer"
+            return self.__class__.__name__
 
         with open(config_sentence_transformers_json_path, encoding="utf8") as fIn:
             config = json.load(fIn)
-            return config.get("model_type", "SentenceTransformer")  # Default to "SentenceTransformer" if not specified
+            return config.get("model_type", self.__class__.__name__)  # Default to caller's class if not specified


tomaarsen:
Hello!

Apologies for the delay. I was finishing up the v5.4 refactor.

Earlier in #3536 I suggested overriding `_get_model_type` to default to `self.__class__.__name__`, which is essentially what this PR does. During the refactor I moved that responsibility elsewhere, so the rationale has shifted: hardcoding `"SentenceTransformer"` is now deliberate. If there's no `config_sentence_transformers.json` (or no `model_type` in it) but there is a `modules.json`, it's very likely an old `SentenceTransformer` model, and the loader should be informed as such. That then routes through `_load_converted_modules`, where a subclass can take model-type-specific paths, e.g., recognize a `SentenceTransformer` checkpoint and load it with extra components on top, or see a `CrossEncoder` and load it entirely differently. By default `_load_converted_modules` just calls `_load_default_modules`, so non-subclassing users get the normal behavior.

So I think I'll close this, as the refactor changed this up a bit. It also resulted in a large merge conflict.

Tom Aarsen

PR #3719, opened 2026-04-12
Fix: infer_modality misclassifies 1D numpy string arrays as audio

### LINKED ISSUE #3718: Bug: `infer_modality` misclassifies 1D numpy string arrays as audio, raising `ValueError: Modality 'audio' is not supported`
https://github.com/huggingface/sentence-transformers/blob/25f0694290dbee841e39343551e82799e34648e6/sentence_transformers/base/modality.py#L571

## Description

`SentenceTransformer.encode()` raises `ValueError: Modality 'audio' is not supported by this SentenceTransformer model. Supported modalities: text` when the input is a 1D **numpy Unicode string array** (e.g. `dtype=<U28`), such as the output of `np.unique()` applied to a list of strings.

This regression was introduced in v5.4.0 with the multimodal `infer_modality` logic.

---

## Steps to Reproduce

```python
import numpy as np

model = SentenceTransformer("sentence-transformers/all-MiniLM-L6-v2")

# This is the problematic input shape — produced e.g. by np.unique()
texts = np.array(["Access Management", "Press Coordination", "Financial Reports"])
print(texts.dtype)  # dtype('<U21') — numpy Unicode string array, NOT float32

model.encode(texts)  # ← raises ValueError
```

**Output:**
```
ValueError: Modality 'audio' is not supported by this SentenceTransformer model. Supported modalities: text
```

---

## Root Cause

The bug is in [`sentence_transformers/base/modality.py`](https://github.com/huggingface/sentence-transformers/blob/main/sentence_transformers/base/modality.py) — the `infer_modality` function. The `match/case` block dispatches on the Python type `np.ndarray`, then infers modality from `ndim` alone, never checking the element dtype:

```python
case np.ndarray() | torch.Tensor():
    if sample.ndim in (1, 2):  # mono or multi-channel waveform
        return "audio"
    ...
```

A 1D numpy string array (`dtype=<U...`) has `ndim=1`, so it unconditionally returns `"audio"`.

---

## Proposed Fix

Check `dtype.kind` before dispatching on `ndim`:

```python
case np.ndarray() | torch.Tensor():
    if isinstance(sample, np.ndarray) and sample.dtype.kind in ("U", "S"):
        return "text"
    if sample.ndim in (1, 2):
        return "audio"
    ...
```

**Workaround today:** call `.tolist()` before passing to `encode()`.

---

## Environment

| | |
|---|---|
| `sentence-transformers` | 5.4.0 |
| Python | 3.12 |

idodoron11:
## Description

Fixes #3718.

`infer_modality` was routing 1D numpy string arrays (dtype kind `'U'` or `'S'`, produced by e.g. `np.array(list_of_strings)` or `np.unique()`) into the `np.ndarray | torch.Tensor` branch, which classifies 1D arrays as audio (mono waveform). This caused `SentenceTransformer.encode()` to raise:

```
ValueError: Modality 'audio' is not supported by this SentenceTransformer model. Supported modalities: text
```

## Root Cause

The `case np.ndarray() | torch.Tensor():` match in `infer_modality` only inspects `ndim`, so a 1D string array (e.g. from `np.unique()`) was indistinguishable from an audio waveform.

## Fix

Added an early `case` guard in `infer_modality` that checks `sample.dtype.kind in ("U", "S")` before the ndim-based branch and returns `"text"` for those arrays. Also updated the `InputFormatter.get_inputs` docstring to document this behaviour.

## Tests

Added regression tests in `tests/base/test_modality.py` covering:

- 1D Unicode string array (`dtype=<U...`) → `"text"`
- Result of `np.unique()` → `"text"`
- 1D byte-string array (`dtype=|S...`) → `"text"`
- Numeric 1D arrays (int16, uint8, float32, float64, complex64) still correctly route to `"audio"`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,7 @@ tmp
 .DS_Store
 /runs
 /tmp_trainer/
+
+# Lock files
+poetry.lock
+uv.lock
--- a/sentence_transformers/base/modality.py
+++ b/sentence_transformers/base/modality.py
@@ -196,7 +196,9 @@ class InputFormatter:
                 - tuple/list of mixed types: Non-text pairs (e.g. image + text)
                 - dict: Chat messages, audio data, or multimodal inputs
                 - PIL.Image.Image: Image inputs
-                - np.ndarray/torch.Tensor: Audio (1-2D) or video (3-5D) inputs
+                - np.ndarray/torch.Tensor: Audio (1-2D), image (3D), or video (4-5D) inputs.
+                  Numpy string arrays (dtype kind ``'U'`` or ``'S'``) are treated as text
+                  regardless of ndim.
 
         Returns:
             Tuple of (modality, processor_inputs_dict, extra_modality_kwargs) where:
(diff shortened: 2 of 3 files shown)


tomaarsen:
Superseded by #3720 as described in https://github.com/huggingface/sentence-transformers/issues/3718#issuecomment-4234573185

- Tom Aarsen

PR #3733, opened 2026-04-19
 Avoid OOM in compute_dataset_metrics for multimodal datasets with path columns

yjoonjang:
Hi @tomaarsen !

## Pull Request overview

* Skip `model.preprocess` in `compute_dataset_metrics` for string columns that hold media file paths, to prevent CPU RAM exhaustion during `Trainer.__init__` on multimodal datasets.

### The Issue
For every string-typed column, `BaseModelCardData.compute_dataset_metrics` calls `model.preprocess(dataset[:1000][column], task="document")` to produce token-length stats. 
On multimodal models, `model.preprocess` auto-detects the modality from each sample's contents via `infer_modality`. As a result, a column whose values are `"/path/to/img.jpg"` strings ends up loading 1000 real images through the image processor — before any training step even runs.

### Impact
With a typical multimodal retrieval schema (`anchor`, `positive`, `negative_1..7` → 8 path-holding columns) this results in `1000 media loads × 8 columns × N DDP ranks`. In the run that I surfaced this (Qwen3-VL-Embedding-2B on 4-GPU DDP), RSS grew to ~60 GB per rank (~240 GB total) during trainer init, exhausting system RAM and spilling into swap long before the first step.

*(Note: Datasets using the typed `datasets.Image`/`Audio`/`Video` features are unaffected, as their first element is a decoder/PIL.Image rather than a string, bypassing the text branch. However, storing media as path strings is very common in large-scale multimodal retrieval pipelines (ColPali-style, VDR, etc.) because it keeps the Arrow files tiny and lets images live on disk.)*

### The Fix
The fix probes the modality of a single sample via `infer_modality` (passing `self.model.modalities` so text-only models don't misclassify file-path-looking text). For non-text columns, it emits a compact `"string (<modality> path)"` stat with the sample count instead of running the expensive preprocess call. Pure text columns and typed-media columns keep their existing behavior.

### Testing
Added `TestComputeDatasetMetricsStringMediaPath` with two tests:
1. Asserts that `model.preprocess` is **not** called for an image-path column, and the stat is rendered as `"string (image path)"` with the correct sample count.
2. A regression guard confirming plain-text columns still produce token-length stats. 

Full `tests/base/test_model_card.py` passes (79 passed, 3 skipped, no regressions).


tomaarsen:
Hello!

This was indeed a big issue, thank you! 
I built on your implementation to take the changes a step further (although I kept your fix identical with infer_modality on 1 -> skip if not text), I also updated to ever only sample 100 values (no need to do a full 1k I think), I now chunk the loading for memory savings, and add a modality row to all tables. At its core, it should still solve your issue, while ideally also helping others with similar issues (e.g. loading 1000 videos from a VideoDecoder at once to determine their avg length).

Please have a look at let me know what you think.

- Tom Aarsen

yjoonjang:
Thanks for the thorough extension ! 
This looks much nicer than my original scope.

I ran tests, and the full `tests/base/test_model_card.py` files all passed (87 passed, 3 skipped). And I confimed that the two `TestComputeDatasetMetricsStringMediaPath` regression tests I added still pass under the new `{dtype, modality, data}` format.

yjoonjang:
I just thought one follow-up idea: Since your new `_compute_column_stats` already surfaces resolution/duration stats for `datasets.Image` / `Audio` / `VideoDecoder` columns, we could mirror the same behavior for string columns that `infer_modality` classifies as image/audio/video paths. 
That way the model card looks consistent regardless of whether the media is stored as a typed feature or as a file path string.

The access pattern stays bounded-memory because the underlying APIs are all lazy metadata reads, not full decodes:

| API | Returns | Peak RAM (my quick check) |
|-----|---------|----------------------------|
| `PIL.Image.open(path).size` | `(width, height)` | ~9 KB vs 78 KB full decode |
| `soundfile.info(path)` | `.duration` / `.samplerate` / `.channels` / `.frames` | ~1.5 KB vs 3.5 MB full decode |
| `av.open(path).streams.video[0]` | `.width` / `.height` / `.average_rate` / `.duration` | ~3.5 KB vs 73 KB full decode |

So at the 100-sample cap with `stat_chunk_size=8`, resolution/duration stats stay under ~1 MB per column even before GC.

Just thought it would round out the memory-bounded stats nicely.

tomaarsen:
I'm not 100% sure that'll work as we'd hope. A lot of the URL/path loading functionality is in `tranformers`, behind the processor call. So, we either:
1. have to call the underlying `transformers` processor and it gets a bit slow
2. have to copy a bunch of functionality from `transformers` surrounding URL/path loading

And neither sound particularly appealing. If we can do it more easily, that would be great, though.

- Tom Aarsen

yjoonjang:
Makes sense. I agree it's not worth adding a `transformers` dependency or duplicating code just to make the model card stats match. I think we can wrap up here.
Thank you!

PR #3734, opened 2026-04-19
docs(SimCSE): migrate README example to SentenceTransformerTrainer

MukundaKatta:
## Summary

Refs #3621.

Updates the SimCSE README code snippet to use the `SentenceTransformerTrainer` API (recommended since v3) instead of the soft-deprecated `model.fit()` call. Keeps the example minimal and in the same spirit as the original — train on pairs of the same sentence so in-batch negatives drive contrast.

Specifically:

- Replaces `InputExample` + `DataLoader` + `model.fit(...)` with `datasets.Dataset.from_dict(...)` + `SentenceTransformerTrainingArguments` + `SentenceTransformerTrainer.train()`.
- Uses `model.save_pretrained(...)` (the new API) instead of `model.save(...)`.
- Sets `save_strategy="no"` so the tiny example does not emit intermediate checkpoint dirs.
- Fixes a stale comment that said "CLS pooling": `Pooling(embedding_dim)` defaults to `pooling_mode="mean"`, which is also the mode the ablation table later in the README highlights.

Style follows the already-migrated [`train_askubuntu_simcse.py`](examples/sentence_transformer/unsupervised_learning/SimCSE/train_askubuntu_simcse.py) in the same directory.

## Testing

- `python -c "import ast; ast.parse(open(...).read())"` against the extracted fenced python block — parses cleanly.
- Did not execute the training example (requires downloading `distilbert/distilroberta-base` and a GPU to be meaningful); changes are confined to the README code block and mirror the pattern used in sibling training scripts already in the repo.

### DIFF
--- a/examples/sentence_transformer/unsupervised_learning/SimCSE/README.md
+++ b/examples/sentence_transformer/unsupervised_learning/SimCSE/README.md
@@ -11,12 +11,17 @@ The idea is to encode the same sentence twice. Due to the used dropout in transf
 SentenceTransformers implements the [MultipleNegativesRankingLoss](../../../../docs/package_reference/sentence_transformer/losses.md#multiplenegativesrankingloss), which makes training with SimCSE trivial:
 
 ```python
-from sentence_transformers import SentenceTransformer, InputExample
-from sentence_transformers.sentence_transformer.modules import Transformer, Pooling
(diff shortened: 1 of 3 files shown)


tomaarsen:
Hello!

Thanks for tackling this! I had another quick look and found some slight issues with the existing scripts/README, so I grouped those small edits in this PR as well. I appreciate it!

- Tom Aarsen

PR #3753, opened 2026-05-06
[feat] Per-call processing_kwargs override in Transformer.preprocess

tomaarsen:
Hello!

## Pull Request overview
* Add per-call `processing_kwargs` override to `Transformer.preprocess`
* Allowlist it through `encode`/`predict` and document it on the entry points

## Details
A dependent project wanted to pass different `processing_kwargs` on each call rather than only at construction time. I've added a `processing_kwargs` parameter to `Transformer.preprocess` that merges on top of the instance-level `self.processing_kwargs` with shallow per-modality semantics, so individual settings (e.g. only `max_length`) can be overridden without replacing the whole modality dict.

The merged dict is also threaded through `_call_processor` to `_process_chat_messages`, since the `chat_template` slot was read directly from `self.processing_kwargs` rather than via the existing `modality_kwargs`/`common_kwargs` plumbing. Without this, per-call `{"chat_template": {...}}` overrides would silently no-op. `SentenceTransformer.encode` and `SparseEncoder.encode` now allowlist `processing_kwargs` alongside `task`; `CrossEncoder.predict` and `BaseModel.preprocess` already forward `**kwargs` through.

cc @NohTow 

- Tom Aarsen

### DIFF
--- a/tests/sentence_transformer/test_model.py
+++ b/tests/sentence_transformer/test_model.py
@@ -1208,6 +1208,24 @@ def test_encode_with_dataset_column(stsb_bert_tiny_model: SentenceTransformer) -
     assert embeddings.shape == (2, model.get_embedding_dimension())
 
 
+def test_encode_per_call_processing_kwargs(stsb_bert_tiny_model: SentenceTransformer) -> None:
+    """Per-call ``processing_kwargs`` should be accepted by ``encode`` and reach ``preprocess``.
+
+    If the kwarg were silently dropped (or rejected by encode's allowlist), truncated and full
+    encodings of the same text would either raise or produce identical embeddings.
+    """
+    model = stsb_bert_tiny_model
+    text = "this sentence is much longer than four tokens for sure"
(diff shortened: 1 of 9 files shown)


tests/sentence_transformer/test_model.py
```diff
@@ -1208,6 +1208,24 @@ def test_encode_with_dataset_column(stsb_bert_tiny_model: SentenceTransformer) -
...
+    If the kwarg were silently dropped (or rejected by encode's allowlist), truncated and full
+    encodings of the same text would either raise or produce identical embeddings.
+    """
+    model = stsb_bert_tiny_model
+    text = "this sentence is much longer than four tokens for sure"
+    truncated = model.encode(
+        [text],
+        convert_to_tensor=True,
+        processing_kwargs={"text": {"max_length": 4, "truncation": True}},
+    )
```
tomaarsen, line 1223: This is how it can be used for actual models from this PR onwards.

PR #3769, opened 2026-05-11
docs: fix grammar in parallel-sentence-mining README

Karthikkolli17:
Grammar fixes in the bitext mining example README:

1. "an set" → "a set"
2. "an example sentences" → "an example sentence" (two occurrences, kept the German translation match consistent)

Spotted it while reading examples.


tomaarsen:
Well spotted, thank you!

- Tom Aarsen

PR #3771, opened 2026-05-12
[`examples`] Avoid LoggingHandler, silence httpx in examples

tomaarsen:
Hello!

## Pull Request overview
* Avoid LoggingHandler, it's a bit unnecessary
* Silence httpx in examples

## Details
LoggingHandler doesn't really need to be used by users, they can roll their own logging tooling if they wish. Beyond that, we should silence httpx INFO logs because they're really noisy.

- Tom Aarsen


wadeKeith (commented):
Withdrawn; this review was too cursory.

Issue #2689, opened 2024-05-29
[feature request] Pre-training, GPL, LLM2Vec


dmacko232:
Hello @tomaarsen,

I hope u are doing well. I love the new Sentence Transfomers release!

I am writing to propose a few feature requests that I believe would enhance Sentence Transformers:
- Add examples of more pre-training approaches such as [RetroMAE and DupMAE](https://github.com/staoxiao/RetroMAE) which have shown SOTA performance when pre-training for retrieval. It would be nice to be able to easily use them in Sentence Transformers without relying on another library.
-  Example of fine-tuning using [GPL](https://github.com/UKPLab/gpl), as the training code changed and it could be done more efficiently, compactly, and directly withing Sentence Transformers instead of relying on another library. Additionally, this could enable pseudo-labeling using LLMs, such as GPT-4, instead of Cross-encoders. It would be also nice to decompose this in a way that it is possible to easily generate queries (GenQ) and mine hard negatives without the pseudo-labeling part (in case the pseudo-labeling is not desired).
- Add easy merging of fine-tuned and base model, inspired from [LM-Cocktail](https://github.com/FlagOpen/FlagEmbedding/tree/master/LM_Cocktail)
- Integration with [LLM2Vec](https://github.com/McGill-NLP/llm2vec)

austinmw:
Would also like to see these features!

tomaarsen:
Hello!

- [ ] I completely agree. RetroMAE/DupMAE are quite valuable and commonly used in research. They use a wholly different training approach than Sentence Transformers, but I think some examples & documentation on RetroMAE/DupMAE would be useful, e.g. very much like this [MLM example](https://github.com/UKPLab/sentence-transformers/blob/master/examples/unsupervised_learning/MLM/train_mlm.py). I'm definitely open to PRs here!
- [ ] I'm aware of GPL via the [documentation](https://sbert.net/examples/domain_adaptation/README.html#gpl-generative-pseudo-labeling), but have never used it myself. As a result, I'm not quite sure how challenging the implementation here would be. That said, for some of the components (e.g. hard negatives mining) it would be useful to have helper methods. And I agree that query generation with T5 is a bit dated nowadays, and Reranking should have some more flexibility as well. I'm also open to PRs for e.g. Negative Mining here, and for PRs on training examples in the Unsupervised Learning code (and the documentation): https://github.com/UKPLab/sentence-transformers/tree/master/examples/unsupervised_learning
[...]

PR #3776, opened 2026-05-13
[`feat`] Add model merging via mergekit (SentenceTransformer / SparseEncoder / CrossEncoder)

yjoonjang:
Hello!

# Pull Request overview
* Add `BaseModel.merge(...)`, a classmethod that combines several fine-tuned checkpoints into one model. The transformer body goes through [`mergekit`](https://github.com/arcee-ai/mergekit); the surrounding ST modules (`Pooling`, `Normalize`, `Dense`, `LayerNorm`, ...) are handled in-tree so the result is a usable `SentenceTransformer` directory.
* Addresses the merge feature request in #2689 and follows the direction @tomaarsen suggested in [arcee-ai/mergekit#286](https://github.com/arcee-ai/mergekit/issues/286#issuecomment-2152200524) — a `mergekit + Sentence Transformers` integration that preserves the Pooling and other modules.

## Details

`BaseModel.merge` works for all three model classes (`SentenceTransformer`, `SparseEncoder`, `CrossEncoder`) since they all inherit from `BaseModel`. The transformer body is delegated to `mergekit`, so every method it supports (`linear`, `slerp`, `task_arithmetic`, `ties`, `dare_ties`, `dare_linear`, `breadcrumbs`, `della`, `model_stock`, `sce`, ...) is exposed for free. Around the body, we handle the ST module stack ourselves: validate that the inputs match, merge weight-bearing modules at the state-dict level, copy the stateless ones, and write the ST-specific top-level files mergekit doesn't know about (`modules.json`, `config_sentence_transformers.json`, `sentence_bert_config.json`). The result reloads cleanly with `cls(output_path)` and the per-class task head is preserved.

### Example usage

```python
SentenceTransformer.merge(
    models=["sentence-transformers/all-MiniLM-L6-v2",
            "sentence-transformers/paraphrase-MiniLM-L6-v2"],
    weights=[0.6, 0.4],
    method="linear",                # or "slerp", "ties", "task_arithmetic", ...
    base_model=None,                # required for delta-based methods
    output_path="merged-minilm/",
    dtype="float16",
    device="cpu",
)
```

Same signature for `SparseEncoder.merge(...)` and `CrossEncoder.merge(...)`.

### How it works

The merge runs in three phases:

| Phase | What happens |
|-------|--------------|
| **1. Validate** | All inputs must agree on `modules.json` — same classes, same order, same path. Per-module `config.json`s are compared after canonicalizing through the Module class (`load_config` → constructor → `get_config_dict`), so older saves that predate newer pooling keys (`pooling_mode_weightedmean_tokens`, `include_prompt`, ...) compare equal to newer saves with the same effective config.  If one target model has `modules.json` and others don't, the merge is rejected with a clear error; if all target models have (or all don't have) `modules.json`, the merge proceeds normally. `Asym` / `Router` and a `save_in_root=False` first module raise `NotImplementedError` up front; other non-standard module types (e.g. `BoW`, `StaticEmbedding`, `CLIPModel`) aren't explicitly rejected here but will fail later inside mergekit / ST loading. |
| **2. Body** | Build a mergekit `MergeConfiguration` from the user's args and `run_merge` it to a temp directory; then move the outputs into `output_path`. |
| **3. Other modules** | Weight-bearing modules (`Dense`, `LayerNorm`, `WeightedLayerPooling`) are merged state-dict-wise; stateless ones (`Pooling`, `Normalize`, `Dropout`, `SpladePooling`) are validated and copied from the first input. A subdirectory that's missing on disk (HF Hub doesn't track empty dirs — common for `Normalize`) or present-but-empty is treated as a stateless empty module, so we just `mkdir` the target and continue. Finally, any top-level sidecar from the first input that mergekit didn't write is copied over — that's the ST configs (`modules.json`, `config_sentence_transformers.json`, `sentence_bert_config.json`) and multimodal extras (`preprocessor_config.json`, `processor_config.json`, `chat_template.json`, ...).|

### Module dispatch table

| Module class | Has weights? | Merge strategy |
|---|:-:|---|
| `Transformer` / `MLMTransformer` (HF body) | yes | mergekit, any supported method |
| `Pooling` / `Normalize` / `Dropout` / `SpladePooling` | no | config-equality check + copy from first input |
| `Dense` / `LayerNorm` / `WeightedLayerPooling` | yes | state-dict: linear weighted average, or `task_arithmetic` deltas when `base_model` is given |
| `Asym` / `Router` | varies | rejected up front with a clear error — would need recursive merging of child modules |
| `BoW` / `StaticEmbedding` / `CLIPModel` and other non-standard module types | varies | not explicitly handled; merge will surface an error from mergekit / module loading |

### Default weights

The natural "uniform" default depends on the method:

| Method family | `weights=None` defaults to | Why |
|---|---|---|
| Blend (`linear`, `slerp`) | `[1/n] * n` | weights form a weighted average |
| Delta (`task_arithmetic`, `ties`, `dare_*`, `breadcrumbs`, `della`) | `[1.0] * n` | weights scale per-model deltas — `[0.5, 0.5]` would silently halve every delta |

The mergekit `parameters.normalize` flag is set per-method so user-supplied scaling carries through end-to-end for delta methods too.

### ST-side modules merge linearly
For weight-bearing ST modules (`Dense`, `LayerNorm`, `WeightedLayerPooling`), only `linear` and `task_arithmetic` have state-dict implementations. `task_arithmetic` further requires the `base_model` to also have the matching module on disk so we can read its tensors as the base; otherwise we fall back to linear here too. For `ties`/`dare_*`/`slerp` the body uses that method but these ST-side tensors are merged with a plain linear weighted average and a `WARNING` is logged. Most ST checkpoints don't have a Dense head (the common stack is body + Pooling + Normalize), so the practical footprint is small, but worth flagging. 

### mergekit 0.1.4 bugs hit during integration

Two upstream bugs are worked around in `_require_mergekit()`:

1. **Pydantic forward refs not resolved on `ConfiguredModuleArchitecture` / `ConfiguredModelArchitecture`.** First call raises `PydanticUserError: ... is not fully defined; ... call ConfiguredModuleArchitecture.model_rebuild()`. Reported as [arcee-ai/mergekit#681](https://github.com/arcee-ai/mergekit/issues/681) — still open, no maintainer reply. The fix is to import torch and call `model_rebuild()` once before invoking mergekit.

2. **[`mergekit/plan.py:172`](https://github.com/arcee-ai/mergekit/blob/main/mergekit/plan.py#L172) tuple/list concatenation `TypeError`.** `WeightInfo.aliases` is `Tuple[str, ...]` (the pydantic model needs to stay hashable), but `plan.py` does `[w_in.name] + (w_in.aliases or [])` and crashes. Only fires for architectures whose JSON uses `aliases` — basically BERT variants (`LayerNorm.beta`/`gamma` legacy aliases). I couldn't find related reports for this anywhere. The workaround pulls `plan_tensor`'s source via `inspect.getsource`, swaps `+` for `*` unpacking, and `exec`s it back into mergekit. Sentinel-gated and logs a `WARNING` so it's visible.

Both workarounds can be deleted once upstream is fixed.

### Reconciling mergekit's arch registry with the input checkpoints

mergekit ships a set of hand-crafted architecture JSON files (e.g. `bert-masked-lm.json`) that list which weights to merge. These can disagree with what's actually in the input checkpoints in two ways:

1. **Required weights that aren't always saved.** `bert-masked-lm.json` requires `bert.pooler.dense.{weight,bias}`, but `BertForMaskedLM` checkpoints commonly don't save the pooler. Mergekit aborts with `Tensor X required but not present`.
2. **Weights present in the checkpoint but missing from the arch JSON.** The same `bert-masked-lm.json` doesn't list `cls.predictions.transform.*`, so those tensors silently get dropped from the merged output.

Before calling `run_merge`, we scan the input safetensors keys and compare against the arch's listed weights:

* If any non-layer-templated input key isn't covered by the arch, we delete the arch entry from `NAME_TO_ARCH`, which makes mergekit fall back to its auto-inference path. Auto-inference lists every input tensor and merges them all, which is the safer default whenever the hand-crafted arch is incomplete.
* Otherwise (arch fully covers the inputs), we only mark required-but-absent weights as optional and keep the hand-crafted arch's tied-weights / aliases semantics.

### Compatibility / refusal cases (tested)

| Refusal | Trigger |
|---|---|
| Module count mismatch | inputs differ in length of `modules.json` |
| Module class mismatch | inputs have different class at the same index |
| Module config mismatch | e.g. one model has `pooling_mode="cls"`, another `"mean"` |
| Inputs disagree on `modules.json` presence | one input has it, another doesn't |
| Unsupported method | typo in `method=` |
| Missing `base_model` | delta method without `base_model` |
| Two-model method with 3+ models | wrong arity for `slerp`/`nuslerp`/`nearswap` |
| Two-model method with duplicate model paths | `len(set(models)) < 2` |
| `slerp` with `base_model` not in `models` | ambiguous |
| Single model | `len(models) < 2` |
| `output_path=None` | required argument |

## Validation

I ran a linear merging on a few embedding/reranker model pairs:

| Pair | Result | Notes |
|---|:-:|---|
| [`intfloat/multilingual-e5-large`](https://huggingface.co/intfloat/multilingual-e5-large) ⊕ [`nlpai-lab/KoE5`](https://huggingface.co/nlpai-lab/KoE5) | ✓ | Older e5 has fewer pooling-config keys than KoE5 (`pooling_mode_weightedmean_tokens`, `pooling_mode_lasttoken`, `include_prompt`. multilingual-e5-large is also missing the `2_Normalize/` directory entirely (empty dir not tracked on the Hub) — handled by the empty-stateless-module logic. |
| [`BAAI/bge-m3`](https://huggingface.co/BAAI/bge-m3) ⊕ [`nlpai-lab/KURE-v1`](https://huggingface.co/nlpai-lab/KURE-v1) | ✓ | Same older/newer pooling asymmetry, same missing-`2_Normalize/`. |
| [`yjoonjang/splade-ko-v1`](https://huggingface.co/yjoonjang/splade-ko-v1) ⊕ [`telepix/PIXIE-Splade-v1.5`](https://huggingface.co/telepix/PIXIE-Splade-v1.5) | ✓ | Both are full bert-base `ModernBertForMaskedLM ` with pooler weights present. End-to-end SparseEncoder merge works. |
| [`naver/splade-cocondenser-ensembledistil`](https://huggingface.co/naver/splade-cocondenser-ensembledistil) ⊕ [`tomaarsen/splade-cocondenser-ensembledistil-sts`](https://huggingface.co/tomaarsen/splade-cocondenser-ensembledistil-sts) | ✓ | Surfaced two issues: (a) one checkpoint has `embedding_dimension=None` in its `SpladePooling` config while the other has `30522` — that field is lazy-initialized at first forward pass, so the canonical-config validator now treats `None` as a wildcard; (b) mergekit's `bert-masked-lm.json` doesn't list `cls.predictions.transform.*`, so without intervention those weights would silently drop from the merged output — fixed by falling back to auto-inference when the arch's listed weights don't cover the inputs. |
| [`BAAI/bge-reranker-v2-m3`](https://huggingface.co/BAAI/bge-reranker-v2-m3) ⊕ [`dragonkue/bge-reranker-v2-m3-ko`](https://huggingface.co/dragonkue/bge-reranker-v2-m3-ko) | ✓ | Plain HF `AutoModelForSequenceClassification` — no `modules.json` on either. Handled by treating modules.json-less inputs as body-only. |
| [`Qwen/Qwen3-VL-Embedding-2B`](https://huggingface.co/Qwen/Qwen3-VL-Embedding-2B) ⊕ [`whybe-choi/Qwen3-VL-Embedding-2B-ko-vdr-preview-v0.2`](https://huggingface.co/whybe-choi/Qwen3-VL-Embedding-2B-ko-vdr-preview-v0.2) | ✓ | Multimodal (vision-language) embedding, ~2B params. Surfaced that mergekit only copies the tokenizer — multimodal sidecars (`preprocessor_config.json`, `processor_config.json`, `chat_template.{json,jinja}`, ...) need to come from the first input separately, or the merged dir won't reload. The copy step now picks up every top-level sidecar from the first input that mergekit didn't already write, while explicitly skipping weight files so the merged shards aren't shadowed. |

Companion examples live under `examples/sentence_transformer/training/other/model_merging.py`, `examples/cross_encoder/training/other/model_merging.py`, and `examples/sparse_encoder/training/other/model_merging.py`.

## Files changed

| File | Change |
|---|---|
| `sentence_transformers/base/merging.py` | new — core merge logic |
| `sentence_transformers/base/model.py` | added `BaseModel.merge` classmethod |
| `pyproject.toml` | added `merge = ["mergekit>=0.0.5"]` optional extra |
| `tests/test_merging.py` | new tests (per-method parametrize + refusal cases) |
| `examples/sentence_transformer/training/other/model_merging.py` | new example |
| `examples/cross_encoder/training/other/model_merging.py` | new example |
| `examples/sparse_encoder/training/other/model_merging.py` | new example |

## TODO / open questions

- [ ] File an upstream issue + 1-line PR for the `plan.py:172` tuple/list bug in mergekit; both workarounds in this PR can be dropped once mergekit ships a fix.
- [ ] Implement `ties`/`dare_*`/`slerp` for ST-side weight modules (`Dense`/`LayerNorm`/...) instead of always falling back to linear.
- [ ] Worth supporting mergekit's slice-based "frankenmerge" YAML pass-through (`SentenceTransformer.from_merge_config(yaml_path)`)?
- [ ] Cross-reference [arcee-ai/mergekit#286](https://github.com/arcee-ai/mergekit/issues/286) once this lands.

---

Note: Claude Code helped with parts of the implementation and drafted the test cases; everything's been manually reviewed and verified end-to-end. But there may be something I could have missed, so feel free to add/delete/update features.

Thank you!

- Youngjoon Jang


tomaarsen:
Hello!

Oh wow, I've tried some merging stuff myself, but never considered adding it as a method on the BaseModel, I think that's very clean. When I tried merging manually earlier, I could never get it to work very well, but I'm very curious to try it here with this. 

I'll read through your PR in more details in a bit.

- Tom Aarsen

whybe-choi:
Hello, @yjoonjang and @tomaarsen !

I think the lower bound for the `merge` extra should be derived from the mergekit interfaces and methods this PR exposes. I checked the currently published mergekit versions on PyPI: `0.0.5`, `0.0.5.1`, `0.0.6`, and `0.1.4`.

A few relevant differences:

1. Up to `mergekit==0.0.6`, the architecture code is still exposed as [`mergekit/architecture.py`](https://github.com/arcee-ai/mergekit/blob/v0.0.6/mergekit/architecture.py). The newer package layout referenced in this PR only exists in `0.1.4`, e.g. [`mergekit/architecture/base.py`](https://github.com/arcee-ai/mergekit/blob/v0.1.4/mergekit/architecture/base.py) and [`mergekit/architecture/json_definitions.py`](https://github.com/arcee-ai/mergekit/blob/v0.1.4/mergekit/architecture/json_definitions.py).

A quick repro with `mergekit==0.0.6`:

```python
from mergekit.architecture.base import ConfiguredModelArchitecture
from mergekit.architecture.json_definitions import NAME_TO_ARCH
```

raises:

```text
ModuleNotFoundError: No module named 'mergekit.architecture.base'; 'mergekit.architecture' is not a package
ModuleNotFoundError: No module named 'mergekit.architecture.json_definitions'; 'mergekit.architecture' is not a package
```

2. More importantly, this PR includes `arcee_fusion` and `karcher` in `SUPPORTED_METHODS`, but those methods are not implemented in `mergekit==0.0.6`.

The `get()` helper resolves methods from `REGISTERED_MERGE_METHODS` and raises `RuntimeError` if the method name is missing:

- [`mergekit==0.0.6`: `get()`](https://github.com/arcee-ai/mergekit/blob/v0.0.6/mergekit/merge_methods/__init__.py#L27-L30)
- [`mergekit==0.0.6`: `REGISTERED_MERGE_METHODS`](https://github.com/arcee-ai/mergekit/blob/v0.0.6/mergekit/merge_methods/registry.py#L133-L135)

In `0.0.6`, the registry includes methods such as `LinearMerge`, `SlerpMerge`, `NuSlerpMerge`, `SCEMerge`, and `NearSwapMerge`, but not `ArceeFusionMerge` or `KarcherMerge`:

- [`mergekit==0.0.6` registry imports/static methods](https://github.com/arcee-ai/mergekit/blob/v0.0.6/mergekit/merge_methods/registry.py#L18-L40)

In `0.1.4`, both are imported and registered:

- [`mergekit==0.1.4`: `ArceeFusionMerge` / `KarcherMerge` imports and registration](https://github.com/arcee-ai/mergekit/blob/v0.1.4/mergekit/merge_methods/registry.py#L6-L27)
- [`mergekit==0.1.4`: `REGISTERED_MERGE_METHODS`](https://github.com/arcee-ai/mergekit/blob/v0.1.4/mergekit/merge_methods/registry.py#L103-L105)

Repro with `mergekit==0.0.6`:

```python
from mergekit.merge_methods import get

get("arcee_fusion")
get("karcher")
```

raises:

```text
RuntimeError: Unimplemented merge method arcee_fusion
RuntimeError: Unimplemented merge method karcher
```

Both methods are available in `mergekit==0.1.4`.

So if we want to keep the current `SUPPORTED_METHODS` list as-is, I think the extra should be:

```toml
merge = ["mergekit>=0.1.4"]
```

pyproject.toml
```diff
@@ -52,6 +52,7 @@ image = ["transformers[vision]"]
 audio = ["transformers[audio]"]
 video = ["transformers[video]"]
 train = ["datasets>=2.0.0", "accelerate>=0.20.3"]
+merge = ["mergekit>=0.0.5"]
```
whybe-choi:
```suggestion
  merge = ["mergekit>=0.1.4"]
  ```

yjoonjang:
Good catch, thanks. The package layout and `arcee_fusion`/`karcher` registration only appeared in 0.1.4, so it definitely should be `>=0.1.4`.

tomaarsen:
Hello!

I've had a chance to review this over the last few days. In short: I have a few concerns:
1. The integration is very big, and we're relying on quite a lot of `mergekit` internals and version-specific code (`ConfiguredModelArchitecture`, `MergePlanner`, etc., but also e.g. the supported methods).
2. The `mergekit` bug with `[w_in.name] + (w_in.aliases or [])` really should not be in Sentence Transformers, especially as the fix uses `exec`. It should be upstreamed instead, but I'm not sure if mergekit still accepts PRs like these.
3. It looks like `mergekit` is quite restrictive with a few dependencies (`safetensors~=0.5.2`, `accelerate~=1.6.0`, `pydantic~=2.10.6`), which can prevent clean installs, e.g. if newer `transformers` wants more modern `accelerate` versions.

So, despite `BaseModel.merge` being a very clean place to support this merging, I'm wary that it'll be risky to integrate with `mergekit` like this: we're at risk if they update (concern 1) and if they don't (concerns 2 & 3). For now, I think I'm leaning to leaving this be, and perhaps revisiting it at a later time or implementing it as a separate library from sentence-transformers or something.

- Tom Aarsen

PR #3780, opened 2026-05-20
CLIP Notebooks Refresh

lbourdois:
@tomaarsen, as mentioned a couple weeks ago on Slack, I updated the 5 notebooks showing how to use CLIP with ST.  

Specifically:
- this now uses https://huggingface.co/datasets/sentence-transformers/unsplash-lite rather than having to download the images manually
- I pre-calculated some embeddings (https://huggingface.co/datasets/sentence-transformers/unsplash-lite/discussions/2 would have to be merged) to prevent the user from doing it as part of the tutorial
- I’ve added some examples of trimmed models for next week’s release.

The whole thing was executed in Google Colab to make sure it works well for people who don’t have a GPU.


tomaarsen:
Heya Loïck!

Thank you for your contribution here! I've made some edits here and there, most notably removing the TIPSv2 code as that model simply isn't nicely integrated in transformers or sentence transformers. Feel free to have a look, but I'm okay with the current status 🤗 

- Tom Aarsen

lbourdois:
Everything looks good to me for the notebooks.

I just want to point out that for `Image_Clustering.ipynb`, your run produces different image clusters than what I got in mine (we have exactly the same results for the other notebooks).
Since both you and I are using pre-computed embeddings, I would have expected the outputs to be the same.
The difference seems to stem from the `community_detection` function. I’m reusing the one from the old version of the notebook, whereas you’re using the one from the utility.
So I’m wondering if this is expected or if there might be an issue.

tomaarsen:
That's all good, I wanted to simplify the notebook to avoid people getting lost in the community detection implementation, so I used the pre-implemented one in the utils. I understand that you didn't make this change though, as there was already a community_detection function implemented.

I think this should be good then!

- Tom Aarsen

PR #3781, opened 2026-05-23
fix TSDAE weight tying on transformers v5

### LINKED ISSUE #3737: `DenoisingAutoEncoderLoss` and Transformers v5
I just stumbled into the following  message while using TSDAE : 
```
RuntimeError: Tying encoder and decoder weights is not currently supported for transformers version >= 5.0.0. Please either install transformers<5.0.0 or set tie_encoder_decoder=False.
```
Where the reason is : 
> TSDAE won't work nicely with transformers v5 with tie_encoder_decoder as it refactored weight ties. They're now a bit harder to do without all  weights being inside of the same PreTrainedModel instance. So, for now it's just not supported with an error to warn users.

https://github.com/huggingface/sentence-transformers/pull/3619

Is there an plan to make it work for transformers v5 ?

omkar-334:
`DenoisingAutoEncoderLoss` raised a `RuntimeError` on transformers >= 5.0.0 whenever `tie_encoder_decoder=True` (the default). Since tying is the recommended TSDAE setting, this effectively broke TSDAE on v5 unless you pinned an older transformers.  

| loss code | transformers 4.57.6 (<5) | transformers 5.9.0 (>=5) |
|---|---|---|
| before this PR | ✅ works | ❌ `RuntimeError` |
| with this PR | ✅ works | ✅ works |  

The loss tied the encoder and decoder through the private `PreTrainedModel._tie_encoder_decoder_weights`. transformers v5 removed that helper and moved tying to a per model `_tied_weights_keys` mapping that only applies within a single `PreTrainedModel`. TSDAE keeps the encoder and decoder as two separate models, so the native v5 machinery cannot bridge them, and #3619 added the `RuntimeError` as a stopgap.

Fixes #3737

### DIFF
--- a/sentence_transformers/sentence_transformer/losses/denoising_auto_encoder.py
+++ b/sentence_transformers/sentence_transformer/losses/denoising_auto_encoder.py
@@ -14,22 +14,22 @@ logger = logging.getLogger(__name__)
 
 
 def _tie_encoder_decoder_weights(encoder: nn.Module, decoder: nn.Module) -> None:
-    """Share the encoder's parameters with the identically-named decoder parameters.
+    """Tie the encoder's parameters to the decoder's: each encoder parameter is replaced by the
+    identically-named decoder parameter, so the two share storage and the decoder's values become
+    the shared initialization.
 
-    Replaces ``transformers.PreTrainedModel._tie_encoder_decoder_weights``, removed in
-    transformers 5.0.0 (which moved tying to an in-model ``_tied_weights_keys`` mapping that
-    cannot span two separate model instances). Tying requires encoder and decoder to share an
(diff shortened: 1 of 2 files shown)


omkar-334:
#### `transformers 4.57.6` - Before and after  
Screenshot 2026-05-24 at 5 21 46 AM  

#### `transformers 5.9.0` - Before  
Screenshot 2026-05-24 at 5 22 35 AM  

#### `transformers 5.9.0` - After  
Screenshot 2026-05-24 at 5 22 58 AM

tomaarsen:
Hello!

This looks great, thank you for the detailed before and after. I wrote a tiny cleanup and added some simple tests to avoid further regressions (hopefully), and then I think this is ready!

- Tom Aarsen

PR #3782, opened 2026-05-24
Fix example datasets that crash on datasets 4.x

omkar-334:
Several examples crash at startup on the current `datasets` 4.x / `huggingface_hub` 1.x stack, for
two related reasons:

- **script-based datasets were removed** (quora):
  ```text
  RuntimeError: Dataset scripts are no longer supported, but found quora.py
  ```
- **bare single-segment dataset ids are no longer allowed** (yahoo_answers_topics, nq_open):
  ```text
  HfUriError: Repository id must be 'namespace/name', got 'yahoo_answers_topics'
  ```

This swaps each broken id for a maintained Hugging Face dataset. The data is preserved (verified),
only the loading id (and, for quora, the access pattern) changes.

## Data verification

- quora: corpus is byte-identical, proven over the full list. The old code flattens the 404290 quora
  pairs into 808580 questions (with duplicates, in pair order). Flattening `quora-duplicates`
  `pair-class` the same way yields a list that is exactly equal (`old == new`, all 808580 elements,
  same order) to the raw quora tsv flattened. See `scripts/verify_quora_corpus_identical.py`.
- yahoo, nq_open: the bare ids redirect to the namespaced repos and resolve to the same commit
  sha, so the data is byte-identical (only the namespace changes).

### DIFF
--- a/examples/sentence_transformer/applications/computing-embeddings/computing_embeddings_streaming.py
+++ b/examples/sentence_transformer/applications/computing-embeddings/computing_embeddings_streaming.py
@@ -28,7 +28,7 @@ if __name__ == "__main__":
     encode_batch_size = 128  # Batch size of the model
 
     # Load a large dataset in streaming mode. more info: https://huggingface.co/docs/datasets/stream
-    dataset = load_dataset("yahoo_answers_topics", split="train", streaming=True)
+    dataset = load_dataset("community-datasets/yahoo_answers_topics", split="train", streaming=True)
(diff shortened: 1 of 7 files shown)


omkar-334:
#### `quora` dataset
old  
image  


new  
image

tomaarsen:
Hello!

Thanks a bunch for finding these. I ran them all locally, and they seem to work. The updated datasets also seem to match the old ones. Merging now!

- Tom Aarsen

PR #3783, opened 2026-05-24
Migrate MS MARCO examples to hf datasets

omkar-334:
MS MARCO examples downloaded raw files from `msmarco.z22.web.core.windows.net` (a 1GB tarball
plus qrels/queries) and parsed them by hand. This replaces those raw downloads with the maintained
Hf datasets

## Files

- `training/ms_marco/eval_msmarco.py`
- `training/ms_marco/multilingual/translate_queries.py`
- `training/ms_marco/eval_cross_encoder_trec_dl.py`

## Data verification

- corpus: byte-identical (500/500 sampled passages match the azure `collection.tsv`; row count
  matches exactly at 8,841,823).
- query coverage: dev 6980/6980 and train 502939/502939 query ids are 100% present in the HF
  queries; qrels are equivalent

### DIFF
--- a/examples/cross_encoder/training/ms_marco/eval_cross_encoder_trec_dl.py
+++ b/examples/cross_encoder/training/ms_marco/eval_cross_encoder_trec_dl.py
@@ -22,6 +22,7 @@ from collections import defaultdict
 import numpy as np
 import pytrec_eval
 import tqdm
+from datasets import load_dataset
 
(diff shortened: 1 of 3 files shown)


tomaarsen:
Hello!

Thank you, I did a slight cleanup, but this was already looking strong.

- Tom Aarsen

PR #3785, opened 2026-05-24
migrate quora semantic-search / clustering examples to hf

omkar-334:
These examples downloaded the raw `quora_duplicate_questions.tsv` from `qim.fs.quoracdn.net` andbuilt a deduplicated question corpus from it (capped at `max_corpus_size`). They are not broken (the raw download still works), so this is modernization, not a crash fix. The same quora questions live in `sentence-transformers/quora-duplicates`, so the download + `csv` parse is replaced with `load_dataset(...)`.

The unused `csv` / `http_get` imports are dropped and `from datasets import load_dataset` added.

## Files (5 changed)

- `applications/semantic-search/semantic_search_quora_annoy.py`
- `applications/semantic-search/semantic_search_quora_faiss.py`
- `applications/semantic-search/semantic_search_quora_hnswlib.py`
- `applications/semantic-search/semantic_search_quora_pytorch.py`
- `applications/clustering/fast_clustering.py`

## Excluded

- `applications/semantic-search/semantic_search_quora_elasticsearch.py`: it keys the corpus by `qid` (`all_questions[row["qid1"]] = ...`), and `quora-duplicates` `pair-class` has no qids. The only qid-bearing config (`quora-duplicates-mining` `questions`) is a deduplicated set in a different order, so it would not be byte-identical. Left on the raw download to preserve behavior.
- `training/quora_duplicate_questions/create_splits.py`: a ~450-line utility that *generates* the quora-IR splits from the raw tsv (builds a transitive-closure duplicate graph). It produces the very dataset that the HF mirrors were built from, so there is nothing to load from HF. Left as-is.

#### Old 
image  

#### New
image

### DIFF
--- a/examples/sentence_transformer/applications/clustering/fast_clustering.py
+++ b/examples/sentence_transformer/applications/clustering/fast_clustering.py
@@ -12,38 +12,29 @@ The method for finding the communities is extremely fast, for clustering 50k sen
 In this example, we download a large set of questions from Quora and then find similar questions in this set.
 """
 
-import csv
-import os
 import time
 
+from datasets import load_dataset
+
 from sentence_transformers import SentenceTransformer
-from sentence_transformers.util import community_detection, http_get
+from sentence_transformers.util import community_detection
 
 # Model for computing sentence embeddings. We use one trained for similar questions detection
 model = SentenceTransformer("sentence-transformers/all-MiniLM-L6-v2")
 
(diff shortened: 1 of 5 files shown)


tomaarsen:
Hello!

Thanks, this is great. I simplified some of the scripts like fast_clustering.py 🤗 

- Tom Aarsen

PR #3788, opened 2026-05-30
Fix formatting in contrastive_tension.py documentation

jadermcs:
Updated documentation formatting in contrastive_tension.py

The previous format wasn't rendering as a table (Inputs section).

image


tomaarsen:
Hello!

Well spotted, and thanks for the fix! I've confirmed that it indeed works with your fix.

- Tom Aarsen

PR #3791, opened 2026-06-02
[`examples`] Benchmark plain and graph compilation

kddubey:
Hello!

This PR:
1. Adds an example of bucket-based CUDA graph compilation and benchmarks it against no compilation and "plain" compilation: `model[0].compile(dynamic=True)`. Graph compilation achieves a modest speedup by eliminating Python overhead b/t torch and CUDA. Sharing this code as a dependency-free speedup for latency-sensitive services running standard sentence transformer models on CUDA for shorter sequences.
2. Adds a tip to use `model[0].compile(dynamic=True)` w/ a link to the benchmark. The current [documented method, `model.compile()`](https://sbert.net/docs/package_reference/sentence_transformer/model.html#sentence_transformers.sentence_transformer.model.SentenceTransformer.compile), is a no-op for inference.

The benchmark results are in the [README](https://github.com/huggingface/sentence-transformers/pull/3791/files#diff-7f9f55c565315f06010db64dc7dd7cfcee6fb3526b5f951ec9978e607bec3789).

I'm contributing graph compilation as an example rather than library code. Reasons to be cautious:
- `gte-modernbert-base` had numerical drift when graph-compiled
- The code assumes a pretty standard tokenizer. `compiled.SentenceTransformer` overrides the tokenizer, whose implementation can change a bit, e.g., v5.3 changed `tokenize` -> `preprocess`
- Compilation can add a lot of startup time.

<details>
<summary>Another thing I'm less sure about</summary>
Deploying the graph-compiled model to prod caused an error where concurrent GPU usage triggered an opaque device-side assert. Our deployment has many replicas and it has health checks to self-heal, so this error wasn't a big deal. But it's not obvious how to root cause an error from a graph-compiled call, even w/ CUDA_LAUNCH_BLOCKING=1. I needed to put a lock on GPU model calls to fix it, and now compilation has been running in prod for a few weeks w/o error and 3x lower latency.
</details>

Disclaimer: a handful of trivial lines in compiled.py and benchmark.py were written by AI, and some of the tests were implemented by AI.


sentence_transformers/sentence_transformer/model.py
```diff
@@ -214,6 +214,27 @@ def __init__(
...
+    def compile(self, *args, **kwargs):
+        """
+        Compile this Module's forward using :func:`torch.compile`.
+
+        This Module's ``__call__`` method is compiled and all arguments are passed as-is
+        to :func:`torch.compile`.
+
+        See :func:`torch.compile` for details on the arguments for this function.
+
+        .. tip::
```
kddubey, line 230: I noticed these `.. tip::` sections don't show up in my IDE b/c it tries to render markdown on hover. Can change to something like—

```rst
Note
----
```

—to make it friendly to RST and markdown

sentence_transformers/sentence_transformer/model.py
```diff
@@ -214,6 +214,27 @@ def __init__(
                     "Either update the model configuration or call `model.set_pooling_include_prompt(False)` after loading the model."
                 )
 
+    def compile(self, *args, **kwargs):
```
kddubey, line 221: This change keeps the existing [`compile` doc](https://sbert.net/docs/package_reference/sentence_transformer/model.html#sentence_transformers.sentence_transformer.model.SentenceTransformer.compile) and behavior, but adds a tip about how to compile a `SentenceTransformer`. By itself, the method is a no-op for inference. In a future change we can consider adding a method which auto-finds the `Transformer` module and compiles it

tomaarsen:
Hello!

Apologies for the delayed response. I wasn't 100% sure what we should do with this PR yet.

Thank you for this, the benchmarks and write-up made it easy to reason about. The most valuable finding here is the one in your second point: `model.compile()` is effectively a no-op for inference, because `encode` calls `self.forward` directly and never goes through `__call__`, which is where `torch.compile` installs its compiled path. That is a real bug worth fixing on its own, and it applies equally to `CrossEncoder.predict` and `SparseEncoder.encode`.

I've opened #3848 as a lighter-weight counterpart to this PR. Rather than taking the bucketed `compiled.SentenceTransformer` into the examples, it does two things:

* Fixes the root cause: `encode()` / `predict()` now run the forward pass through `__call__` in `SentenceTransformer`, `CrossEncoder`, and `SparseEncoder`, so `model.compile(...)` actually applies to inference. It also adds a documented `BaseModel.compile()` override with the `dynamic=True` tip, and a regression test per model class that fails on the old direct-`forward` path.
* Documents `torch.compile` on the "Speeding up Inference" pages for all three model types: `dynamic=True` for variable-length inputs, `mode="reduce-overhead"` (CUDA graphs) as the real speedup lever, and padding to a fixed length via `processing_kwargs` when you want the stable shapes that CUDA graphs need. That last part is the one-liner version of your bucketing.

## Benchmarks

I benchmarked on an RTX 3090 (bf16, batch size 1, varying input lengths) before writing the docs, across a range of model sizes:

| model | params | `compile(dynamic=True)` | `compile(mode="reduce-overhead")` |
|-------|--------|-------------------------|-----------------------------------|
| `all-MiniLM-L6-v2` | 23M | ~0.6x | ~1.0x |
| `bge-small-en-v1.5` | 33M | ~1.3x | ~3.0x |
| `modernbert-embed-large` | 395M | ~1.3x | ~3.7x |

The takeaways lined up with yours, with a couple of extra data points:

* `reduce-overhead` (CUDA graphs) is where the real win is, roughly 3-3.7x on the mid and large models. Plain `compile(dynamic=...)` is modest, and on a tiny model on a fast GPU it does nothing (or slightly regresses), since inference there is dominated by tokenization and Python overhead rather than the forward pass.
* The gain grows with model size, so the smallest model benefits the least.
* Padding every input up to a large `max_seq_length` is a trap. On `modernbert-embed-large` (8192) a fixed graph at full length runs at ~0.11x, i.e. about 9x slower than not compiling at all. This is exactly why padding close to your typical length, or bucketing, is the thing that matters.

## On the bucketed example

I'd rather not take `compiled.SentenceTransformer` into the examples, and I want to be clear that this is not a knock on the code. It subclasses `SentenceTransformer` and overrides `preprocess` / `tokenize` / `forward` behind a version gate, and it is excluded from CI, so it would quietly rot the next time those internals move (you flagged the `tokenize` -> `preprocess` rename yourself). It reads more like a very good optimization write-up than a maintainable example, and I think it would live better as a blog post or gist.

Then, after #3848 is merged, users can at least start using `model.compile(...)` instead of `model[0].compile(...)`. What do you think?

- Tom Aarsen

PR #3800, opened 2026-06-06
[fix] Cast learning-to-rank loss logits to float32 for bf16/fp16 CrossEncoder training

### LINKED ISSUE #3793: Reranking listwise loss occur float and bfloat16 conflict
I use PListMLELoss/LambdaLoss/... to train a reranker model based on bge-reranker-v2-m3, then at plist_mle.py line 222
`
logits_matrix[batch_indices, doc_indices] = logits
`
Runtime Error: Index put requires the source and destination dtype match, got Float for the destination and BFloat16 for the source.

This maybe a bug, because I find in the following code
`
logits_matrix = torch.full((batch_size, max_docs), 1e-16, device=self.model.device)
`
the type of 1e-16 is float32.

Incheonkirin:
Fixes #3793.

Training a CrossEncoder reranker in bf16/fp16 crashes in the learning-to-rank losses. The logits matrix is built with `torch.full(...)`, which is float32 by default, and then the model's logits get scattered into it — so as soon as the model's in bf16 you get:

```
RuntimeError: Index put requires the source and destination dtypes match, got Float for the destination and BFloat16 for the source.
```

I just cast the logits to float32 (`.float()` after the activation) rather than the `dtype=logits.dtype` from the issue. Matching the matrix to the logits dtype doesn't fully fix it — the next line that writes the labels (`labels_matrix[...] = torch.cat(labels).float()`) crashes the same way in reverse. On top of that these losses do `exp`/`cumsum`/`log`/`softmax` with no max-shift, so bf16 is shaky and fp16 overflows. Running the loss in fp32 avoids both, and it's a no-op when you're already in float32.

Same pattern was in `PListMLELoss`, `LambdaLoss`, `ListNetLoss` and `ADRMSELoss`. `ListMLELoss` and `RankNetLoss` subclass those two, so they're covered too.

Added a test that runs all six on a bf16 model — fails on main, passes here.

### DIFF
--- a/sentence_transformers/cross_encoder/losses/adr_mse.py
+++ b/sentence_transformers/cross_encoder/losses/adr_mse.py
@@ -178,7 +178,7 @@ class ADRMSELoss(nn.Module):
             logits_list.append(logits)
 
         logits = torch.cat(logits_list, dim=0)
-        logits = self.activation_fn(logits)
+        logits = self.activation_fn(logits).float()
 
         # Place logits into a padded matrix
         logits_matrix = torch.full((batch_size, max_docs), -1e16, device=self.model.device)
--- a/sentence_transformers/cross_encoder/losses/adr_mse.py
+++ b/sentence_transformers/cross_encoder/losses/adr_mse.py
@@ -177,8 +177,8 @@ class ADRMSELoss(nn.Module):
             logits = self.model(tokens)["scores"].view(-1)
             logits_list.append(logits)
 
-        logits = torch.cat(logits_list, dim=0)
-        logits = self.activation_fn(logits).float()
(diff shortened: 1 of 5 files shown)


tomaarsen:
Hello!

I was able to reproduce the issue, and the issue indeed seems fixed with this. I made one small change: move the .float() call to before the activation (as the activation can perhaps also be fragile with bf16). It looks good to me beyond that. Thank you!

- Tom Aarsen

Incheonkirin:
Thanks so much, Tom!! Great call moving the upcast before the activation
I hadn't thought about the activation itself being fragile in bf16 too. And excluding fp16 on CPU makes total sense.
Really appreciate you reproducing it and tightening everything up!

PR #3802, opened 2026-06-08
docs : added a hardware constraint warning for low vram GPUs for preventing memory thrashing

sreyanshacharya:
hey team! 

while working with sentence-transformers in a local rag setup alongside quantized 4-bit llms on consumer hardware (specifically an rtx 4050 with 6gb vram), i noticed severe compute thrashing and inter-token latency degradation when the embedding model and llm shared cuda execution.

isolating the embedding model to the cpu completely stabilized generation latency with negligible impact on retrieval speed. 

i noticed the efficiency page doesn't explicitly touch on low-vram/consumer gpu orchestration limits when sharing compute with a generator, so i added a small warning block under the pytorch section to save other local developers from hitting this specific kv-cache expansion trap. 

let me know if any formatting or wording needs tweaking!


tomaarsen:
Hello!

Thanks for the writeup, and for digging into the latency behavior on your setup.

I'm afraid I see a few issues with the block as written: the `.. warning::` content isn't indented under the directive so it won't actually render inside the admonition, and recommending CPU as a general default sits a little awkwardly on this page, which is all about making the embedding model *faster* (CPU embedding is usually much slower than GPU). The "empirical benchmarks show..." framing is also a bit strong for a single-setup observation.

That said, the core takeaway is useful and I'm happy to include it. Could we trim it down to a brief 1-2 line note that just recommends keeping an eye on VRAM and latency when running a Sentence Transformers model and a generative model on the same hardware? Something like:

```rst
.. note::

   When running a Sentence Transformers model alongside a generative LLM on the same GPU, keep an eye on VRAM usage and generation latency, as the two can contend for memory and compute. For latency-sensitive local setups, moving small embedding models to the CPU (``device="cpu"``) can help.
```

And also probably move it to below the PyTorch GPU tabs (i.e. just above the ONNX header). If you're on board, I can make that change in this PR and merge it, or if you'd like to make the change yourself, feel free to push an update to this branch and I'll review it.

- Tom Aarsen

sreyanshacharya:
hey Tom,

thanks for the great feedback! i completely agree with the restructuring. i've gone ahead and removed the warning block, converted it to the note format you suggested, and relocated it right above the ONNX header. 

just pushed the updates to the branch—let me know if it looks good to go!

docs/sentence_transformer/usage/efficiency.rst
```diff
@@ -130,6 +131,13 @@ If you're using a GPU, then you can use the following options to speed up your i
       for a full overview of available ``attn_implementation`` options, including ``"flash_attention_2"``,
       ``"flash_attention_3"``, ``"sdpa"``, and more.
 
+.. note::
+
+   When running a Sentence Transformers model alongside a generative LLM on the same GPU, keep an eye on VRAM usage and generation latency, as the two can contend for memory and compute. For latency-sensitive local setups, moving small embedding models to the CPU (``device="cpu"``) can help.
+
+ONNX
+----
+
```
tomaarsen, line 139:
```suggestion
```
This header is accidentally duplicated now

docs/sentence_transformer/usage/efficiency.rst
```diff
@@ -48,6 +48,7 @@ The PyTorch backend is the default backend for Sentence Transformers. If you don
    sentences = ["This is an example sentence", "Each sentence is converted"]
    embeddings = model.encode(sentences)
 
+
```
tomaarsen:
```suggestion
```
This just keeps the diff a bit cleaner 🤗

tomaarsen (approved):
Some minor comments, but it looks good!

PR #3813, opened 2026-06-11
[`ci`] Avoid model2vec distill install in CI as it limits the transformers version

tomaarsen:
Hello!

## Pull Request overview
* Stop installing `model2vec[distill]` on the `transformers>=5.0.0` CI leg so it tests the latest transformers instead of being capped to `<5.4.0`
* Keep `model2vec[distill]` on the `transformers<5.0.0` leg, where the distillation tests actually run

## Details
The `model2vec[distill]` extra pins `transformers<5.4.0`, and the model2vec install step ran on every matrix leg, so the `transformers>=5.0.0` job was silently resolving transformers 5.3.0 rather than the latest release. I caught this from the huggingface_hub release-candidate CI: the downgraded transformers 5.3.0 got paired with the 5.11-era `tokenizers==0.23.0rc0` that the first install had pulled in, and that mismatch broke RoBERTa/CLIP tokenizer construction with `RobertaProcessing.__new__() got an unexpected keyword argument 'cls'`.

Base `model2vec` has no `transformers` (or `torch`) dependency at all; only the `distill` extra constrains it. The distillation tests (`test_from_distillation`) already skip on transformers v5+, so the `>=5.0.0` leg never needed the extra in the first place. I now install `model2vec[distill]` only on the `<5.0.0` leg (where the `<5.4.0` cap is moot and distillation actually runs) and plain `model2vec` elsewhere, so the latest-transformers leg resolves the latest transformers and stays internally consistent. I also dropped the `contains(fromJSON(...))` guard on the install step, which was always true now that the matrix is `3.10`–`3.13`.

- Tom Aarsen

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -80,9 +80,15 @@ jobs:
         if: ${{ matrix.transformers-version == '>=5.0.0' }}
         run: uv pip install '.[train]' --group test 'transformers>=5.0.0' 'accelerate>=1.13.0' --excludes excludes.txt --system
 
-      - name: Install model2vec
+      # model2vec[distill] pins transformers<5.4.0, so use the distill extra only on
+      # the <5.0.0 leg (distillation tests skip on v5+ anyway) and base model2vec elsewhere.
+      - name: Install model2vec (with distillation extra)
         run: uv pip install model2vec[distill] --system
-        if: ${{ contains(fromJSON('["3.10", "3.11", "3.12", "3.13"]'), matrix.python-version) }}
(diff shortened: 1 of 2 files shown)


tomaarsen:
Ideally, this'll solve the Sentence Transformers <> `huggingface_hub` automatic release candidate tests that trigger whenever the latter has a new release coming up. E.g. from this branch: https://github.com/huggingface/sentence-transformers/tree/refs/heads/ci-test-huggingface-hub-1.19.0.rc1-release.

- Tom Aarsen

PR #3816, opened 2026-06-12
[fix] Avoid materializing the full similarity matrix in mine_hard_negatives without FAISS

### LINKED ISSUE #3164: RAM explosion when using mine hard negatives
Im trying to run this example snippet, but after the batches get processed, the RAM explodes exponentially and crashes.

I tested using my local setup(32 GB RAM and rtx 4060) and colab(16 GB and T4) with cuda enabled

```
# Load a Sentence Transformer model
# Load a dataset to mine hard negatives from
dataset = load_dataset("sentence-transformers/natural-questions", split="train")

dataset = mine_hard_negatives(
     dataset=dataset,
     model=model,
     range_min=10,
     range_max=50,
     max_score=0.8,
     margin=0.1,
     num_negatives=5,
     sampling_strategy="random",
     batch_size=128,
     use_faiss=False,
)
```

Incheonkirin:
## Summary

Fixes #3164.

With `use_faiss=False`, `mine_hard_negatives` currently computes the full
`(n_queries, n_corpus)` similarity matrix before running `topk`:

```python
scores = model.similarity(query_embeddings, corpus_embeddings).to(device)
scores, indices = torch.topk(scores, k=range_max + max_positives, dim=1)
```

For the scale mentioned in #3164 (Natural Questions, roughly 100k queries x 100k
corpus entries), that float32 matrix is about 40GB before overhead, so it can
easily exhaust 16-32GB machines. The FAISS branch already avoids this by searching
in `faiss_batch_size` chunks over the query axis.

This PR applies the same query-axis batching to the non-FAISS branch: compute
similarity + `topk` per chunk, then concatenate the per-chunk results. Since each
row is ranked independently, this preserves the mined negatives while bounding
the intermediate similarity matrix to `faiss_batch_size * len(corpus)`.

I also generalized the `faiss_batch_size` docstring, added a clear `ValueError`
for `faiss_batch_size <= 0`, and made both search progress bars respect `verbose`.

Local peak RSS delta with `all-MiniLM-L6-v2` on CPU and a synthetic corpus:

| queries x corpus | before | after |
|---|---:|---:|
| 5k x 50k | 1.28 GB | 0.55 GB |
| 10k x 100k | 4.56 GB | 1.09 GB |

In local equivalence checks, the mined outputs matched exactly before/after for
n-tuple and triplet formats, including non-divisible chunk sizes.

## Tests

Added a regression test that checks both the query-axis chunking behavior
(`[3, 3, 2]` for `faiss_batch_size=3`) and output equivalence with a single-batch
run. The same test fails on the previous implementation, which calls
`model.similarity` once for all 8 queries.

```bash
pytest tests/util/test_hard_negatives.py -q
ruff check sentence_transformers/util/hard_negatives.py tests/util/test_hard_negatives.py
```

### DIFF
--- a/sentence_transformers/util/hard_negatives.py
+++ b/sentence_transformers/util/hard_negatives.py
@@ -467,7 +467,7 @@ def mine_hard_negatives(
         scores_list = []
         indices_list = []
         # Iterate over query embeddings in batches so we can track the progress
-        for i in trange(0, len(query_embeddings), faiss_batch_size, desc="Querying FAISS index", disable=not verbose):
+        for i in trange(0, len(query_embeddings), faiss_batch_size, desc="Querying FAISS index"):
             query_chunk = query_embeddings[i : i + faiss_batch_size]
             scores, indices = index.search(query_chunk, k=range_max + 1)
             scores_list.append(scores)
@@ -482,7 +482,7 @@ def mine_hard_negatives(
         # Compute the similarity scores between the queries and the corpus in batches, to avoid
         # materializing the full (n_queries, n_corpus) similarity matrix
         for i in trange(
-            0, len(query_embeddings), faiss_batch_size, desc="Computing similarity scores", disable=not verbose
+            0, len(query_embeddings), faiss_batch_size, desc="Computing similarity scores"
         ):
             chunk_scores = model.similarity(query_embeddings[i : i + faiss_batch_size], corpus_embeddings).to(device)
 
--- a/sentence_transformers/util/hard_negatives.py
+++ b/sentence_transformers/util/hard_negatives.py
@@ -478,6 +478,7 @@ def mine_hard_negatives(
     else:
         scores_list = []
         indices_list = []
+        corpus_embeddings = torch.as_tensor(corpus_embeddings)
         # Compute the similarity scores between the queries and the corpus in batches, to avoid
         # materializing the full (n_queries, n_corpus) similarity matrix
         for i in trange(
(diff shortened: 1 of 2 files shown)


sentence_transformers/util/hard_negatives.py
```diff
@@ -462,7 +467,7 @@ def mine_hard_negatives(
         scores_list = []
         indices_list = []
         # Iterate over query embeddings in batches so we can track the progress
-        for i in trange(0, len(query_embeddings), faiss_batch_size, desc="Querying FAISS index"):
+        for i in trange(0, len(query_embeddings), faiss_batch_size, desc="Querying FAISS index", disable=not verbose):
```
tomaarsen:
```suggestion
        for i in trange(0, len(query_embeddings), faiss_batch_size, desc="Querying FAISS index"):
```
With disabling verbose, I just want to disable the logs/stats, not necessarily the progress bars. I think I'd prefer to keep this as it was.

sentence_transformers/util/hard_negatives.py
```diff
@@ -471,11 +476,21 @@ def mine_hard_negatives(
...
     else:
-        # Compute the similarity scores between the queries and the corpus
-        scores = model.similarity(query_embeddings, corpus_embeddings).to(device)
+        scores_list = []
+        indices_list = []
+        corpus_embeddings = torch.as_tensor(corpus_embeddings)
+        # Compute the similarity scores between the queries and the corpus in batches, to avoid
+        # materializing the full (n_queries, n_corpus) similarity matrix
+        for i in trange(
+            0, len(query_embeddings), faiss_batch_size, desc="Computing similarity scores", disable=not verbose
```
tomaarsen:
```suggestion
            0, len(query_embeddings), faiss_batch_size, desc="Computing similarity scores"
```
Idem.

tomaarsen (commented):
Thanks for this. I've made some modifications re. `verbose`, which I'll merge.

PR #3818, opened 2026-06-12
fix `MPS` errors

omkar-334:
Three spots in the library still hardcode CUDA, even though MPS is listed as a supported device in `get_device_name()`. This PR fixes them.

## What changes

1. `sparse_encoder/model.py`: `SparseEncoder.sparsity()` calls `to_sparse_csr()`,
which has no kernel on `SparseMPS`. Wrapped in `try/except NotImplementedError` with a CPU fallback. The result is a Python float so it's bit-identical, and the fallback covers any other backend missing the kernel too.

2. `sentence_transformer/fit_mixin.py` and `cross_encoder/fit_mixin.py`: `fit(use_amp=True)` was hardcoded to `torch.cuda.amp.GradScaler()` (and to `from torch.cuda.amp import autocast` in the SentenceTransformer side). Now both files dispatch on `device.type`: cuda and npu keep their existing namespaces, everything else (mps, xpu, cpu) goes through `torch.amp.*`.

## Testing

Ran the small examples on Apple Silicon end to end against this branch, without `PYTORCH_ENABLE_MPS_FALLBACK`. All passed:

* SparseEncoder smoke (encode, sparsity, similarity, intersection on `naver/splade-cocondenser-ensembledistil`)
* `computing_embeddings.py`
* `clustering/agglomerative.py`, `clustering/kmeans.py`
* `semantic-search/semantic_search.py`
* `cross_encoder/applications/cross_encoder_usage.py`

I also tried the heavier retrieval examples (Quora 100k with pytorch top-k, HNSWlib, Annoy). Those work too.

### DIFF
--- a/sentence_transformers/sparse_encoder/model.py
+++ b/sentence_transformers/sparse_encoder/model.py
@@ -1035,8 +1035,13 @@ class SparseEncoder(BaseModel):
                 "sparsity_ratio": 1.0,
             }
 
-        # CSR gives O(1) per-row non-zero counts via crow_indices
-        embeddings = embeddings.to_sparse_csr()
+        # CSR gives O(1) per-row non-zero counts via crow_indices.
+        # SparseMPS (and some other accelerator backends) lack the to_sparse_csr kernel,
+        # so fall back to CPU when the conversion is not implemented.
+        try:
+            embeddings = embeddings.to_sparse_csr()
+        except NotImplementedError:
+            embeddings = embeddings.cpu().to_sparse_csr()
         crow_indices = embeddings.crow_indices()
         non_zero_per_row = crow_indices[1:] - crow_indices[:-1]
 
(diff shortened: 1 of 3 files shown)


sentence_transformers/sparse_encoder/model.py
```diff
@@ -1035,8 +1035,13 @@ def sparsity(embeddings: torch.Tensor) -> dict[str, float]:
                 "sparsity_ratio": 1.0,
             }
 
-        # CSR gives O(1) per-row non-zero counts via crow_indices
-        embeddings = embeddings.to_sparse_csr()
+        # CSR gives O(1) per-row non-zero counts via crow_indices.
+        # SparseMPS (and some other accelerator backends) lack the to_sparse_csr kernel,
+        # so fall back to CPU when the conversion is not implemented.
```
tomaarsen:
```suggestion
        # CSR gives O(1) per-row non-zero counts via crow_indices, and fall back to CPU for MPS.
```
Let's shrink this comment a bit

tomaarsen (commented):
Overall this looks good! I made a small comment to shrink the new comment a bit. Thanks for verifying that this works on your hardware 🤗

PR #3820, opened 2026-06-13
Fix relative_margin sign-dependence in mine_hard_negatives

### LINKED ISSUE #3819: relative_margin can keep negatives more similar than positives when positive-pair scores are negative
`mine_hard_negatives(relative_margin=...)` keeps negatives that are more similar to the anchor than the positive pair when the positive-pair score is negative.

The threshold is `positive_score * (1 - relative_margin)`. For a positive score of 0.8, `relative_margin=0.05` lowers it to 0.76 and correctly drops closer negatives. For a negative positive-pair score it moves the other way: -0.50 becomes -0.475, so a negative at -0.49 survives even though it is more similar than the positive (-0.49 > -0.50), because -0.49 > -0.475 is false. Enabling the margin then filters less than `relative_margin=0.0`.

Repro:

```python
import math
from types import SimpleNamespace

import torch
from datasets import Dataset


class ControlledModel:
    device = torch.device("cpu")
    model_card_data = SimpleNamespace(base_model="controlled-relative-margin")

    def __init__(self):
        self.vecs = {
            "q": torch.tensor([1.0, 0.0]),
            "p": torch.tensor([-0.50, math.sqrt(1 - 0.50**2)]),
            "n_more_similar": torch.tensor([-0.49, math.sqrt(1 - 0.49**2)]),
            "n_far": torch.tensor([-0.80, math.sqrt(1 - 0.80**2)]),
        }

    def encode_query(self, texts, **kwargs):
        return torch.stack([self.vecs[t] for t in texts]).numpy()

    def encode_document(self, texts, **kwargs):
        return torch.stack([self.vecs[t] for t in texts]).numpy()

    def similarity(self, queries, corpus):
        return torch.as_tensor(queries, dtype=torch.float32) @ torch.as_tensor(corpus, dtype=torch.float32).T

    def similarity_pairwise(self, queries, positives):
        return (torch.as_tensor(queries, dtype=torch.float32) * torch.as_tensor(positives, dtype=torch.float32)).sum(dim=1)


dataset = Dataset.from_dict({"query": ["q"], "positive": ["p"]})
corpus = ["p", "n_more_similar", "n_far"]

for margin in [0.0, 0.05]:
    mined = mine_hard_negatives(
        dataset=dataset,
        model=ControlledModel(),
        anchor_column_name="query",
        positive_column_name="positive",
        corpus=corpus,
        range_max=2,
        num_negatives=1,
        relative_margin=margin,
        output_scores=True,
        verbose=False,
    )
    print(margin, mined.to_list())
```

On `main`:

```text
0.0  [{'query': 'q', 'positive': 'p', 'negative': 'n_far', 'scores': [-0.5, -0.800000011920929]}]
0.05 [{'query': 'q', 'positive': 'p', 'negative': 'n_more_similar', 'scores': [-0.5, -0.49000000953674316]}]
```

A score-order threshold removes the sign dependence:

`relative_threshold = positive_score - abs(positive_score) * relative_margin`

This matches the current behavior for positive scores and stays below the positive for negative ones. I can send a regression test and patch.

vineethsaivs:
## Summary

Fixes #3819.

`mine_hard_negatives(relative_margin=...)` used `positive_score * (1 - relative_margin)` as the cutoff for dropping negatives that are too similar to the anchor. That expression is **sign-dependent**:

- For a positive pair score of `0.8` and `relative_margin=0.05`, the threshold becomes `0.76` and correctly drops closer negatives.
- For a **negative** pair score of `-0.50`, it becomes `-0.475`, which is *higher* than the positive. A negative at `-0.49` (more similar to the anchor than the positive) then survives, so enabling the margin filtered **less** than `relative_margin=0.0`.

## Fix

Use a score-order threshold:

```python
relative_thresholds = max_positive_scores - max_positive_scores.abs() * relative_margin
```

This equals `positive * (1 - relative_margin)` for positive scores (existing behavior unchanged) and always stays below the positive for negative scores, removing the sign dependence.

## Test

Added `test_relative_margin_negative_positive_score`, a deterministic regression test using a tiny controlled model (no model download, CPU only). It reproduces the issue's scenario (positive score `-0.50`, a closer negative at `-0.49`, a farther one at `-0.80`) and asserts that any non-zero `relative_margin` drops the too-similar negative. The test **fails on `main`** (mines `n_more_similar`) and **passes** with this change (mines `n_far`).

Verified on macOS / CPU.

---
*Disclosure: this change was prepared with AI assistance; I reviewed it and can explain every line.*

### DIFF
--- a/sentence_transformers/util/hard_negatives.py
+++ b/sentence_transformers/util/hard_negatives.py
@@ -586,7 +586,13 @@ def mine_hard_negatives(
                 num_candidates -= num_skipped
 
         if relative_margin is not None:
-            removed_indices = scores > max_positive_scores.repeat(scores.size(1), 1).T * (1 - relative_margin)
+            # Use a score-order threshold (positive - |positive| * relative_margin) rather than
+            # positive * (1 - relative_margin). The latter is sign-dependent: for a negative positive-pair
+            # score it raises the threshold (e.g. -0.50 -> -0.475), so negatives that are *more* similar than
+            # the positive can survive. Subtracting the absolute value always keeps the threshold below the
+            # positive while matching the previous behavior for positive scores.
+            relative_thresholds = max_positive_scores - max_positive_scores.abs() * relative_margin
+            removed_indices = scores > relative_thresholds.repeat(scores.size(1), 1).T
             scores[removed_indices] = -float("inf")
 
             num_skipped = removed_indices.sum().item()
--- a/tests/util/test_hard_negatives.py
+++ b/tests/util/test_hard_negatives.py
@@ -1,11 +1,14 @@
 from __future__ import annotations
 
 import importlib.util
+import math
 import os
 from copy import deepcopy
 from pathlib import Path
+from types import SimpleNamespace
 
 import pytest
+import torch
 from torch import Tensor
 
 from sentence_transformers import CrossEncoder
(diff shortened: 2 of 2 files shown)


tomaarsen:
Hello!

Thank you for jumping on this so quickly, and for the clean fix with a solid regression test. Your `positive - abs(positive) * relative_margin` threshold is exactly right.

I'm going to go with #3821 here, though, as it also fixes the same sign-dependent expression in `GISTEmbedLoss` and `CachedGISTEmbedLoss`, which share the identical `* (1 - margin)` cutoff in their `margin_strategy="relative"` path. Since the two PRs overlap on `mine_hard_negatives`, it's cleaner to land the one that covers all three spots in one go.

I hope you understand. Going to close this in favour of #3821.

- Tom Aarsen

PR #3821, opened 2026-06-13
[fix] Make relative margin sign-independent in mining and GIST losses

### LINKED ISSUE #3819: relative_margin can keep negatives more similar than positives when positive-pair scores are negative

Incheonkirin:
## Summary

Fixes #3819.

The relative margin used to discard negatives too similar to the anchor was computed as `positive_score * (1 - margin)`. That works for positive scores, but it is sign-dependent: when the positive-pair score is negative, the threshold moves above the positive score. For example, with a positive-pair score of `-0.50` and `margin=0.05`, the threshold becomes `-0.475`, so a negative scoring `-0.49` survives even though it is more similar than the positive.

This PR applies a score-order threshold, `positive_score - abs(positive_score) * margin`, in the three places that shared the old expression:

- `mine_hard_negatives` (the `relative_margin` argument)
- `GISTEmbedLoss` (`margin_strategy="relative"`)
- `CachedGISTEmbedLoss` (`margin_strategy="relative"`)

It is identical to `positive_score * (1 - margin)` for positive scores (existing behavior unchanged) and keeps the threshold below the positive score when it is negative.

## Tests

Added deterministic, CPU-only regression tests for the negative-score case in all three. They fail on current `main` and pass with this change:

```
pytest tests/util/test_hard_negatives.py tests/sentence_transformer/losses/test_gist_embed.py tests/sentence_transformer/losses/test_cached_gist_embed.py -q
# 101 passed, 1 skipped
```


tomaarsen:
Hello!

Thanks a lot for this, and for reporting the issue in the first place! I appreciate that you also caught the same sign-dependent expression in the `margin_strategy="relative"` path of `GISTEmbedLoss` and `CachedGISTEmbedLoss`. 

I went through it carefully: the new threshold matches the old `positive * (1 - margin)` exactly for positive scores while staying below the positive for negative ones, and each of the three regression tests fails on `main` and passes with the change.

I've pushed a few small doc follow-ups (the training overview and skill reference still had the old wording, plus some docstring tightening from Copilot's review). Merging once the tests go green, thank you for the contribution!

- Tom Aarsen

Incheonkirin:
Thanks @tomaarsen! Really glad this was useful, and thanks for the doc follow-ups and the merge.

PR #3825, opened 2026-06-16
[fix] Pass `axis=-1` to `np.packbits` in `binary`/`ubinary` quantization

JSap0914:
## Bug

`quantize_embeddings()` with `precision="binary"` or `precision="ubinary"` raises a `ValueError` when the embedding dimension is not a multiple of 8.

**Root cause:** `np.packbits(embeddings > 0)` (without an `axis` argument) flattens the entire 2-D array before packing. For `n` embeddings of dimension `d` the packed output has `ceil(n*d/8)` bytes. When `d % 8 != 0` this byte count is not evenly divisible by `n`, so the subsequent `.reshape(n, -1)` raises:

```
ValueError: cannot reshape array of size 5 into shape (4,newaxis)
```

Concrete example (n=4, d=10):
```python
import numpy as np
from sentence_transformers.util import quantize_embeddings

emb = np.random.randn(4, 10).astype(np.float32)
quantize_embeddings(emb, "binary")   # ValueError
quantize_embeddings(emb, "ubinary")  # ValueError
```

## Fix

Pass `axis=-1` to `np.packbits` so NumPy packs each row independently, yielding shape `(n, ceil(d/8))` directly — no cross-row bit mixing, and no reshape issue:

```diff
-    return (np.packbits(embeddings > 0).reshape(embeddings.shape[0], -1) - 128).astype(np.int8)
+    return (np.packbits(embeddings > 0, axis=-1).reshape(embeddings.shape[0], -1) - 128).astype(np.int8)
```

For embedding dimensions that **are** multiples of 8 (the common case, e.g. 768, 1024) the two forms are equivalent — verified in the added tests.

## Verification

```
pytest tests/util/test_quantization.py -v
# 16 tests, all pass
```

> AI-assisted contribution.

### DIFF
--- a/sentence_transformers/util/quantization.py
+++ b/sentence_transformers/util/quantization.py
@@ -435,9 +435,9 @@ def quantize_embeddings(
             return ((embeddings - starts) / steps - 128).astype(np.int8)
 
     if precision == "binary":
-        return (np.packbits(embeddings > 0).reshape(embeddings.shape[0], -1) - 128).astype(np.int8)
+        return (np.packbits(embeddings > 0, axis=-1).reshape(embeddings.shape[0], -1) - 128).astype(np.int8)
 
     if precision == "ubinary":
-        return np.packbits(embeddings > 0).reshape(embeddings.shape[0], -1)
+        return np.packbits(embeddings > 0, axis=-1).reshape(embeddings.shape[0], -1)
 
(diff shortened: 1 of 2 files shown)


tomaarsen (approved):
This makes sense to me! It was an oversight in my original implementation, thank you. I'll merge once the tests go green.

PR #3826, opened 2026-06-17
fix(modality): improve error message for mixed modality batches

### LINKED ISSUE #3722: Mixing modalities in `encode()` doesn't work for CLIP
I'm trying to get CLIP to work with the new multimodal setup, and it seems that it misinterprets lists containing multiple modalities as messages:

```python

ds = load_dataset("nbadrinath/ikea_dataset_5.0", split="train[:1000]")
images = list(ds["image"])
texts = list(ds["desc"])
print(images[0], texts[0])
# <PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=512x512 at 0x7818178354C0>  Product: Small Outdoor Balcony Set...

inputs = images + texts
encoder = SentenceTransformer("clip-ViT-B-32")
encoder.encode(inputs)
```

```
ValueError                                Traceback (most recent call last)
Cell In[61], line 1
----> 1 encoder.encode(inputs)

File ~/.conda/envs/turftopic/lib/python3.12/site-packages/torch/utils/_contextlib.py:120, in context_decorator.<locals>.decorate_context(*args, **kwargs)
    117 @functools.wraps(func)
    118 def decorate_context(*args, **kwargs):
    119     with ctx_factory():
--> 120         return func(*args, **kwargs)

File ~/.conda/envs/turftopic/lib/python3.12/site-packages/sentence_transformers/util/decorators.py:41, in deprecated_kwargs.<locals>.decorator.<locals>.wrappe
r(*args, **kwargs)
     39         else:
     40             kwargs.pop(old_name)
---> 41 return func(*args, **kwargs)

File ~/.conda/envs/turftopic/lib/python3.12/site-packages/sentence_transformers/sentence_transformer/model.py:644, in SentenceTransformer.encode(self, inputs,
 prompt_name, prompt, batch_size, show_progress_bar, output_value, precision, convert_to_numpy, convert_to_tensor, device, normalize_embeddings, truncate_dim,
 pool, chunk_size, **kwargs)
    642 for start_index in trange(0, len(inputs_sorted), batch_size, desc="Batches", disable=not show_progress_bar):
    643     inputs_batch = inputs_sorted[start_index : start_index + batch_size]
--> 644     features = self.preprocess(inputs_batch, prompt=prompt, **kwargs)
    646     if is_hpu:
    647         features = self._pad_features_for_hpu(features)

File ~/.conda/envs/turftopic/lib/python3.12/site-packages/sentence_transformers/base/model.py:532, in BaseModel.preprocess(self, inputs, prompt, **kwargs)
    527     if isinstance(modality, tuple) and all(part in self.modalities for part in modality):
    528         message += (
    529             f"\nThis model supports {' and '.join(modality)} individually, "
    530             "but not in the same input. Please process each modality separately."
    531         )
--> 532     raise ValueError(message)
    534 # Backwards compatibility: fall back to preprocess/tokenize without prompt if the
    535 # input module doesn't accept it. Only the main path (preprocess with prompt) will
    536 # be supported in the future.
    537 try:

ValueError: Modality 'message' is not supported by this SentenceTransformer model. Supported modalities: text, image
```

anishesg:
The `raise_unsupported_modality_error` function in `sentence_transformers/base/modality.py` previously raised an unhelpful error when users attempted to encode mixed modality batches (e.g., images and text together) with a model that doesn't support the 'message' modality. The error occurred because the function couldn't re-infer individual sample modalities from a batch that had already been classified as requiring 'message' support. This fix adds a fallback error path that explicitly explains the problem: when a batch contains mixed modalities and the model doesn't support 'message', each modality must be encoded separately. The new error message clearly states that the batch contains inputs that cannot be encoded together and provides actionable guidance to encode each modality in separate calls. A test was added to verify that mixed modality batches (images and text) now produce a clear, helpful error message.

Fixes #3722

### DIFF
--- a/sentence_transformers/base/modality.py
+++ b/sentence_transformers/base/modality.py
@@ -733,6 +733,15 @@ def raise_unsupported_modality_error(
                 f"This batch mixes multiple modalities ({present_str}), which this {source} cannot "
                 f"encode in a single batch. Encode each modality separately (one call per modality)."
             )
+        # If we couldn't re-infer individual sample modalities but the batch modality is "message",
+        # it's likely because the batch contains mixed modalities. Provide a helpful error message.
+        raise ValueError(
+            f"This batch contains inputs that cannot be encoded together by this {source}. "
+            f"When a model does not support the 'message' modality, all inputs in a batch must be "
+            f"of the same type. This error typically occurs when mixing different modalities "
+            f"(e.g., images and text) in a single batch. Encode each modality separately "
+            f"(one call per modality). Supported modalities: {supported_str}."
+        )
 
     # A single combined input (e.g. a {"text": ..., "image": ...} dict) with parts the source
     # supports individually but cannot fuse without "message" support: encode each separately.
--- a/tests/base/test_modality.py
+++ b/tests/base/test_modality.py
@@ -13,6 +13,7 @@ from sentence_transformers.base.modality import (
     is_audio_url_or_path,
     is_image_url_or_path,
     is_video_url_or_path,
+    raise_unsupported_modality_error,
 )
 from sentence_transformers.base.modality_types import MODALITY_TO_PROCESSOR_ARG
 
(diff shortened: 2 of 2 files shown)


sentence_transformers/base/modality.py
```diff
@@ -733,6 +733,15 @@ def raise_unsupported_modality_error(
                 f"This batch mixes multiple modalities ({present_str}), which this {source} cannot "
                 f"encode in a single batch. Encode each modality separately (one call per modality)."
             )
+        # If we couldn't re-infer individual sample modalities but the batch modality is "message",
+        # it's likely because the batch contains mixed modalities. Provide a helpful error message.
+        raise ValueError(
```
tomaarsen, line 738: There might still be an uncovered edge case somewhere here (looking into it now), but this `raise_unsupported_modality_error` is already the improved error message for mixed modality batches. It doesn't "raise an unhelpful error when users attempted to encode mixed modality batches".

tests/base/test_modality.py
```diff
@@ -1146,3 +1147,28 @@ def test_multiple_text_messages(self):
             ]
         ]
         assert InputFormatter.is_text_only_messages(batch) is True
+
+
+class TestRaiseUnsupportedModalityError:
+    def test_mixed_modality_batch_clear_error(self):
```
tomaarsen, line 1153: This test never reaches the new code, the `assert` at the bottom is just checking the already existing errors.

tomaarsen (commented):
I think we shouldn't make these changes, the error is already pretty helpful due to a previous PR, and your tests don't cover your new lines. I did notice another issue, an unnecessary try-except that hides an error that should probably just be raised. I'll make a follow-up PR for that, but I think I'll close this one.

PR #3827, opened 2026-06-17
[fix] Exclude padding from the Plackett-Luce normalizer in ListMLE/PListMLE losses

Incheonkirin:
## Summary

`PListMLELoss` (and `ListMLELoss`, which subclasses it) scatter a list's logits into a
`(batch_size, max_docs)` matrix and pad shorter lists with a logit of `1e-16`. The Plackett-Luce
log-likelihood uses a reverse cumulative sum of `scores = logits.exp()` as the per-position
normalizer. Because `exp(1e-16) ~= 1`, every padded slot enters that normalizer with roughly unit
mass, inflating the denominator of every real document's log-probability. The padding mask is only
applied *after* the cumulative sum (`log_probs[~mask] = 0.0`), so it removes the padded positions'
own terms but not their pollution of the real positions' normalizers.

As a result, a query's loss and gradients depend on `max_docs` — i.e. on how many documents the
longest *other* query in the batch happens to have. For a single list (real labels `[2, 1]`,
logits `[1.5, 0.3]`):

| padding width `max_docs` | loss | grad |
| --- | --- | --- |
| 2 (no padding) | 0.2633 | `[-0.2315, +0.2315]` |
| 3 (1 pad) | 0.9759 | `[-0.3440, -0.2280]` |
| 4 (2 pads) | 1.4671 | … |
| 6 (4 pads) | 2.1627 | … |

The same list's loss grows with padding width and the second document's gradient flips sign, so a
variable-length-list batch (the common case for reranker training) trains on a corrupted signal.

This excludes padded entries from the normalizer by zeroing their scores before the cumulative sum,
using a mask in the same (sorted) order as the logits:

```python
scores = sorted_logits.exp().masked_fill(~sorted_mask, 0.0)
```

This matches how reference ListMLE implementations handle padding (e.g. allRank sets padded
predictions to `-inf` before the exp, which is the same as zero mass in the normalizer). For a list
with no padding the change is a no-op, so existing single-length behavior is unchanged.

## Tests

Added a deterministic padding-invariance regression test for `ListMLELoss` and `PListMLELoss`,
parametrized over `respect_input_order` (True/False): it batches a 2-document query with a
3-document query and asserts the batched loss equals the mean of the two per-query losses. It fails
on current `main` and passes with this change:

```
pytest tests/cross_encoder/losses/test_listwise_padding.py -q
# main:  4 failed
# fix:   4 passed
```


tomaarsen:
Hello!

Thanks for tracking this down and for the clean padding-invariance test! I wanted to see how much it actually moves the needle in real training, so I finetuned a few `microsoft/MiniLM-L12-H384-uncased` rerankers on MS MARCO v1.1 (1 epoch, identical seed/config), fix vs. no-fix, for both affected losses.

NanoBEIR R100 mean ndcg@10 (best checkpoint):

| Loss | no fix | with this PR |
|:--|:--:|:--:|
| **ListMLE** | ~0.39 | **0.529** |
| **PListMLE** | 0.514 | **0.525** |

(Small note: The ListMLE 'no fix' was based on older ST/transformers versions, I didn't rerun that baseline)
The size of the effect lines up neatly with how much each loss leans on the padded (bottom) positions. It's strictly ≥ no-fix everywhere, and a genuine correctness win for ListMLE. Models + full training logs:

- ListMLE + fix: https://huggingface.co/tomaarsen/reranker-msmarco-v1.1-MiniLM-L12-H384-uncased-listmle-pr3827
- PListMLE + fix: https://huggingface.co/tomaarsen/reranker-msmarco-v1.1-MiniLM-L12-H384-uncased-plistmle-pr3827
- PListMLE, no fix (same setup, fix reverted): https://huggingface.co/tomaarsen/reranker-msmarco-v1.1-MiniLM-L12-H384-uncased-plistmle-pr3827-baseline
- ListMLE, no fix (older release): https://huggingface.co/tomaarsen/reranker-msmarco-v1.1-MiniLM-L12-H384-uncased-listmle

I also pushed two tiny follow-ups to your branch: trimmed the inline comment and folded the regression test into `tests/cross_encoder/losses/test_misc.py` (trying to avoid one-off test files). Merging when everything is green.

- Tom Aarsen

Incheonkirin:
Nice, bigger ListMLE bump than I'd have guessed! Thanks for benchmarking both and publishing the runs 😄 test_misc.py works for me

PR #3832, opened 2026-06-23
[fix] Memory growth in community_detection with larger datasets

AbdelRahmanYaghi:
`community_detection` suffers from increased memory usage on larger datasets because it keeps populating the same python list, i.e., `extracted_communities` with python integers. 

The solution proposed aims to switch the current python list implementation with a numpy implementation, which saves around 5 times the memory usage. Here is the calculation behind the 5 times, along with an example:

In python lists, each element roughly costs roughly 20 bytes per index. On the other hand, numpy stores each item in 4 bytes only. And hence, the memory saved is around 20/4 = 5 times less memory usage.

Using the current implementation
```
Finding clusters:   0%|                              | 0/59 [00:00<?, ?it/s]  extracted_communities: 193.33 MB, 486 communities
Finding clusters:   3%|█▏                            | 2/59 [00:03<01:45,  1.85s/it]  extracted_communities: 591.14 MB, 1461 communities
Finding clusters:   7%|██▎                           | 4/59 [00:10<02:42,  2.96s/it]  extracted_communities: 997.35 MB, 2444 communities
Finding clusters:  10%|███▌                          | 6/59 [00:20<03:50,  4.34s/it]  extracted_communities: 1391.05 MB, 3443 communities
Finding clusters:  14%|████▋                         | 8/59 [00:35<05:03,  5.95s/it]  extracted_communities: 1774.79 MB, 4425 communities
Finding clusters:  17%|█████▊                        | 10/59 [00:52<06:00,  7.36s/it]  extracted_communities: 2153.58 MB, 5413 communities
Finding clusters:  20%|██████▉                       | 12/59 [01:13<07:07,  9.09s/it]  extracted_communities: 2526.29 MB, 6379 communities
Finding clusters:  24%|████████                      | 14/59 [01:38<08:04, 10.77s/it]  extracted_communities: 2904.04 MB, 7356 communities 
```

Using the numpy implementation
```
Finding clusters:   0%|                                           | 0/59 [00:00<?, ?it/s]extracted_communities: 43.01 MB, 486 communities
Finding clusters:   3%|█▏                                 | 2/59 [00:00<00:04, 12.02it/s]extracted_communities: 131.50 MB, 1461 communities
Finding clusters:   7%|██▎                                | 4/59 [00:00<00:04, 12.33it/s]extracted_communities: 221.87 MB, 2444 communities
Finding clusters:  10%|███▌                               | 6/59 [00:00<00:04, 12.46it/s]extracted_communities: 309.45 MB, 3443 communities
Finding clusters:  14%|████▋                              | 8/59 [00:00<00:04, 12.49it/s]extracted_communities: 394.82 MB, 4425 communities
Finding clusters:  17%|█████▊                            | 10/59 [00:00<00:03, 12.59it/s]extracted_communities: 479.09 MB, 5413 communities
Finding clusters:  20%|██████▉                           | 12/59 [00:00<00:03, 11.82it/s]extracted_communities: 562.00 MB, 6379 communities
Finding clusters:  24%|████████                          | 14/59 [00:01<00:03, 11.90it/s]extracted_communities: 646.04 MB, 7356 communities
```

### DIFF
--- a/sentence_transformers/util/retrieval.py
+++ b/sentence_transformers/util/retrieval.py
@@ -333,7 +333,9 @@ def community_detection(
                         sort_max_size = min(2 * sort_max_size, len(embeddings))
                         top_val_large, top_idx_large = cos_scores[i].topk(k=sort_max_size, largest=True)
 
-                    extracted_communities.append(top_idx_large[top_val_large >= threshold].cpu().numpy().astype(np.uint32))
+                    extracted_communities.append(
+                        top_idx_large[top_val_large >= threshold].cpu().numpy().astype(np.uint32)
+                    )
 
     # Largest cluster first
     extracted_communities = sorted(extracted_communities, key=lambda x: len(x), reverse=True)
@@ -353,8 +355,7 @@ def community_detection(
             extracted_ids.update(non_overlapped_community)
 
     unique_communities = [
-        [int(idx) for idx in community]
-        for community in sorted(unique_communities, key=lambda x: len(x), reverse=True)
+        [int(idx) for idx in community] for community in sorted(unique_communities, key=lambda x: len(x), reverse=True)
     ]
 
     return unique_communities
--- a/sentence_transformers/util/retrieval.py
+++ b/sentence_transformers/util/retrieval.py
@@ -352,6 +352,9 @@ def community_detection(
             unique_communities.append(non_overlapped_community)
             extracted_ids.update(non_overlapped_community)
 
-    unique_communities = sorted(unique_communities, key=lambda x: len(x), reverse=True)
+    unique_communities = [
+        [int(idx) for idx in community]
+        for community in sorted(unique_communities, key=lambda x: len(x), reverse=True)
+    ]
 
     return unique_communities
--- a/sentence_transformers/util/retrieval.py
+++ b/sentence_transformers/util/retrieval.py
@@ -317,7 +317,7 @@ def community_detection(
 
             # Use the row-wise count to slice the indices
             for count, indices in zip(row_wise_count, top_k_indices):
-                extracted_communities.append(indices[:count].tolist())
+                extracted_communities.append(indices[:count].cpu().numpy().astype(np.uint32))
         else:
             # Minimum size for a community
             top_k_values, _ = cos_scores.topk(k=min_community_size, largest=True)
@@ -333,7 +333,7 @@ def community_detection(
                         sort_max_size = min(2 * sort_max_size, len(embeddings))
                         top_val_large, top_idx_large = cos_scores[i].topk(k=sort_max_size, largest=True)
 
-                    extracted_communities.append(top_idx_large[top_val_large >= threshold].tolist())
+                    extracted_communities.append(top_idx_large[top_val_large >= threshold].cpu().numpy().astype(np.uint32))
 
     # Largest cluster first
     extracted_communities = sorted(extracted_communities, key=lambda x: len(x), reverse=True)
--- a/sentence_transformers/util/retrieval.py
+++ b/sentence_transformers/util/retrieval.py
@@ -315,9 +315,10 @@ def community_detection(
             k = row_wise_count.max()
             _, top_k_indices = cos_scores.topk(k=k, largest=True)
 
-            # Use the row-wise count to slice the indices
-            for count, indices in zip(row_wise_count, top_k_indices):
-                extracted_communities.append(indices[:count].cpu().numpy().astype(np.uint32))
+            # Move the batch indices to CPU once, then slice per community
+            top_k_indices = top_k_indices.cpu()
+            for count, indices in zip(row_wise_count.tolist(), top_k_indices):
+                extracted_communities.append(indices[:count].numpy().astype(np.uint32))
         else:
             # Minimum size for a community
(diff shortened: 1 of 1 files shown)


tomaarsen (changes requested):
Hello!

This looks like a solid improvement, but I do want to be wary that we're not updating the output dtype of the function, or we risk breaking people's integrations. I didn't run this myself, but I think it might now output a `list[list[np.uint32]]` instead of a `list[list[int]]`. It should be as simple as converting back to `int` at line 349, I think. 

P.s. again: I didn't run this, so feel free to let me know if the above is wrong 🤗 

- Tom Aarsen

AbdelRahmanYaghi:
Hello Tom,

Apologies for missing the output type. I've updated the function as you suggested, and here are the results of the pytest:

```
================================================================== test session starts ===================================================================
platform linux -- Python 3.10.15, pytest-9.1.1, pluggy-1.6.0
rootdir: /home/user/sentence-transformers
configfile: pyproject.toml
plugins: env-1.6.0, xdist-3.8.0, cov-7.1.0, hypothesis-6.155.7
collected 10 items / 1 deselected / 9 selected                                                                                                           

tests/util/test_retrieval.py ........s                                                                                                             [100%]

==================================================================== warnings summary ====================================================================
.venv/lib/python3.10/site-packages/torch/cuda/__init__.py:187
  /home/user/sentence-transformers/.venv/lib/python3.10/site-packages/torch/cuda/__init__.py:187: UserWarning: CUDA initialization: The NVIDIA driver on your system is too old (found version 12040). Please update your GPU driver by downloading and installing a new version from the URL: http://www.nvidia.com/Download/index.aspx Alternatively, go to: https://pytorch.org to install a PyTorch version that has been compiled with your version of the CUDA driver. (Triggered internally at /pytorch/c10/cuda/CUDAFunctions.cpp:119.)
    return torch._C._cuda_getDeviceCount() > 0

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================= 8 passed, 1 skipped, 1 deselected, 1 warning in 1.00s ==================================================
```

tomaarsen:
Hello @AbdelRahmanYaghi!

Thanks for this. Indeed, `extracted_communities` seems to balloon the memory in `community_detection`, and storing each community as a compact `uint32` array rather than a list of Python ints is a solid fix.

I spent a bit of time on the runtime side as well, and pushed two changes on top of yours.

Some context on why first: switching `.tolist()` to numpy helps Step 1, but on its own it's roughly runtime-neutral, and a bit slower when communities are large. The reason is Step 2 (the overlap removal): it walks each community element by element, and iterating a numpy array boxes a fresh `np.uint32` per element, which is slower than the original list-of-ints loop. Once the data is numpy though, we can speed up both steps:

1. Step 1, CUDA path: move the batch's `top_k_indices` to CPU once, then slice per community, instead of calling `.cpu()` inside the loop. On CUDA every per-community `.cpu()` forces a device sync, so a batch with thousands of communities pays thousands of tiny syncs.

2. Step 2, overlap removal: replace the Python `set` and element-wise loop with a dense boolean `used` mask, so the dedup is vectorized (`community[~used[community]]`). As a bonus this is lighter on memory than the `set` of boxed ints.

The return type stays `list[list[int]]` via a `.tolist()` at the end, so nothing downstream changes.

Results (RTX 3090, `all-MiniLM-L6-v2`):

| Scenario | `main` | this PR | speedup |
| --- | --- | --- | --- |
| 50k Quora questions (the `fast_clustering` example) | 226 ms | 197 ms | 1.15x |
| 50k, dense communities | 2465 ms | 650 ms | 3.8x |
| CPU, dense communities | 3457 ms | 2818 ms | 1.23x |

I had an agent check the output against `main` across 243 seeded configurations (both branches, a range of sizes / thresholds / `min_community_size`, plus a numpy input) and it's identical, including the order of the communities and the central point within each. The existing tests still pass.

Thanks again for getting this going! What do you think of these changes?

- Tom Aarsen

tomaarsen (approved):

AbdelRahmanYaghi:
@tomaarsen Yep, this makes perfect sense.

 I just wanted to confirm how much overhead `.cpu()` actually causes, and hence I ran this tiny experiment:

```
import torch
import random
import time

SIZE = 1_000_000
CHUNKS = 100

a = torch.tensor([random.random() for _ in range(SIZE)], device='cuda')
b = torch.tensor([[random.random() for u in range(CHUNKS)] for _ in range(int(SIZE/CHUNKS))], device='cuda')

print(f"`a` shape: {a.shape}")
print(f"`b` shape: {b.shape}")

start = time.time()
a.cpu()
print(f"`a` to cpu() took {round(time.time() - start, 3)}")

start = time.time()
for b_x in b:
  b_x.cpu()
print(f"`b` to cpu() took {round(time.time() - start, 3)}")
```

And got:

```
`a` shape: torch.Size([1000000])
`b` shape: torch.Size([10000, 100])
`a` to cpu() took 0.002
`b` to cpu() took 0.176
```

Never realized that `.cpu()` had this much overhead. 

Regardless, looks great.

- Abdelrahman Yaghi

tomaarsen:
@AbdelRahmanYaghi Thanks for the quick test, I didn't realise `.cpu()` was that slow initially either. I'll merge this once the tests are green 🤗 

- Tom Aarsen

PR #3833, opened 2026-06-24
Make SentenceTransformer.config settable (fixes #3830)

### LINKED ISSUE #3830: Optimum ONNX Export with SentenceTransformer results in property 'config' has no setter
Hi there,

the following command to export a model worked fine until recently:
`optimum-cli export onnx --framework pt --task sentence-similarity --optimize O3 --model intfloat/multilingual-e5-small model`

~~~
pip show accelerate optimum onnx onnxruntime sentence-transformers
Name: accelerate
Version: 1.14.0
...
---
Name: optimum
Version: 2.1.0
...
---
Name: onnx
Version: 1.22.0
...
---
Name: onnxruntime
Version: 1.27.0
...
---
Name: sentence-transformers
Version: 5.4.1
...
~~~

Works fine with `sentence-transformers==5.4.1`.

However with `sentence-transformers==5.5.0`, results in:
~~~
  File ".local/bin/optimum-cli", line 6, in <module>
    sys.exit(main())
             ~~~~^^
  File ".local/lib/python3.14/site-packages/optimum/commands/optimum_cli.py", line 219, in main
    service.run()
    ~~~~~~~~~~~^^
  File ".local/lib/python3.14/site-packages/optimum/commands/export/onnx.py", line 264, in run
    main_export(
    ~~~~~~~~~~~^
        model_name_or_path=self.args.model,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<20 lines>...
        **input_shapes,
        ^^^^^^^^^^^^^^^
    )
    ^
  File ".local/lib/python3.14/site-packages/optimum/exporters/onnx/__main__.py", line 325, in main_export
    model = TasksManager.get_model_from_task(
        task,
    ...<12 lines>...
        **loading_kwargs,
    )
  File ".local/lib/python3.14/site-packages/optimum/exporters/tasks.py", line 1222, in get_model_from_task
    TasksManager.standardize_model_attributes(model, library_name=library_name)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".local/lib/python3.14/site-packages/optimum/exporters/tasks.py", line 1043, in standardize_model_attributes
    model.config = model[0].auto_model.config
    ^^^^^^^^^^^^
  File ".local/lib/python3.14/site-packages/torch/nn/modules/module.py", line 2074, in __setattr__
    super().__setattr__(name, value)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
AttributeError: property 'config' of 'SentenceTransformer' object has no setter
~~~

Is this a known issue / on purpose? 

Should this be fixed here on the `sentence-transformers` side or is it supposed to be fixed on the `optimum` side?

SriniV-1:
### Summary

`SentenceTransformer.config` was made a **read-only property** (to delegate `model.config` reads for Deepspeed / `transformers.Trainer`). That broke tools which *assign* `model.config = ...` — most notably `optimum` during ONNX export, which now fails with:

```
AttributeError: property 'config' of 'SentenceTransformer' object has no setter
```

This is a regression: `optimum-cli export onnx ... --model intfloat/multilingual-e5-small` worked on `sentence-transformers==5.4.1` and broke on `5.5.0`. Reported in #3830.

### Fix

Add a `config` setter that forwards the assignment to the underlying transformers model (so reads and writes stay consistent), falling back to local storage when there is no underlying transformers model — mirroring the getter's existing `None` handling.

### Tests

Added to `tests/base/test_model.py`:
- `test_config_is_settable` — assigning `model.config` no longer raises and round-trips through the underlying model.
- `test_config_setter_without_underlying_model` — assignment on a StaticEmbedding-only model stores locally instead of raising.

The existing `config` getter tests still pass (delegation unchanged).

Fixes #3830.

### DIFF
--- a/sentence_transformers/base/model.py
+++ b/sentence_transformers/base/model.py
@@ -1582,7 +1582,25 @@ This pull request has been automatically generated to add {self.__class__.__name
         transformers model can be located (e.g. a pure StaticEmbedding-only setup).
         """
         underlying = self.transformers_model
-        return getattr(underlying, "config", None)
+        if underlying is not None:
+            return getattr(underlying, "config", None)
+        return getattr(self, "_config", None)
+
+    @config.setter
+    def config(self, value: PretrainedConfig | None) -> None:
+        """Allow assigning ``model.config``.
+
+        Before the read-only :attr:`config` property was introduced, ``model.config = ...``
+        simply set an instance attribute. Tools such as ``optimum`` rely on this during
+        ONNX export, and broke with ``property 'config' has no setter`` (#3830). Forward the
+        assignment to the underlying transformers model so reads stay consistent, falling
+        back to local storage when there is no underlying model.
+        """
+        underlying = self.transformers_model
+        if underlying is not None:
+            underlying.config = value
+        else:
+            object.__setattr__(self, "_config", value)
 
     @property
     def _target_device(self) -> torch.device:
--- a/tests/base/test_model.py
+++ b/tests/base/test_model.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import copy
 import json
 import logging
 import os
@@ -264,6 +265,28 @@ def test_config_returns_none_when_no_underlying_transformers_model(
     `model.config` returns `None` instead of raising.
     """
     assert static_embedding_model.config is None
+
+
+def test_config_is_settable(stsb_bert_tiny_model: SentenceTransformer) -> None:
+    """`model.config` must be assignable, not just readable (#3830).
+
+    A read-only `config` property broke tools that assign `model.config = ...`
+    (e.g. optimum during ONNX export) with "property 'config' has no setter".
+    The setter forwards to the underlying transformers model, so the value
+    round-trips.
+    """
+    new_config = copy.deepcopy(stsb_bert_tiny_model.config)
+    stsb_bert_tiny_model.config = new_config  # must not raise
+    assert stsb_bert_tiny_model.config is new_config
+    assert stsb_bert_tiny_model.transformers_model.config is new_config
+
+
+def test_config_setter_without_underlying_model(static_embedding_model: SentenceTransformer) -> None:
+    """Assigning `config` on a model with no underlying transformers model stores
+    the value locally instead of raising (mirrors the getter's None handling)."""
+    sentinel = object()
+    static_embedding_model.config = sentinel
+    assert static_embedding_model.config is sentinel
 
 
 def test_sentence_transformer(stsb_bert_tiny_model: SentenceTransformer) -> None:


SriniV-1:
Heads-up: the two failing `Run unit tests (3.13, …, >=5.0.0)` jobs look unrelated to this change. They fail during dependency installation, not in the tests — the resolver selects `llvmlite==0.36.0`, which can't build on Python 3.13:

```
× Failed to build `llvmlite==0.36.0`
RuntimeError: Cannot install on Python version 3.13.14; only versions ... are supported
```

This PR touches only `sentence_transformers/base/model.py` and `tests/base/test_model.py` (no dependency changes), and passes on every other matrix entry — including 3.13 with the lower-bound dependency set. Happy to rebase once a resolver/pin fix lands on `main`.

tomaarsen:
> Heads-up: the two failing `Run unit tests (3.13, …, >=5.0.0)` jobs look unrelated to this change. They fail during dependency installation, not in the tests — the resolver selects `llvmlite==0.36.0`, which can't build on Python 3.13:
> 
> ```
> × Failed to build `llvmlite==0.36.0`
> RuntimeError: Cannot install on Python version 3.13.14; only versions ... are supported
> ```
> 
> This PR touches only `sentence_transformers/base/model.py` and `tests/base/test_model.py` (no dependency changes), and passes on every other matrix entry — including 3.13 with the lower-bound dependency set. Happy to rebase once a resolver/pin fix lands on `main`.

I'm also noticing this yeah, I think a downstream dependency changed/updated, should be unrelated.

- Tom Aarsen

tomaarsen:
Thank you for the very thorough PR.
After comparing the two fixes for #3830, I'm going to build on #3831, which was opened a few days earlier and shares the same core setter (forwarding the assignment to the underlying transformers model).

The one place the two PRs differ is the `_config` fallback plus the getter change, for the case where there is no underlying transformers model. I've decided not to take that part, for two reasons:

- It guards a path that isn't reachable in practice. `optimum` only assigns `model.config` after reading `model[0].auto_model.config`, so an underlying model always exists by that point, and nothing inside Sentence Transformers assigns `config` to a model that lacks one.
- Storing the value locally would conflict with the design from #3764, where a model with no underlying transformers model (a StaticEmbedding-only setup) deliberately has no config and returns `None`. With the fallback, such a model would start reporting a config as soon as something assigned one.

So config-less models will keep returning `None` and simply ignore the assignment.

That said, your tests were genuinely useful, and I'll ask to get that incorporated in #3831. I'll close this in favor of #3831 once that one is merged. Thanks again.

- Tom Aarsen

SriniV-1:
Makes sense — building on #3831 is the right call since it came first, and your reasoning on dropping the `_config` fallback (unreachable in practice + preserving the StaticEmbedding "no config" design from #3764) is sound. Glad the regression tests are useful; happy to see them land in #3831. Thanks for the thorough look!

tomaarsen:
Thanks for your understanding! I've merged #3831, so I'll close this now 🤗 

- Tom Aarsen

PR #3837, opened 2026-06-26
Forward token to AutoProcessor.from_pretrained in Transformer.__init__

### LINKED ISSUE #3836: `token` not forwarded to `AutoProcessor.from_pretrained()` in `Transformer.__init__` — breaks private/gated models since 5.4.0
**Description**

Since 5.4.0, `Transformer.__init__` calls `AutoProcessor.from_pretrained()` to load a processor config, but does not forward the `token` that was passed to `SentenceTransformer()`. This causes a 401 Unauthorized error when loading private or gated Hugging Face models, even when a valid token is supplied.

**To reproduce**

```python

model = SentenceTransformer(
    "<my-repo/my-model>",   # any private/gated repo
    token="<your_valid_hf_token>",
)
```

**Expected behavior**

The model loads successfully, just as it did with sentence-transformers <5.4.0.

**Actual behavior**

```
OSError: <my-repo/my-model> is not a local folder and is not a valid model identifier listed on 'https://huggingface.co/models'
If this is a private repository, make sure to pass a token having permission to this repo either by logging in with `hf auth login` or by passing `token=<your_token>`
```

**Full traceback (abridged):**

```
File ".../sentence_transformers/base/modules/transformer.py", line 662, in __init__
    self.processor = AutoProcessor.from_pretrained(
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
huggingface_hub.errors.RepositoryNotFoundError: 401 Client Error.
Repository Not Found for url: .../processor_config.json.
```

The model_weights download (.bin / .safetensors) succeeds with the token, but the AutoProcessor call immediately after does not receive it.

**Environment**

sentence-transformers: 5.4.0 (works in <5.4.0)
transformers: (latest)
huggingface_hub: (latest)
Python: 3.12

**Root cause (likely)**

`SentenceTransformer.__init__` receives token, but when `_load_default_modules` constructs the `Transformer` object, the `token` is not included in the `kwargs` passed to `Transformer.__init__`, so `AutoProcessor.from_pretrained(model_name_or_path)` is called without authentication.

**Workaround**

Set the `HF_TOKEN` environment variable. Because `huggingface_hub` reads this automatically for all its calls, it reaches the `AutoProcessor.from_pretrained()` call even without explicit token forwarding:

**Suggested fix**

Forward the `token` kwarg from `SentenceTransformer` all the way to `AutoProcessor.from_pretrained()` inside `Transformer.__init__`, analogous to how it is already forwarded for the weights download.

WaterXiao-git:
Fix token not being forwarded to AutoProcessor.from_pretrained when loading private/gated HF models.

Since v5.4.0, Transformer.__init__ calls AutoProcessor.from_pretrained() but the auth token may not reach sub-component loads.

Closes #3836

### DIFF
--- a/sentence_transformers/base/modules/transformer.py
+++ b/sentence_transformers/base/modules/transformer.py
@@ -667,9 +667,18 @@ class Transformer(InputModule):
 
         if max_seq_length is not None and "model_max_length" not in processor_kwargs:
             processor_kwargs["model_max_length"] = max_seq_length
+
+        # Ensure the token is explicitly forwarded to AutoProcessor.from_pretrained.
+        # The token may already be in processor_kwargs from the hub_kwargs merge
+        # in _load_init_kwargs or _load_default_modules, but we extract it here
+        # to pass it as an explicit keyword argument for safety, since
+        # AutoProcessor.from_pretrained may not forward all **kwargs to
+        # sub-component loads (tokenizer, feature extractor, image processor).
+        processor_token = processor_kwargs.pop("token", None) or model_kwargs.get("token")
         with suggest_extra_on_exception():
             self.processor = AutoProcessor.from_pretrained(
                 tokenizer_name_or_path if tokenizer_name_or_path is not None else model_name_or_path,
+                token=processor_token,
                 **processor_kwargs,
             )
 


tstadel:
As mentioned in https://github.com/huggingface/sentence-transformers/issues/3836
`processor_kwargs` already contains the token. It was the underlying `AutoProcessor.from_pretrained` of `transformers<5.4.0` which didn't forward the token correctly.

tomaarsen:
Hello!

Based on @tstadel 's findings, I'll be closing this, as it was a dependency issue instead of a bug in Sentence Transformers.

- Tom Aarsen

PR #3838, opened 2026-06-27
Warn when ContrastiveLoss gets non-binary labels

### LINKED ISSUE #3382: Contrastive loss with non-binary labels
Contrastive loss allows for training with non-binary labels under the hood, i.e., it doesn't throw an error if you provide labels like 0.1 or 0.6. Maybe it would be good to catch non-binary labels in the function with an assertion statement to avoid misapplication?

oyinkanchekwas:
## What changed

- Added a one-time warning when `ContrastiveLoss` receives labels outside `0` or `1`.
- Added focused tests for non-binary labels and ordinary binary labels using a tiny fake model.

## Why

`ContrastiveLoss` documents binary labels, but the current implementation still produces a finite loss for dense similarity-style labels. The maintainer discussion on #3382 suggested that this should not necessarily become a hard error, because some existing training workflows may be relying on that behaviour.

This uses a one-time warning instead of an assertion so accidental non-binary labels are visible, while existing runs are not broken or flooded with repeated warnings.

## Testing

- `PYTHONPATH=/private/tmp/st-pytest:$PYTHONPATH /opt/homebrew/bin/python3.11 -m pytest tests/sentence_transformer/losses/test_contrastive.py -q`
- `/opt/homebrew/bin/python3.11 -m py_compile sentence_transformers/sentence_transformer/losses/contrastive.py tests/sentence_transformer/losses/test_contrastive.py`

Fixes #3382

### DIFF
--- a/sentence_transformers/sentence_transformer/losses/contrastive.py
+++ b/sentence_transformers/sentence_transformer/losses/contrastive.py
@@ -84,7 +84,7 @@ class ContrastiveLoss(nn.Module):
         self.margin = margin
         self.model = model
         self.size_average = size_average
-        self._warned_non_binary_labels = False
+        self._checked_labels = False
 
     def get_config_dict(self) -> dict[str, Any]:
         distance_metric_name = getattr(self.distance_metric, "__name__", str(self.distance_metric))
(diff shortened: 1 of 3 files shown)


tomaarsen:
Hello!

Thanks for this. I've gone ahead and made a few small adjustments on top of your commit:

- The label check now runs only once, on the first forward. Previously `labels.ne(0).logical_and(labels.ne(1)).any().item()` ran on every forward until a non-binary label showed up, so a run with correct binary labels paid for a `.item()` device sync every single step (a sync that wasn't there before). It's now behind a `_checked_labels` flag that's set on the first call. The trade-off is that non-binary labels appearing only in a later batch won't be caught, but labels are fixed per run in practice, so I think that's a fine compromise.

- Added the same warning to `OnlineContrastiveLoss`. It has the same binary-label contract, and non-binary labels are arguably worse there: `distance_matrix[labels == 0]` / `[labels == 1]` silently drop any pair whose label isn't exactly 0 or 1, so those pairs just vanish from the loss rather than being binarized.

- Tweaked the wording to describe what actually happens: `ContrastiveLoss` treats the labels as if binarized at a 0.5 threshold, and `OnlineContrastiveLoss` ignores the off-spec pairs.

Thanks again for the contribution!

- Tom Aarsen

tomaarsen (approved):

tomaarsen:
Thanks for e568eaeb6bec433fefc1b7773fbd6b3123d9640b, that's indeed correct. I think we should be good to go now!

- Tom Aarsen

Issue #3840, opened 2026-06-27
Passing a list of images as video can work in `transformers` but not in `sentence transformer`


NewUserHa:
When input like this will fail:
```py
emb = model.model.encode_document(
    {
        "video": [
            r"F:\123.png",
            r"F:\123.png",
        ]
    },
    processing_kwargs={...
    },
)
```
because the second function forcely convert it into a pair:
https://github.com/huggingface/sentence-transformers/blob/e6c5b63b240722f24c76d243ef444f9f96818fab/sentence_transformers/base/modules/transformer.py#L963
https://github.com/huggingface/sentence-transformers/blob/e6c5b63b240722f24c76d243ef444f9f96818fab/sentence_transformers/base/modality.py#L411-L439

But if the list in the dict had 3 paths instead of 2 paths of image, then it will success and pass it as video to `transformers`. So this could consider it's a bug.

However, the underhood transformers line 284 support video as a list of images:
https://github.com/huggingface/transformers/blob/4fd7f1a05f52a11d7709747be554ca50294bf03f/src/transformers/video_processing_utils.py#L258-L296
> `# Videos sometimes are passed as a list of image URLs, especially through templates`

Hope sentence transformers can fix this.

tomaarsen:
Hello!

Thank you for opening this. It should have been resolved now by @yushuosun 's PR in #3840, and will be fixed fully in the next release.

- Tom Aarsen

NewUserHa:
The PR was being fixed simply using `and len(value) == 2`, what if the user input is video with 1 image, dynamically number of images
```py
emb = model.model.encode_document(
    {
        "video": [
            r"F:\123.png",
        ]
    },
    processing_kwargs={...
    },
)
```
It also should be treated as video in this sematic.

PR #3841, opened 2026-06-29
[Fix] batch_to_message routes non-text 2-element inputs (e.g. 2-frame video) to pair_to_messages (#3840)

yushuosun:
## Motivation

Passing a 2-frame video as a list of image paths works in `transformers` but breaks in `sentence-transformers`: a 2-element list is silently treated as a text query/document pair (#3840). 1-frame and 3+-frame inputs work, which makes the failure look arbitrary.

## Root cause

In `sentence_transformers/base/modality.py`, `batch_to_message()` routes any single-modality input whose value is a 2-element list/tuple of strings to `pair_to_messages()` — without checking the modality:

```python
if len(typed_input) == 1:
    value = next(iter(typed_input.values()))
    if isinstance(value, (tuple, list)) and len(value) == 2 and all(isinstance(v, str) for v in value):
        messages.append(self.pair_to_messages(value))
        continue
```

So `{"video": [path1, path2]}` (a 2-frame video) is mis-handled as a text pair. The sibling routing elsewhere in this module already guards on `mod == "text"`.

## Modifications

`sentence_transformers/base/modality.py` (`batch_to_message`): capture the modality key and only route to `pair_to_messages` when it is `"text"`, matching the existing text-pair guard.

## Duplicate-check

- Track A — `gh api 'repos/huggingface/sentence-transformers/pulls?state=open&per_page=100' --paginate` grepped for `3840` → no open PR references it or touches `batch_to_message`.
- Track B — issue #3840 has no comments and no in-progress claim.

### DIFF
--- a/sentence_transformers/base/modality.py
+++ b/sentence_transformers/base/modality.py
@@ -431,7 +431,12 @@ class InputFormatter:
             # Text pairs (e.g. ("query", "document")) are routed to pair_to_messages instead
             if len(typed_input) == 1:
                 mod, value = next(iter(typed_input.items()))
-                if mod == "text" and isinstance(value, (tuple, list)) and len(value) == 2 and all(isinstance(v, str) for v in value):
+                if (
+                    mod == "text"
+                    and isinstance(value, (tuple, list))
+                    and len(value) == 2
+                    and all(isinstance(v, str) for v in value)
+                ):
                     messages.append(self.pair_to_messages(value))
                     continue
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Thanks a lot for opening this, and thank you @NewUserHa for opening the related issue. I think this fix is clean, I should have been checking for the text modality all along here. I reformatted with `pre-commit run --all` and added some simple tests, should be ready!

- Tom Aarsen

PR #3844, opened 2026-07-01
fix: declare tokenizers as a direct dependency

### LINKED ISSUE #3519: tokenizers isn't listed in dependenxcies in pyproject.toml
It is imported in the module.

Version: 5.1.0

aadhar-build:
Fixes #3519

`sentence_transformers` imports `tokenizers` directly (`Tokenizer`, `normalizers`) in `static_embedding.py`, `base/modules/input_module.py`, and `base/modules/transformer.py`, but `pyproject.toml` only lists `transformers` and relies on it to pull `tokenizers` in transitively.

Added `tokenizers>=0.19` as a direct dependency, matching the same floor that `transformers>=4.41.0` itself requires.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,7 @@ classifiers = [
 ]
 dependencies = [
     "transformers>=4.41.0,<6.0.0",
+    "tokenizers>=0.19",
     "huggingface-hub>=0.23.0",
     "torch>=1.11.0",
     "numpy>=1.20.0",


tomaarsen:
Hello!

I agree, since the Sentence Transformers v5.4 release, I'm using `tokenizers` in more places than before, and my previous comment in https://github.com/huggingface/sentence-transformers/issues/3519#issuecomment-3312328760 is a bit moot. ~~However, I think the `>=0.19.0` feels risky, as I exclusively want to use whatever `tokenizers` `transformers` wants. Because of that, perhaps we should keep the version as loose as possible? What do you think?~~

Eh, perhaps the `>=0.19.0` is fine, we have the same for e.g. `huggingface_hub` and `accelerate`.

- Tom Aarsen

tomaarsen (approved):

PR #3848, opened 2026-07-07
[Fix] Make model.compile() apply to encode/predict

tomaarsen:
Hello!

## Pull Request overview
* Make `model.compile()` apply to inference by running `encode()` / `predict()` through `__call__`
* Document `torch.compile` (including `reduce-overhead` and fixed-shape padding) on the three "Speeding up Inference" pages

## Details

`model.compile()` is currently a no-op for inference. `SentenceTransformer.encode`, `CrossEncoder.predict`, and `SparseEncoder.encode` all call `self.forward(...)` directly, so they never go through `__call__`, which is where `torch.compile` installs its compiled path. In other words, calling `model.compile()` before encoding does nothing. This came up in #3791.

The fix is to run those calls through `self(...)` instead of `self.forward(...)` in all three model classes. There are no forward hooks anywhere in the library and `__call__` resolves to the same `BaseModel.forward`, so the output is bit-identical, but `model.compile(...)` now actually takes effect. I've added a small comment at each call site so it doesn't get "simplified" back, a `BaseModel.compile()` override that documents the `dynamic=True` recommendation, and a regression test per model class that registers a forward hook and asserts it fires during `encode` / `predict`. That hook only fires when the call goes through `__call__`, so the test fails on the old direct-`forward` code.

On top of the fix, I've documented `torch.compile` on the "Speeding up Inference" pages for all three model types: `dynamic=True` for variable-length inputs, `mode="reduce-overhead"` (CUDA graphs) as the real speedup lever, and padding to a fixed length via `processing_kwargs` when you want the stable shapes that CUDA graphs need.

I benchmarked on an RTX 3090 (bf16, batch size 1, varying input lengths) to back the docs:

| model | params | `compile(dynamic=True)` | `compile(mode="reduce-overhead")` |
|-------|--------|-------------------------|-----------------------------------|
| `all-MiniLM-L6-v2` | 23M | ~0.6x | ~1.0x |
| `bge-small-en-v1.5` | 33M | ~1.3x | ~3.0x |
| `modernbert-embed-large` | 395M | ~1.3x | ~3.7x |

`reduce-overhead` is the big lever (~3-3.7x on the larger models), the gain grows with model size, and on a tiny model on a fast GPU plain compilation does nothing (or slightly regresses), since inference there is dominated by tokenization and Python overhead rather than the forward pass. Padding to a large `max_seq_length` is a trap: on `modernbert-embed-large` at 8192 a fixed graph at full length drops to ~0.11x, i.e. about 9x slower than eager. I verified every snippet in the new docs runs and matches the eager output for all three model types.

This is intentionally lighter than #3791, which also adds a bucketed CUDA-graph example to the examples folder.

- Tom Aarsen

### DIFF
--- a/docs/sentence_transformer/usage/efficiency.rst
+++ b/docs/sentence_transformer/usage/efficiency.rst
@@ -163,7 +163,10 @@ If you're using a GPU, then you can use the following options to speed up your i
 
       embeddings = model.encode(sentences, processing_kwargs={"text": {"padding": "max_length", "max_length": 256}})
 
-   Padding up to a large ``max_seq_length`` (for example 8192) makes every call process the full length and is slower
+   ``max_length`` sets the fixed length that shorter inputs are padded up to and longer inputs are truncated down to.
+   It is optional and defaults to the tokenizer's ``model_max_length``.
+
+   Padding up to a large ``model_max_length`` (for example 8192) makes every call process the full length and is slower
    than not compiling at all. CUDA graphs also reuse output buffers, so clone anything you keep across calls (the
    default ``convert_to_numpy=True`` already copies off the GPU and is safe). Compilation is lazy, so warm the model
    up on representative inputs before benchmarking or serving.
--- a/docs/sentence_transformer/usage/efficiency.rst
+++ b/docs/sentence_transformer/usage/efficiency.rst
@@ -148,9 +148,9 @@ If you're using a GPU, then you can use the following options to speed up your i
       sentences = ["This is an example sentence", "Each sentence is converted"]
       embeddings = model.encode(sentences)
 
-   ``dynamic=True`` compiles a single kernel that handles any sequence length. For the largest speedup, use
-   ``mode="reduce-overhead"`` instead: it applies CUDA graphs to remove the per-kernel launch overhead that dominates
-   batch-size-1 inference.
+   ``dynamic=True`` enables dynamic shapes so a compiled graph can handle variable sequence lengths, reducing
+   recompilation when your inputs vary in length. For the largest speedup, use ``mode="reduce-overhead"`` instead:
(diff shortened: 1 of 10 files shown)


sentence_transformers/base/model.py
```diff
@@ -511,6 +511,21 @@ def forward(self, input: dict[str, Tensor], **kwargs) -> dict[str, Tensor]:
             input = module(input, **module_kwargs)
         return input
 
+    def compile(self, *args, **kwargs) -> None:
```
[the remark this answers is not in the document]
tomaarsen: I believe `compile` just returns None

docs/sentence_transformer/usage/efficiency.rst
```diff
@@ -130,6 +130,44 @@ If you're using a GPU, then you can use the following options to speed up your i
...
+   .. code-block:: python
+
+      model.compile(mode="reduce-overhead")
+
+   However, CUDA graphs capture one graph per input shape, so they need stable shapes. Pad every input to a fixed
+   length near your typical length by passing ``padding="max_length"`` through ``processing_kwargs`` when encoding:
+
+   .. code-block:: python
+
+      embeddings = model.encode(sentences, processing_kwargs={"text": {"padding": "max_length", "max_length": 256}})
```
kddubey, line 164: Does `"max_length": 256` also cause longer text to get truncated to `256` tokens?

tomaarsen: Yes, in this case it would. I've not made it very clear here, but you can exclude the `max_length` to just pad to the existing maximum length (but this can be very slow for some models, e.g. if they have an 8k input length).

tomaarsen: I clarified it a bit more in d3fa32d3b907efd45a65c51f80c9e538059664b3

PR #3852, opened 2026-07-08
Fix select_max_active_dims mutating its input tensor in place and correct its docstring

teddytennant:
## What

`select_max_active_dims` in `sentence_transformers/util/tensor.py` no longer mutates the tensor it is given. The final step, which keeps only the top-k values per row, previously used the in-place assignment `embeddings[~mask] = 0` and returned the same object; it now returns a new tensor via `embeddings.masked_fill(~mask, 0)`.

I also corrected the docstring, which was factually wrong:
- The summary and Returns lines claimed the function "creates a sparse tensor" / returns "A sparse tensor". It actually returns a dense tensor of the same shape with the non-top-k values set to zero (`.to_sparse()` is never called). These now describe the dense result.
- `max_active_dims` was typed as `int`, but `None` is accepted (and returns the embeddings unchanged). The type is now `int | None` with a note about the `None` case.

## Why

Because the function returned the same object it zeroed in place, a `torch.Tensor` passed in by an external caller was silently modified. That is surprising for what reads like a pure transformation, and it is inconsistent with the other two branches of the same function: the numpy branch copies through `torch.tensor(...)`, and the `None` branch returns the input untouched.

The only in-tree caller (`SparseEncoder.encode`, `sparse_encoder/model.py`) runs under `torch.inference_mode()` and reassigns the return value, so its observable behavior is unchanged by this fix.

## Testing

Added `test_select_max_active_dims_keeps_top_k_and_does_not_mutate_input` to `tests/util/test_tensor.py`. It checks the top-k selection result, that exactly `max_active_dims` values remain non-zero per row, that the input tensor is left unmodified, and that `max_active_dims=None` returns the input unchanged. The no-mutation assertion fails on the previous in-place implementation and passes with this change.

```
$ pytest tests/util/test_tensor.py -q
2 passed, 1 warning in 0.02s
```

`ruff check` and `ruff format --check` pass on the changed files.

### DIFF
--- a/sentence_transformers/util/tensor.py
+++ b/sentence_transformers/util/tensor.py
@@ -134,36 +134,56 @@ def truncate_embeddings(embeddings: np.ndarray | torch.Tensor, truncate_dim: int
     return embeddings[..., :truncate_dim]
 
 
-def select_max_active_dims(embeddings: np.ndarray | torch.Tensor, max_active_dims: int | None) -> torch.Tensor:
+@overload
+def select_max_active_dims(embeddings: np.ndarray | torch.Tensor, max_active_dims: int) -> torch.Tensor: ...
+
+
+@overload
+def select_max_active_dims(embeddings: np.ndarray, max_active_dims: None) -> np.ndarray: ...
+
+
+@overload
+def select_max_active_dims(embeddings: torch.Tensor, max_active_dims: None) -> torch.Tensor: ...
+
+
+def select_max_active_dims(
+    embeddings: np.ndarray | torch.Tensor, max_active_dims: int | None
+) -> np.ndarray | torch.Tensor:
     """
-    Keeps only the top-k values (in absolute terms) for each embedding and returns a new dense tensor with all other values set to zero.
+    Returns a new tensor with only the top-k values (in absolute terms) of each embedding, all others set to zero.
+
+    The input embeddings are never modified in place.
 
     Args:
-        embeddings (Union[np.ndarray, torch.Tensor]): Embeddings to sparsify by keeping only top_k values.
-        max_active_dims (int | None): Number of values to keep as non-zeros per embedding. If None, the embeddings are returned unchanged.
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Thanks for opening. Indeed, `select_max_active_dims` was updating in place. I've made a few changes on top of yours before merging, so here's a quick rundown.

The main one: `masked_fill(~mask, 0)` does fix the mutation, but it's the expensive way to get there, since it keeps the whole mask apparatus (a full-size bool `mask`, plus `~mask` allocating a second one) and then adds a third full-size dense allocation for the output. `SparseAutoEncoder.top_k` already solves the same problem in the same package with `zeros_like` + `scatter_`, so I've switched to that, and the `mask`/`batch_indices` machinery disappears entirely. The output is bit-exact with yours, but it no longer pays for the fix (3090, `B=256, V=30522, k=256`):

| | peak extra alloc | time |
|---|---|---|
| in-place (before this PR) | 31.3 MiB | 0.899 ms |
| `masked_fill` | 45.7 MiB | 1.015 ms |
| `scatter_` | 31.3 MiB | 0.759 ms |

I also switched `dim=1` to `dim=-1`, because the function crashed outright on a 1D input (`not enough values to unpack (expected 2, got 1)`), which meant `select_max_active_dims(model.encode("one sentence"), 64)` never worked at all. And `max_active_dims=0` now raises a `ValueError` rather than quietly returning all-zero embeddings.

Copilot's note on the return type was fair, and we ship a `py.typed`, so downstream type checkers do take our annotations at their word. I added `@overload`s keyed on `max_active_dims` instead, matching `truncate_embeddings` right above it. Lastly I kept your test and parametrized it over `torch.tensor` and `np.array`, added coverage for the 1D, `None` and non-positive cases, and reworded the comments to be timeless per Copilot's other note.

Thanks again for catching this, it seems to work well now.

- Tom Aarsen

PR #3853, opened 2026-07-09
Fix BinaryClassificationEvaluator KeyError('cosine') with multiple non-cosine metrics

vineethsaivs:
When `BinaryClassificationEvaluator` (or `SparseBinaryClassificationEvaluator`) is given two or more similarity functions via `similarity_fn_names` and cosine is not one of them, calling the evaluator raises `KeyError: 'cosine'`:

```python
from sentence_transformers.evaluation import BinaryClassificationEvaluator

evaluator = BinaryClassificationEvaluator(
    sentences1, sentences2, labels,
    similarity_fn_names=["dot", "euclidean"],   # no cosine
)
evaluator(model)   # KeyError: 'cosine'
```

### Root cause

In `__call__`, when more than one similarity function is requested, the `max_*` metrics are aggregated with:

```python
for metric in scores["cosine"]
```

`scores` is keyed by the requested similarity-function names (built in `compute_metrics`), so `"cosine"` is only present when the user actually requested cosine. With two or more non-cosine functions the subscript raises `KeyError("cosine")`. The single-function path takes the `else` branch and is unaffected, and the default single function is usually cosine, which is why this went unnoticed.

### Fix

Iterate the metric names from any requested function's score dict instead of hard-coding `"cosine"`:

```python
for metric in next(iter(scores.values()))
```

Every per-function score dict is built with the same metric keys, so the aggregation is identical when cosine is present (verified: `max_ap` etc. unchanged) and no longer crashes when it is absent. This branch only runs when at least one function is requested, so `next(iter(...))` is always safe. `SparseBinaryClassificationEvaluator` inherits the fix via `super().__call__`.

### Test

Adds `test_BinaryClassificationEvaluator_multiple_non_cosine_metrics`, which constructs the evaluator with `similarity_fn_names=["dot", "euclidean"]`, stubs the embedding-dependent `compute_metrics` (so no model is needed), and asserts the `max_*` metrics are produced. It raises `KeyError("cosine")` before the change and passes after. ruff is clean.

### DIFF
--- a/sentence_transformers/sentence_transformer/evaluation/binary_classification.py
+++ b/sentence_transformers/sentence_transformer/evaluation/binary_classification.py
@@ -210,7 +210,7 @@ class BinaryClassificationEvaluator(BaseEvaluator):
             metrics.update(
                 {
                     f"max_{metric}": max(scores[short_name][metric] for short_name in scores)
-                    for metric in scores["cosine"]
+                    for metric in next(iter(scores.values()))
                 }
             )
             self.primary_metric = "max_ap"
--- a/tests/sentence_transformer/evaluation/test_binary_classification_evaluator.py
+++ b/tests/sentence_transformer/evaluation/test_binary_classification_evaluator.py
@@ -5,8 +5,10 @@ Tests the correct computation of evaluation scores from BinaryClassificationEval
 from __future__ import annotations
 
 import numpy as np
+import pytest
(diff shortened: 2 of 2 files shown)


tomaarsen (approved):
Hello!

Thanks for this, it's indeed a problematic oversight. I've just simplified the test a bit, but otherwise this is ready to go!

- Tom Aarsen

PR #3854, opened 2026-07-09
Fix MegaBatchMarginLoss forward_mini_batched: KeyError crash and dropped gradients

vineethsaivs:
## Summary

`MegaBatchMarginLoss.forward_mini_batched` (the default `use_mini_batched_version=True` path) was broken in two ways. I originally opened this PR for the dropped-gradient bug; while adding an end-to-end regression test I found the method also crashes on the first forward, so this PR now fixes both.

### 1. Crash: `KeyError` on the `modality` marker

`preprocess()` adds a non-tensor `"modality"` entry (e.g. `"text"`) to each feature dict. `forward_mini_batched` sliced **every** value by batch index (`{k: v[start:end] for k, v in positive.items()}`), so `"text"[0:2]` became `"te"` and the model's modality dispatch raised `KeyError: 'te'` before any real work happened. The fix slices only tensor values and forwards non-tensor entries unchanged, mirroring `_create_minibatch` in the cached MNRL loss.

### 2. Dropped gradients

The per-mini-batch backward was guarded by `if end_idx < len(cos_scores):`, but `cos_scores` has one row per anchor in the current mini-batch, so `len(cos_scores)` equals the mini-batch size and the guard is never `True`. The in-loop `backward()` never fired, and since `losses` is reassigned each iteration only the final mini-batch's graph reached the outer training loop, so every earlier mini-batch contributed nothing to the update. Compared against the loop bound `len(all_positive_emb)` (the full batch size) instead, matching the intent stated in the comment.

## Test

`tests/sentence_transformer/losses/test_mega_batch_margin.py` runs the loss end to end (with dropout disabled for determinism) and asserts the mini-batched gradient equals `num_mini_batches` times the non-mini-batched gradient. That invariant only holds if the method runs at all and every mini-batch contributes: before the fix the test raised `KeyError`, and with only the guard fixed it would fail the scaling check.

### DIFF
--- a/sentence_transformers/sentence_transformer/losses/mega_batch_margin.py
+++ b/sentence_transformers/sentence_transformer/losses/mega_batch_margin.py
@@ -11,6 +11,16 @@ from sentence_transformers import util
 from sentence_transformers.sentence_transformer.model import SentenceTransformer
 
 
+def _slice_features(features: dict[str, Any], start_idx: int, end_idx: int) -> dict[str, Any]:
+    """Slice tensor values along the batch dimension, passing non-tensor values through unchanged.
+
+    ``preprocess`` adds non-tensor entries such as the ``modality`` marker to the feature dict; those
+    must be forwarded to the model as-is rather than sliced like the ``input_ids``/``attention_mask``
+    tensors (slicing the ``"text"`` string would corrupt it into e.g. ``"te"``).
+    """
+    return {key: value[start_idx:end_idx] if isinstance(value, Tensor) else value for key, value in features.items()}
+
+
 class MegaBatchMarginLoss(nn.Module):
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

I mentioned this PR briefly in #3859 and #3862, but in short: this loss would likely benefit from GradCache-style functionality when mini-batching. Because this is far from the first time that's happened, I've chosen to extract the GradCache functionality into a Mixin that losses can subclass from. That's what #3862 is trying to do. I'll probably try to patch MegaBatchMarginLoss via that PR instead. Thank you for pointing out the bug and setting up this PR!

- Tom Aarsen

vineethsaivs:
That sounds great, thank you! A shared GradCache mixin is a much nicer home for the mini-batching than fixing each loss individually, so patching `MegaBatchMarginLoss` through #3862 makes sense to me. Happy to close this PR in favor of that whenever it lands, or leave it open as a reference for the specific bug, whichever you prefer. Thanks for taking it on.

tomaarsen:
Thanks for understanding! I've just merged #3862, which should've also fixed this issue, so I'll close this now.

- Tom Aarsen

PR #3855, opened 2026-07-11
add tests for append_to_last_row

RavSinghChandan:
`append_to_last_row` in `util/misc.py` didn't have any tests, so added a few to `tests/util/test_misc.py`.

it's used to tack extra columns onto the last row of a results csv, and the behaviour that's easy to break in a refactor is the guard: it only writes when there's a header plus at least one data row, otherwise it no-ops and returns False. so the tests cover:

- appending one value to the last data row
- appending multiple values
- header-only file -> returns False, file untouched
- empty file -> returns False, file untouched

ruff check + format are clean.

### DIFF
--- a/tests/util/test_misc.py
+++ b/tests/util/test_misc.py
@@ -92,33 +92,36 @@ def test_import_module_class_local_path_with_trust_does_not_warn(tmp_path):
     assert cls is fake_class
 
 
-def test_append_to_last_row_extends_last_data_row(tmp_path):
+@pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

I think this is pretty low priority, but it doesn't really hurt. I merged the tests into one parametrize, as they were all very very similar.

- Tom Aarsen

tomaarsen:
The test failures seem to be related to torch 2.13 + Windows + BF16 on CPU. I suspect they're missing/not importing a required CPU instruction. I'll leave it be for now.

- Tom Aarsen

PR #3859, opened 2026-07-14
[`feat`] Add `mini_batch_num_tokens`: token-budget mini-batching for cached losses

### LINKED ISSUE #3389: Interest in variable mini-batch size for CMNRL?
Hi,

We're training on inputs that have widely varying lengths. When using flash attention 2, models like ModernBERT can pack inputs together and avoid padding tokens. In order to make the most efficient use of these models it's best to vary the number of examples per mini-batch so that the number of tokens is roughly constant. I have a fork where I've implemented this feature: https://github.com/mermerico/sentence-transformers/tree/packed-mini-batches

Would you be interested in a PR?

mermerico:
Resolves #3389

## Motivation

The cached losses (`CachedMultipleNegativesRankingLoss`, `CachedGISTEmbedLoss`, `CachedSpladeLoss`) split their large batch into mini-batches of a fixed number of sequences, controlled by `mini_batch_size`. When input lengths vary, the work per mini-batch varies with them: a mini-batch of 32 short queries costs a fraction of a mini-batch of 32 long documents. With flash attention input flattening (#3554), compute and memory scale with the number of real tokens, so the cost of a fixed-count mini-batch is directly proportional to however many tokens happen to land in it.

This forces an unpleasant choice when tuning `mini_batch_size` on variable-length data. Sizing it for the longest inputs leaves the GPU badly under-utilized on the short ones, which are typically the majority. Sizing it for the average means an unlucky mini-batch of long documents can OOM hours into a run. In the extreme (see the long-document benchmark below), the worst-case mini-batch is more than 10x the average, and the safe setting wastes most of the hardware.

`mini_batch_num_tokens` replaces the fixed sequence count with a token budget. Each mini-batch greedily packs as many sequences as fit within the budget, so every mini-batch does a near-constant amount of work. The budget can sit close to the GPU's limit and stay there regardless of how lengths are distributed.

## What changed

The mini-batch boundary computation is factored into a module-level helper, `_minibatch_ranges(sentence_feature, mini_batch_size, mini_batch_num_tokens)`, next to the existing `_create_minibatch`. Per-sequence token counts come from `cu_seq_lens_q` for flattened inputs (already cumulative, so each boundary is a bisect away) or from `attention_mask.sum(1).cumsum(0)` for padded inputs.

- `CachedMultipleNegativesRankingLoss` gains `mini_batch_num_tokens: int | None = None`. When set, it overrides `mini_batch_size` for the embedding forward passes. It is also included in `get_config_dict()`, so it shows up in model cards.
- `CachedMultipleNegativesSymmetricRankingLoss` forwards the parameter to its parent.
- `CachedGISTEmbedLoss` and `CachedSpladeLoss` use the same helper, imported the same way they already import `_create_minibatch`.
- Docs: a usage note and example in the Flash Attention tab of `docs/sentence_transformer/usage/efficiency.rst`, plus parameter documentation in all touched docstrings.

The cross-encoder `CachedMultipleNegativesRankingLoss` is not included; it builds mini-batches through a different `predict_minibatch_iter` path and would be a follow-up.

### Semantics and edge cases

- The parameter sits at the end of each `__init__` signature, so existing positional calls keep their meaning.
- A single sequence longer than the budget forms its own mini-batch, so progress is always made and no error is raised.
- Boundaries are computed once in `forward` and replayed in the backward hook, the same way `random_states` is. This matters because modules can modify the features in-place between the two passes: `Pooling` with `include_prompt=False` zeroes prompt tokens in the attention mask, and recomputing boundaries from the mutated mask would misalign the replayed mini-batches with the cached gradients.
- `calculate_loss` still chunks the embedding matrix by `mini_batch_size`; that part is independent of sequence length.
- If the features contain neither `cu_seq_lens_q` nor `attention_mask` (e.g. image-only inputs), a `ValueError` explains that token counts are unavailable.
- `mini_batch_num_tokens <= 0` raises a `ValueError` at construction.

## Benchmarks

Both benchmarks train `answerdotai/ModernBERT-base` (bf16, `attn_implementation="flash_attention_2"`) with `CachedMultipleNegativesRankingLoss` on an RTX 5090 (32 GB), flash-attn 2.8.3, transformers 5.13.1, torch 2.9.1. The timed quantity is `loss(features).backward()`, which covers all three GradCache passes, averaged over the batches after one warmup batch. "Padded" rows set `model[0].unpad_inputs = False`; the rest use flattened inputs. Benchmark script and raw results: https://gist.github.com/mermerico/c062d5fc50958f33bb1a424c7c20001a

### Mixed retrieval data (natural-questions)

Batches of 1024 (query, answer) pairs from [sentence-transformers/natural-questions](https://huggingface.co/datasets/sentence-transformers/natural-questions). Sequence lengths: mean 77 tokens, p50 = 23, p95 = 237, max = 811. This is ordinary retrieval training data, short queries mixed with paragraph-length answers.

| Inputs | Mini-batch setting | Time / batch | Peak observed VRAM | Throughput |
|---|---|---:|---:|---:|
| padded | `mini_batch_size=16` | 14.56 s | 5.6 GB | 10.8k tok/s |
| padded | `mini_batch_size=32` | 11.35 s | 10.5 GB | 13.9k tok/s |
| padded | `mini_batch_size=64` | 10.17 s | 20.2 GB | 15.5k tok/s |
| padded | `mini_batch_size=128` | | OOM | |
| flattened | `mini_batch_size=16` | 8.87 s | 2.3 GB | 17.8k tok/s |
| flattened | `mini_batch_size=32` | 4.51 s | 3.6 GB | 34.9k tok/s |
| flattened | `mini_batch_size=64` | 2.64 s | 5.9 GB | 59.7k tok/s |
| flattened | `mini_batch_size=128` | 2.18 s | 10.8 GB | 72.3k tok/s |
| flattened | `mini_batch_size=256` | 2.07 s | 20.6 GB | 76.1k tok/s |
| flattened | `mini_batch_num_tokens=4096` | 2.92 s | 2.8 GB | 54.0k tok/s |
| flattened | `mini_batch_num_tokens=8192` | 1.88 s | 4.8 GB | 83.8k tok/s |
| flattened | `mini_batch_num_tokens=16384` | 1.86 s | 9.0 GB | 84.9k tok/s |
| flattened | `mini_batch_num_tokens=32768` | 1.98 s | 17.4 GB | 79.6k tok/s |

Benchmark: token-budget vs fixed-size mini-batches

At comparable peak memory (5 to 6 GB), the token budget is 1.4x faster than the best fixed-size setting: `mini_batch_num_tokens=8192` reaches 83.8k tok/s in 4.8 GB, while `mini_batch_size=64` reaches 59.7k tok/s in 5.9 GB. The token budget at 4.8 GB also beats fixed-size at any memory footprint; `mini_batch_size=256` uses 20.6 GB and still tops out at 76.1k tok/s, because its mini-batches remain unbalanced no matter how large they get. Peak memory scales predictably with the budget (2.8 GB at 4k up to 17.4 GB at 32k), so the knob can be tuned once against the GPU and trusted, whereas the memory of a fixed-size setting depends on which sequences happen to land together.

### Long documents with heavy-tailed lengths

When I originally proposed this feature (#3389) I had a corpus whose longest documents are more than 10x the mean, trained at `max_seq_length=8192`. For this benchmark, I generated a synthetic corpus. Each document concatenates randomly sampled natural-questions answers up to a target token count drawn from a lognormal distribution (median 300 tokens, sigma 1.1), clipped at 8192. Document lengths: mean 581 tokens, p50 = 392, p95 = 1771, max = 7993, so the longest document is 13.8x the mean. Batches of 512 (query, document) pairs, flattened inputs throughout.

The worst case of a fixed `mini_batch_size=N` is N x 8192 tokens, so the largest setting guaranteed to fit a given memory limit is very small. A mini-batch of one document costs at most one 8192-token forward pass, about 4.7 GB here. The same guarantee expressed as a token budget is `mini_batch_num_tokens=8192`, which packs an average of 14 documents per mini-batch on this data and only drops to a single document when it actually meets an 8192-token one.

| Mini-batch setting | Worst-case tokens per mini-batch | Time / batch | Peak observed VRAM | Throughput |
|---|---:|---:|---:|---:|
| `mini_batch_size=1` | 8,192 | 75.48 s | 4.7 GB | 4.0k tok/s |
| `mini_batch_size=2` | 16,384 | 35.34 s | 5.5 GB | 8.6k tok/s |
| `mini_batch_size=4` | 32,768 | 17.37 s | 6.1 GB | 17.5k tok/s |
| `mini_batch_size=8` | 65,536 | 9.01 s | 8.3 GB | 33.7k tok/s |
| `mini_batch_size=16` | 131,072 | 5.56 s | 10.7 GB | 54.6k tok/s |
| `mini_batch_size=32` | 262,144 | 4.63 s | 15.7 GB | 65.7k tok/s |
| `mini_batch_num_tokens=8192` | 8,192 | 3.58 s | 4.8 GB | 84.9k tok/s |
| `mini_batch_num_tokens=16384` | 16,384 | 3.62 s | 9.0 GB | 83.9k tok/s |
| `mini_batch_num_tokens=32768` | 32,768 | 3.92 s | 17.4 GB | 77.6k tok/s |
| `mini_batch_num_tokens=49152` | 49,152 | 4.10 s | 25.7 GB | 74.0k tok/s |

benchmark_mini_batch_num_tokens_long

Comparing settings with the same worst-case memory: `mini_batch_num_tokens=8192` is 21x faster than `mini_batch_size=1` (84.9k vs 4.0k tok/s at 4.7 GB), and `mini_batch_num_tokens=16384` is 9.8x faster than `mini_batch_size=2`, the largest fixed setting whose worst case fits a 16 GB card. The observed VRAM peaks for the fixed settings understate their true risk: they reflect the two batches that were sampled, while the worst-case column is what a long training run will eventually hit. `mini_batch_size=32` measured 15.7 GB here, but its worst case is 262k tokens, far beyond any of these budgets, and it is still 23% slower than the 4.8 GB token-budget setting.

## Usage

```python

model = SentenceTransformer(
    "answerdotai/ModernBERT-base",
    model_kwargs={"attn_implementation": "flash_attention_2", "torch_dtype": "bfloat16"},
)
# Pack mini-batches up to ~32k real tokens instead of a fixed number of sequences
loss = losses.CachedMultipleNegativesRankingLoss(model, mini_batch_num_tokens=32768)
```

Pick the budget the way you would pick `mini_batch_size`: as high as the GPU allows. Since memory now scales with the budget rather than with luck, a value tuned on one batch holds for the whole run. A reasonable starting point is the largest `mini_batch_size` that survives your longest inputs, multiplied by your maximum sequence length.

The gains are largest when input lengths vary a lot and the model runs without padding (flash attention input flattening, or models such as ModernBERT that unpad internally). With near-uniform lengths it behaves like an auto-tuned `mini_batch_size`. For padded execution the budget counts non-padding tokens, which still tracks compute for models that unpad internally; for strictly padded models `mini_batch_size` remains the more faithful memory knob.

## Tests

- `test_cmnrl_same_grad` is parametrized over `{"mini_batch_size": 2}` and `{"mini_batch_num_tokens": 12}`; the token-budget path produces gradients identical to plain `MultipleNegativesRankingLoss`.
- New `TestMinibatchRanges` unit tests: fixed-size splitting unchanged, padded and flattened inputs produce identical boundaries, padding does not count toward the budget, oversized sequences get their own mini-batch, budgets larger than the batch, exact batch coverage, positional argument compatibility, and both `ValueError` paths.
- `test_cmnrl_token_budget_ranges_survive_mask_mutation` covers the in-place mask mutation case: with `include_prompt=False` and a prompt, the backward hook must replay the boundaries computed from the original mask. Without the caching, this scenario fails with a tensor size mismatch in the gradient replay.
- `tests/sentence_transformer/losses/test_cmnrl.py` and `test_cached_gist_embed.py` pass; `CachedSpladeLoss` was smoke-tested with a token budget on GPU (forward and backward).

### DIFF
--- a/docs/sentence_transformer/usage/efficiency.rst
+++ b/docs/sentence_transformer/usage/efficiency.rst
@@ -124,6 +124,24 @@ If you're using a GPU, then you can use the following options to speed up your i
 
    Flash Attention 2 with input flattening always outperforms standard Flash Attention 2, while using considerably less VRAM. The gains grow with the variance in input length, with the mixed dataset with wildly varying lengths (10-500 tokens) benefitting the most.
 
+   Input flattening also speeds up training. When training with a cached loss such as
+   :class:`~sentence_transformers.losses.CachedMultipleNegativesRankingLoss`, you can additionally set
+   ``mini_batch_num_tokens`` instead of ``mini_batch_size``. Mini-batches are then packed by total token count
+   rather than by sequence count, so every mini-batch performs a similar amount of work regardless of how
+   sequence lengths are distributed within the batch. This keeps memory usage roughly constant, letting you pick
+   a token budget close to your GPU's limit without risking out-of-memory errors on batches full of long texts,
+   and it can substantially increase training throughput on datasets with varying text lengths:
+
+   .. code-block:: python
+
+      from sentence_transformers import SentenceTransformer, losses
+
+      model = SentenceTransformer(
+          "answerdotai/ModernBERT-base",
+          model_kwargs={"attn_implementation": "flash_attention_2", "torch_dtype": "bfloat16"},
+      )
+      loss = losses.CachedMultipleNegativesRankingLoss(model, mini_batch_num_tokens=32768)
+
    .. seealso::
 
       The `Transformers Attention Interface <https://huggingface.co/docs/transformers/en/attention_interface>`_ documentation
--- a/sentence_transformers/sentence_transformer/losses/cached_gist_embed.py
+++ b/sentence_transformers/sentence_transformer/losses/cached_gist_embed.py
@@ -13,7 +13,7 @@ from transformers import PreTrainedTokenizerBase
 
 from sentence_transformers.sentence_transformer.losses.cached_multiple_negatives_ranking import (
     _create_minibatch,
-    _get_batch_size,
+    _minibatch_ranges,
 )
 from sentence_transformers.sentence_transformer.model import SentenceTransformer
 from sentence_transformers.sentence_transformer.modules import StaticEmbedding
@@ -67,14 +67,18 @@ def _backward_hook(
     """A backward hook to backpropagate the cached gradients mini-batch by mini-batch."""
     assert loss_obj.cache is not None
     assert loss_obj.random_states is not None
+    assert loss_obj.minibatch_ranges is not None
     with torch.enable_grad():
-        for sentence_feature, grad, random_states in zip(sentence_features, loss_obj.cache, loss_obj.random_states):
+        for sentence_feature, grad, random_states, ranges in zip(
+            sentence_features, loss_obj.cache, loss_obj.random_states, loss_obj.minibatch_ranges
+        ):
             for (reps_mb, _, _), grad_mb in zip(
                 loss_obj.embed_minibatch_iter(
                     sentence_feature=sentence_feature,
                     with_grad=True,
                     copy_random_state=False,
                     random_states=random_states,
+                    ranges=ranges,
                 ),
                 grad,
             ):
@@ -101,6 +105,7 @@ class CachedGISTEmbedLoss(nn.Module):
         contrast_anchors: bool = True,
         contrast_positives: bool = True,
         gather_across_devices: bool = False,
+        mini_batch_num_tokens: int | None = None,
     ) -> None:
         """
         This loss is a combination of :class:`GISTEmbedLoss` and :class:`CachedMultipleNegativesRankingLoss`.
@@ -140,6 +145,9 @@ class CachedGISTEmbedLoss(nn.Module):
             gather_across_devices: If True, gather the embeddings across all devices before computing the loss.
                 Recommended when training on multiple GPUs, as it allows for larger batch sizes, but it may slow down
                 training due to communication overhead, and can potentially lead to out-of-memory errors.
+            mini_batch_num_tokens: If set, mini-batches are packed by total (non-padding) token count instead of by
+                sequence count, overriding ``mini_batch_size`` for the embedding forward passes. See
+                :class:`CachedMultipleNegativesRankingLoss` for details. The default is None.
 
         References:
             - Efficient Natural Language Response Suggestion for Smart Reply, Section 4.4: https://huggingface.co/papers/1705.00652
@@ -214,8 +222,12 @@ class CachedGISTEmbedLoss(nn.Module):
                 "Both the training model and the guiding model must use a PreTrainedTokenizer from transformers."
             )
         self.mini_batch_size = mini_batch_size
+        if mini_batch_num_tokens is not None and mini_batch_num_tokens <= 0:
+            raise ValueError("mini_batch_num_tokens must be a positive integer or None.")
+        self.mini_batch_num_tokens = mini_batch_num_tokens
         self.cache: list[list[Tensor]] | None = None
         self.random_states: list[list[RandContext]] | None = None
+        self.minibatch_ranges: list[list[tuple[int, int]]] | None = None
         self.show_progress_bar = show_progress_bar
         self.must_retokenize = (
             model.tokenizer.vocab != guide.tokenizer.vocab or guide.max_seq_length < model.max_seq_length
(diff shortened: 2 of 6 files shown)


tomaarsen:
Hello!

I won't review this in too much detail right now, but I appreciate the benchmarks and details. Overall, I think this is a great direction for the GradCache losses. I also recently saw #3854, which should perhaps also be using GradCache to solve its problems.

Overall, this is making me reconsider the current implementation of these losses: perhaps I should have one GradCache implementation and make it easier for losses to subclass some GradCache Mixin to adopt its functionality. That more generic GradCache mixin can then also adopt this `mini_batch_num_tokens`.

I'll be working on that over the next few days, it seems quite promising.

- Tom Aarsen

mermerico:
Thanks! That sounds great!

tomaarsen:
Thanks for understanding! I've just merged #3862, which incorporates the `mini_batch_num_tokens` feature. Thanks for the proposal! I'll close this now.

- Tom Aarsen

PR #3861, opened 2026-07-14
Fix ReciprocalRankFusionEvaluator primary_metric not being prefixed

vineethsaivs:
### Problem

`ReciprocalRankFusionEvaluator` is the only evaluator that overrides `prefix_name_to_metrics`:

```python
def prefix_name_to_metrics(self, metrics, prefix):
    """Prefix all metric names with the evaluator name."""
    if not prefix:
        return metrics
    return {f"{prefix}_{k}": v for k, v in metrics.items()}
```

The base implementation does three things: prefix each metric key with the name, coerce values with `maybe_to_float`, and **rewrite `self.primary_metric` to the prefixed form**. The override only does the first.

So when the evaluator is given a `name`, `__call__` returns keys like `my_eval_ndcg@10`, but `self.primary_metric` stays the unprefixed `ndcg@10`. The standard access pattern `results[evaluator.primary_metric]` (used by the trainer, and by `tests/sparse_encoder/test_train_stsb.py`) then raises `KeyError`.

```python
evaluator = ReciprocalRankFusionEvaluator(dense_samples, sparse_samples, name="my_eval")
results = evaluator()
results[evaluator.primary_metric]   # KeyError: 'ndcg@10'  (results has 'my_eval_ndcg@10')
```

With `name=""` it works only by accident (no prefixing happens).

### Fix

Remove the override so the class inherits the base `prefix_name_to_metrics`, which prefixes the keys, updates `primary_metric`, and adds the float coercion every sibling evaluator already gets. This matches how every other evaluator behaves and reduces surface.

### Test

Adds `tests/sparse_encoder/test_reciprocal_rank_fusion.py`, which constructs the evaluator from precomputed ranked lists (no model needed) and asserts `evaluator.primary_metric in results` (and that it equals `my_eval_ndcg@10`) for the named case, plus the unnamed case as a guard. The named test fails before this change (`KeyError` / unprefixed `primary_metric`) and passes after.

### DIFF
--- a/sentence_transformers/sparse_encoder/evaluation/reciprocal_rank_fusion.py
+++ b/sentence_transformers/sparse_encoder/evaluation/reciprocal_rank_fusion.py
@@ -337,12 +337,6 @@ class ReciprocalRankFusionEvaluator(BaseEvaluator):
 
         return mrr, ndcg, ap
 
-    def prefix_name_to_metrics(self, metrics: dict[str, float], prefix: str) -> dict[str, float]:
-        """Prefix all metric names with the evaluator name."""
-        if not prefix:
-            return metrics
-        return {f"{prefix}_{k}": v for k, v in metrics.items()}
-
     def get_config_dict(self):
         return {
             "at_k": self.at_k,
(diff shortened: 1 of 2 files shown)


tomaarsen (approved):
Looks good overall, let's do it! Thanks.
We can ignore the CI test failure, I believe it's related to Windows + Torch 2.13 + BF16 on CPU

Issue #3197, opened 2025-01-25
Inconsistent encodings with different batch sizes


finitearth:
### Description
When using the `SentenceTransformer.encode()` method, the resulting embeddings appear to be inconsistent depending on the batch_size parameter used. The same input sentences produce different encodings when processed with different batch sizes, which suggests unexpected behavior that could impact downstream applications.

### Minimal Reproducible Example
```python
from sentence_transformers import SentenceTransformer

# Initialize model
model = SentenceTransformer("all-MiniLM-L6-v2")

# Create identical test sentences
sentences = ["test test"] * 100

# Encode with different batch sizes
encodings_batch_64 = model.encode(sentences, batch_size=64)
encodings_batch_1 = model.encode(sentences, batch_size=1)

# Check if encodings are identical
assert (encodings_batch_64 == encodings_batch_1).all()  # AssertionError
```

### Expected Behavior
The encodings should be identical regardless of the batch size used, as the same model is processing the same input sentences.

### Actual Behavior
The encodings differ when using different batch sizes, suggesting that the batch processing might be affecting the model's output in an unexpected way.

### Environment Information
- sentence-transformers version: 3.4.0
- Python version: 3.12.5
- Operating System: Windows
- Model: all-MiniLM-L6-v2

### Additional Notes
- This behavior could potentially impact applications where consistent embeddings are crucial, such as similarity search or clustering.
- The issue appears to be reproducible with identical input sentences, ruling out any input-dependent variations.

### Questions
1. Is this expected behavior?
2. If not, what could be causing the batch size to affect the encodings?
3. Are there any known workarounds to ensure consistent encodings regardless of batch size?

tomaarsen:
Hello!

This is similar to #3109. Please consider reading my comment there: https://github.com/UKPLab/sentence-transformers/issues/3109#issuecomment-2514161357

Note that the differences should(!) be sufficiently small that this does not affect anything. See for example when I compute the pairwise similarities between the 2 sets of embeddings:
```python
from sentence_transformers import SentenceTransformer

# Initialize model
model = SentenceTransformer("all-MiniLM-L6-v2")

# Create identical test sentences
sentences = ["test test"] * 100

# Encode with different batch sizes
encodings_batch_64 = model.encode(sentences, batch_size=64)
encodings_batch_1 = model.encode(sentences, batch_size=1)

# Check the similarities
similarities = model.similarity_pairwise(encodings_batch_1, encodings_batch_64)
[...]

PR #3865, opened 2026-07-16
[fix] Load trained PEFT adapter weights when loading an adapter model

### LINKED ISSUE #3247: Loading a PEFT model uses `peft.peft_model.PeftModelForFeatureExtraction`; incompatible with ST PEFT Methods

Kaif10:
Fixes #3247

**Root cause.** When loading a PEFT adapter repo (e.g. `SentenceTransformer("sentence-transformers-testing/stsb-bert-tiny-lora")`), the Transformer module passed the `PeftConfig` as `config=` to `AutoModel.from_pretrained` and relied on transformers' native PEFT integration to attach the adapter. That integration attaches the adapter *modules* but, depending on transformers/peft versions, fails to map the checkpoint's `base_model.model.`-prefixed weight keys onto the bare transformers model — so the adapter is left freshly-initialized (zero-effect) and the trained weights are silently discarded. The model loads without error and `has_peft_compatible_model()` returns True, but `encode` returns **base-model embeddings**, ignoring the fine-tuned adapter. (On older transformers this instead surfaced as the reported `PeftModelForFeatureExtraction`/`ValueError`; on current `main` the class issue is resolved but the silent weight-drop remains.)

**Fix.** Load the base model without passing the `PeftConfig`, then re-load the adapter weights explicitly with the `base_model.model.` prefix stripped. The underlying model stays a `PeftAdapterMixin`-compatible transformers model (so `add_adapter`/`load_adapter`/`active_adapters`/… keep working) and the trained weights are now applied. The reload is a no-op when weights were already loaded correctly, and the adapter's `revision`/download kwargs are preserved for fetching the adapter checkpoint.

**Tests.** Added `test_load_peft_adapter_model`, asserting both that the underlying model is `PeftAdapterMixin`-compatible (not a `PeftModel*` wrapper) and that the trained adapter weights are actually loaded (embeddings match a peft `PeftModel` reference and differ from the base model). It fails on `main` and passes with this change; existing PEFT tests (`test_load_adapter_with_revision`, `test_multiple_adapters`) and model-loading tests still pass. `ruff` clean.

One thing worth flagging: the underlying weight-mapping failure looks like it may also be a transformers-side bug in the native PEFT-attach path (a plain `AutoModel.from_pretrained(adapter_repo)` and `base.load_adapter(repo)` both exhibit it in transformers 5.13.1 / peft 0.19.1). This fix is a robust, idempotent workaround on the Sentence Transformers side; happy to open an upstream transformers issue too if you think it's warranted.

*Disclosure: I used an AI coding assistant while preparing this change; the fix and tests were reviewed and verified locally on CPU with the tiny test model (repro, weight-diff, and before/after test runs).*

### DIFF
--- a/sentence_transformers/base/modules/transformer.py
+++ b/sentence_transformers/base/modules/transformer.py
@@ -1721,18 +1721,34 @@ class Transformer(InputModule):
             model_kwargs (dict[str, Any]): Keyword arguments passed to the Hugging Face Transformers model.
         """
         if backend == "torch":
-            # When loading a PEFT model, we load the base model first. The revision
-            # (e.g. "main") refers to the adapter checkpoint, not the base model, so
-            # we must not pass it to the base model's from_pretrained.
+            model_cls = TRANSFORMER_TASK_TO_AUTO_MODEL[transformer_task]
+
             if is_peft_model:
+                # Load the base model and let transformers' native PEFT integration attach the
+                # adapter, so the underlying model is a `PeftAdapterMixin`-compatible transformers
+                # model (e.g. `BertModel`) rather than a `peft.PeftModel*` wrapper. This keeps
+                # `SentenceTransformer.add_adapter`/`load_adapter`/... working. See #3247.
+                #
+                # We deliberately do NOT pass the PeftConfig as `config=`: it is a PeftConfig, not
+                # a PretrainedConfig, and passing it prevents the base model's config from being
+                # resolved correctly. The adapter's `revision` (e.g. "main") refers to the adapter
+                # checkpoint, not the base model, so it must not reach the base model's
+                # from_pretrained; we keep it for reloading the adapter weights afterwards.
+                adapter_download_kwargs = {
+                    key: model_kwargs[key]
+                    for key in ("revision", "cache_dir", "token", "subfolder", "local_files_only")
+                    if key in model_kwargs
+                }
                 model_kwargs.pop("revision", None)
+                model = model_cls.from_pretrained(model_name_or_path, **model_kwargs)
+                self._reload_peft_adapter_weights(model, model_name_or_path, adapter_download_kwargs)
+                return model
 
             if transformer_task == "feature-extraction":
                 model = self._load_encoder_only_model(model_name_or_path, config, **model_kwargs)
                 if model is not None:
                     return model
 
-            model_cls = TRANSFORMER_TASK_TO_AUTO_MODEL[transformer_task]
             return model_cls.from_pretrained(model_name_or_path, config=config, **model_kwargs)
         elif backend == "onnx":
             return load_onnx_model(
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

You're right that the weight-drop is real, but it's an upstream bug rather than ours. No Sentence Transformers involved:

```python
from transformers import AutoModel

m = AutoModel.from_pretrained("sentence-transformers-testing/stsb-bert-tiny-safetensors")
m.load_adapter("sentence-transformers-testing/stsb-bert-tiny-lora")
print(max(p.abs().max().item() for n, p in m.named_parameters() if "lora_B" in n))
# 0.0
```

The trained adapter is discarded and the output is bit-identical to the base model.

The cause is narrower than "depending on transformers/peft versions". `build_peft_weight_mapping` in `peft/utils/transformers_weight_conversion.py` emits a single prefix rename:

```python
new_weight_conversions = [WeightRenaming("base_model.model.model.", "model.")]
```

That only strips the prefix when the base model is named `model`, so OPT/Llama-likes are fine and every BERT-like (`base_model.model.encoder...`) silently fails. It's deterministic per architecture family, and specific to `transformers` 5.x: 4.x stripped `base_model.model.` explicitly in `load_adapter`.

This affects anyone calling `AutoModel.from_pretrained("some-bert-lora")`, not just us, so it should be fixed in `peft` rather than worked around here. Could you perhaps open an issue there with the repro above? I couldn't find an existing one.

- Tom Aarsen

Kaif10:
Thanks Tom, that's a much more precise diagnosis than mine — and you're right that it belongs in `peft`, not here.

Good news: it turns out this is **already fixed on `peft` main**, so no Sentence Transformers change (or new peft issue) is needed. In the released `peft` 0.19.1 that I was testing against, `build_peft_weight_mapping` has only the single rename you quoted:

```python
new_weight_conversions = [WeightRenaming("base_model.model.model.", "model.")]
```

But on current `peft` main it's since gained a general fallback rename:

```python
new_weight_conversions = [
    WeightRenaming("base_model.model.model.", "model."),
    WeightRenaming("base_model.model.", ""),
]
```

That second entry strips the prefix for the BERT-like `base_model.model.encoder...` case. It was added in peft PR #3197 ("FIX Transformers weight conversion regression", commit 2bf7bc22), which isn't in the 0.19.1 tag. I ran your exact repro against peft main (0.19.2.dev0, transformers 5.14.1) and it now loads the adapter correctly:

```
max lora_B abs on peft main: 0.0146   # was 0.0 on 0.19.1
```

So this should resolve on the next `peft` release with no action on the ST side. I'll close this PR — thanks for taking the time to dig into it and point me to the real cause.

tomaarsen:
That's excellent news! Thanks for double-checking.

- Tom Aarsen

PR #3868, opened 2026-07-16
Use a pairwise similarity as the SparseCoSENTLoss default

vineethsaivs:
## Summary

`SparseCoSENTLoss` defaults to `similarity_fct=util.cos_sim`, but CoSENT needs a **pairwise** similarity (one score per input pair). `util.cos_sim` returns a full `(B, B)` matrix instead of a `(B,)` vector, which silently changes the objective rather than raising.

The class already documents the intended default in its own docstring:

> `similarity_fct`: Function to compute the **PAIRWISE** similarity between embeddings. Default is `util.pairwise_cos_sim`.

## Why this is a bug, not a deliberate choice

Four places in the codebase already agree the metric should be `pairwise_cos_sim`:

1. `SparseCoSENTLoss`'s own docstring (quoted above).
2. Its `See also` note: "`SparseAnglELoss` is `SparseCoSENTLoss` with `pairwise_angle_sim` as the metric, rather than **`pairwise_cos_sim`**".
3. The same statement in `sparse_cosine_similarity.py`.
4. The parent `CoSENTLoss` defaults to `util.pairwise_cos_sim`.

`SparseAnglELoss` subclasses `SparseCoSENTLoss` and passes `similarity_fct=util.pairwise_angle_sim` explicitly to `super().__init__`, so it never exercises the broken default. That is likely why this went unnoticed. It also shows a pairwise function is correct for sparse embeddings, and `util.pairwise_cos_sim` has an explicit `if a.is_sparse or b.is_sparse` branch for exactly this case.

## Effect

In `CoSENTLoss.compute_loss_from_embeddings`:

```python
scores = self.similarity_fct(embeddings[0], embeddings[1])
scores = scores * self.scale
scores = scores[:, None] - scores[None, :]
labels = labels[:, None] < labels[None, :]
scores = scores - (1 - labels) * 1e12
```

| step | `pairwise_cos_sim` (intended) | `cos_sim` (current default) |
|---|---|---|
| `similarity_fct(...)` | `(B,)` | `(B, B)` |
| `scores[:, None] - scores[None, :]` | `(B, B)` | `(B, B, B)` |
| `scores - (1 - labels) * 1e12` | `(B, B)` | `(B, B, B)`, mask broadcasts |

The `(B, B)` label mask broadcasts cleanly against the `(B, B, B)` tensor, so nothing fails. The loss just optimizes a different objective: the intended CoSENT terms `s_i - s_j` are never formed, and the label mask no longer lines up with the score pairs it is meant to select. The wrong default is also visible in `get_config_dict()` output as `'similarity_fct': 'cos_sim'`.

This path is live in the documented usage, `SpladeLoss(model=model, loss=SparseCoSENTLoss(model), ...)`, which is the snippet in the class docstring.

## Fix

One identifier, `util.cos_sim` to `util.pairwise_cos_sim`, making the code match its own docstring, the parent and the sibling. `util` is already imported.

## Tests

Added `tests/sparse_encoder/losses/test_sparse_cosent.py` using the existing `splade_bert_tiny_model` fixture (CPU only):

- the default `similarity_fct` is `util.pairwise_cos_sim`
- on sparse embeddings, the default gives the same loss as passing `util.pairwise_cos_sim` explicitly

Both fail before this change and pass after. `tests/sparse_encoder/losses/` is green (12 passed), and `ruff check` / `ruff format` are clean.

### DIFF
--- a/sentence_transformers/sparse_encoder/losses/sparse_cosent.py
+++ b/sentence_transformers/sparse_encoder/losses/sparse_cosent.py
@@ -10,7 +10,7 @@ from sentence_transformers.sparse_encoder.model import SparseEncoder
 
 
 class SparseCoSENTLoss(CoSENTLoss):
-    def __init__(self, model: SparseEncoder, scale: float = 20.0, similarity_fct=util.cos_sim) -> None:
+    def __init__(self, model: SparseEncoder, scale: float = 20.0, similarity_fct=util.pairwise_cos_sim) -> None:
         """
         This class implements CoSENT (Cosine Sentence).
         It expects that each of the inputs consists of a pair of inputs (e.g., texts) and a float valued label,
--- a/tests/sparse_encoder/losses/test_sparse_cosent.py
+++ b/tests/sparse_encoder/losses/test_sparse_cosent.py
@@ -6,32 +6,20 @@ from sentence_transformers import SparseEncoder, util
 from sentence_transformers.sparse_encoder.losses import SparseCoSENTLoss
 
 
-def _sparse_embeddings() -> list[torch.Tensor]:
-    torch.manual_seed(12)
-    return [torch.randn(4, 8).relu().to_sparse(), torch.randn(4, 8).relu().to_sparse()]
-
-
-def test_sparse_cosent_loss_default_similarity_fct_is_pairwise(
-    splade_bert_tiny_model: SparseEncoder,
-) -> None:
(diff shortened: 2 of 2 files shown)


tomaarsen:
Hello!

Well spotted, this is indeed a bug. The fix is also perfect, so I'll merge this once the tests are green. P.s. I did merge the two tests, I thought they didn't have to be separate.

- Tom Aarsen

tomaarsen (approved):

PR #3871, opened 2026-07-17
Include query/corpus prompts in mine_hard_negatives cache key

### LINKED ISSUE #3870: mine_hard_negatives: cache_folder key ignores query_prompt/corpus_prompt, silently returns stale embeddings
### Summary

`mine_hard_negatives`'s optional `cache_folder` feature computes its cache key from the model name and the raw query/corpus texts only, it never includes `query_prompt`, `query_prompt_name`, `corpus_prompt`, or `corpus_prompt_name`. Those four parameters are real arguments that get passed straight into `encode_query`/`encode_document` and materially change the resulting embeddings for instruction-tuned models. Two calls with the same texts but a different prompt silently reuse the first call's now-stale cached embeddings on the second call, with no warning or error.

### Where

`sentence_transformers/util/hard_negatives.py`, in `mine_hard_negatives`:

```python
model_name = model.model_card_data.base_model or ""
query_hash = hashlib.sha256((model_name + "".join(queries)).encode(), usedforsecurity=False).hexdigest()
corpus_hash = hashlib.sha256((model_name + "".join(corpus)).encode(), usedforsecurity=False).hexdigest()

query_cache_file = os.path.join(cache_folder, f"query_embeddings_{query_hash}.npy")
corpus_cache_file = os.path.join(cache_folder, f"corpus_embeddings_{corpus_hash}.npy")

if os.path.exists(query_cache_file):
    query_embeddings = np.load(query_cache_file)
    ...
if os.path.exists(corpus_cache_file):
    corpus_embeddings = np.load(corpus_cache_file)
    ...
```

vs. a few lines later, where the actual embedding calls do use the prompt arguments:

```python
corpus_embeddings = model.encode_document(
    corpus, pool=pool, batch_size=batch_size, normalize_embeddings=True,
    convert_to_numpy=True, show_progress_bar=True,
    prompt_name=corpus_prompt_name, prompt=corpus_prompt,
)
...
query_embeddings = model.encode_query(
    queries, pool=pool, batch_size=batch_size, normalize_embeddings=True,
    convert_to_numpy=True, show_progress_bar=True,
    prompt_name=query_prompt_name, prompt=query_prompt,
)
```

`query_prompt`/`query_prompt_name`/`corpus_prompt`/`corpus_prompt_name` never factor into `query_hash`/`corpus_hash`, so changing only the prompt between two calls with the same `cache_folder` produces the exact same cache filenames, and the second call loads the first call's embeddings from disk without ever calling `encode_query`/`encode_document` again.

### Reproduction

I ran the real, unmodified `mine_hard_negatives` function from the current `main` branch, with a small mock `SentenceTransformer`-like model whose `encode_query`/`encode_document` genuinely produce different embeddings depending on the `prompt` argument (mimicking a real instruction-tuned model). I monkeypatched only the filesystem calls (`os.path.exists`, `os.makedirs`, `numpy.load`, `numpy.save`) to an in-memory dict so the test doesn't touch disk, the `mine_hard_negatives` logic itself runs unmodified:

```python
from datasets import Dataset

# MockModel.encode_query/encode_document deterministically depend on `prompt`
dataset = Dataset.from_dict({
    "anchor": [f"anchor {i}" for i in range(6)],
    "positive": [f"positive {i}" for i in range(6)],
})

common_kwargs = dict(
    dataset=dataset, model=model, anchor_column_name="anchor", positive_column_name="positive",
    num_negatives=2, range_max=5, output_format="triplet", verbose=False,
    cache_folder=r"C:\fake\cache\dir",
)

with mock.patch(..., <in-memory fs>):
    ds_a = mine_hard_negatives(query_prompt="Represent the question: ", **common_kwargs)
    ds_b_cached = mine_hard_negatives(query_prompt="Represent the statement: ", **common_kwargs)

print("Cache files written:", list(cache_store.keys()))
print("ds_a negatives:       ", ds_a["negative"])
print("ds_b (cached) negatives:", ds_b_cached["negative"])
```

Output:
```
Cache files written: ['C:\\fake\\cache\\dir\\query_embeddings_c2d33c...npy', 'C:\\fake\\cache\\dir\\corpus_embeddings_9418e9...npy']
Number of distinct cache files: 2 (expected 4 if prompts were respected: 2 prompts x {query,corpus})

ds_a negatives:        ['positive 2', 'positive 1', 'positive 0', 'positive 3', 'positive 0', 'positive 5', ...]
ds_b (cached) negatives: ['positive 2', 'positive 1', 'positive 0', 'positive 3', 'positive 0', 'positive 5', ...]

Identical negative selections despite different query_prompt: True
```

Only one query-embedding cache file (and one corpus-embedding file) was ever written even though the two calls used different `query_prompt` values, proving the hash collision directly. The second call's mined negatives are identical to the first call's, they should differ, since the mock model's embeddings are prompt-dependent by construction and a fresh (non-cached) run with the second prompt produces different results.

### Expected behavior

Calling `mine_hard_negatives` with a different `query_prompt`/`corpus_prompt` (or `*_prompt_name`) should either invalidate the cache and recompute, or the docs should clearly warn that changing the prompt while reusing `cache_folder` returns stale results. Right now it silently does neither.

### Impact

This is easy to hit in a realistic workflow: iterating on prompt/instruction wording while reusing the same `cache_folder` to avoid recomputing embeddings is exactly the use case `cache_folder` seems designed to support (per its docstring: "useful to speed up the process if you want to reuse the embeddings for multiple runs"). Anyone doing that with a prompt-sensitive model gets silently wrong hard-negative mining results (and silently wrong similarity scores if `output_scores=True`) with no indication anything went wrong.

### Suggested fix

Include the prompt arguments in the hash inputs, e.g.:

```python
query_hash = hashlib.sha256(
    (model_name + repr(query_prompt) + repr(query_prompt_name) + "".join(queries)).encode(),
    usedforsecurity=False,
).hexdigest()
corpus_hash = hashlib.sha256(
    (model_name + repr(corpus_prompt) + repr(corpus_prompt_name) + "".join(corpus)).encode(),
    usedforsecurity=False,
).hexdigest()
```

so a prompt change naturally produces a different cache filename instead of colliding with the previous run's.

### Environment

Reproduced against the current `main` branch source of `sentence_transformers/util/hard_negatives.py` (`git clone` of `main`), Python 3.12.10, Windows. No GPU needed, the bug is in the cache-key logic itself, independent of any specific model backend.

ErenAta16:
Fixes #3870.

## Change

`mine_hard_negatives`'s `cache_folder` key is now derived from `repr((model_name, query_prompt, query_prompt_name))` / `repr((model_name, corpus_prompt, corpus_prompt_name))` concatenated with the queries/corpus, following the approach suggested in the issue. A different prompt now produces a different cache file instead of silently reusing embeddings computed with another prompt.

Also updated the `cache_folder` docstring to describe what actually goes into the hash, since that's user-facing behavior someone might reasonably want to know before relying on the cache across script runs.

## Testing

Added `test_cache_respects_prompt` in `tests/util/test_hard_negatives.py`, right next to the existing `test_cache`. It runs `mine_hard_negatives` twice with the same dataset/cache folder but two different `query_prompt` values and asserts the second run adds a new query cache file rather than reusing the first, catching a hash collision directly rather than only checking the mined results happen to differ.

Ran both against the real `static-retrieval-mrl-en-v1` model (same fixture `test_cache` already uses):
```
tests/util/test_hard_negatives.py::test_cache PASSED
tests/util/test_hard_negatives.py::test_cache_respects_prompt PASSED
```

Also ran the full `tests/util/test_hard_negatives.py` file to check the hash format change doesn't affect anything else that touches `cache_folder`.

One intentional side effect worth flagging: since the hash inputs changed, any embeddings cached by an older version under a `cache_folder` will no longer be found by this version (new hash, new filename), so the first run after upgrading will recompute rather than hit stale cache files. That's the correct behavior here rather than something to work around, but calling it out in case it's worth a changelog note.

### DIFF
--- a/sentence_transformers/util/hard_negatives.py
+++ b/sentence_transformers/util/hard_negatives.py
@@ -393,11 +393,11 @@ def mine_hard_negatives(
 
         model_name = model.model_card_data.base_model or ""
         query_hash = hashlib.sha256(
-            (repr((model_name, query_prompt, query_prompt_name)) + "".join(queries)).encode(),
+            repr((model_name, query_prompt, query_prompt_name, queries)).encode(),
             usedforsecurity=False,
         ).hexdigest()
         corpus_hash = hashlib.sha256(
-            (repr((model_name, corpus_prompt, corpus_prompt_name)) + "".join(corpus)).encode(),
+            repr((model_name, corpus_prompt, corpus_prompt_name, corpus)).encode(),
             usedforsecurity=False,
         ).hexdigest()
 
--- a/sentence_transformers/util/hard_negatives.py
+++ b/sentence_transformers/util/hard_negatives.py
@@ -244,7 +244,9 @@ def mine_hard_negatives(
         cache_folder (str, optional): Directory path for caching embeddings. If provided, the function will save
             ``query_embeddings_{hash}.npy`` and ``corpus_embeddings_{hash}.npy`` under this folder after the first run,
             and on subsequent calls will load from these files if they exist to avoid recomputation. The hashes are
-            computed based on the model name and the queries/corpus. Defaults to None.
+            computed based on the model name, the query/corpus prompts, and the queries/corpus, so changing the prompt
+            arguments invalidates the cache rather than silently returning embeddings from a different prompt.
+            Defaults to None.
         as_triplets (bool, optional): Deprecated. Use `output_format` instead. Defaults to None.
         margin (float, optional): Deprecated. Use `absolute_margin` or `relative_margin` instead. Defaults to None.
 
@@ -390,8 +392,14 @@ def mine_hard_negatives(
         os.makedirs(cache_folder, exist_ok=True)
 
         model_name = model.model_card_data.base_model or ""
-        query_hash = hashlib.sha256((model_name + "".join(queries)).encode(), usedforsecurity=False).hexdigest()
-        corpus_hash = hashlib.sha256((model_name + "".join(corpus)).encode(), usedforsecurity=False).hexdigest()
+        query_hash = hashlib.sha256(
+            (repr((model_name, query_prompt, query_prompt_name)) + "".join(queries)).encode(),
+            usedforsecurity=False,
+        ).hexdigest()
+        corpus_hash = hashlib.sha256(
+            (repr((model_name, corpus_prompt, corpus_prompt_name)) + "".join(corpus)).encode(),
+            usedforsecurity=False,
+        ).hexdigest()
 
         query_cache_file = os.path.join(cache_folder, f"query_embeddings_{query_hash}.npy")
         corpus_cache_file = os.path.join(cache_folder, f"corpus_embeddings_{corpus_hash}.npy")
--- a/tests/util/test_hard_negatives.py
+++ b/tests/util/test_hard_negatives.py
@@ -922,6 +922,37 @@ def test_cache(dataset: Dataset, static_retrieval_mrl_en_v1_model: SentenceTrans
     assert len(result1) == len(result2)
 
 
+def test_cache_respects_prompt(
+    dataset: Dataset, static_retrieval_mrl_en_v1_model: SentenceTransformer, tmp_path: Path
+) -> None:
+    """The cache key must incorporate the prompt arguments, so a different query_prompt/corpus_prompt
+    produces its own cache file rather than silently reusing embeddings computed with a different prompt."""
+    model = static_retrieval_mrl_en_v1_model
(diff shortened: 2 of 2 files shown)


tomaarsen:
I agree with the Copilot review comments, so I've merged those, and afterwards I think this is fully ready. Thanks a bunch, I'll let this get merged once the tests pass.

- Tom Aarsen

ErenAta16:
Thanks for the review and for merging, Tom. Appreciate you taking the time on this one.

PR #3872, opened 2026-07-17
Fix FAISS hard-negative mining retrieving too few candidates

Osamaali313:
## Problem

`mine_hard_negatives` (`sentence_transformers/util/hard_negatives.py`) retrieves the top-scoring corpus candidates per query, then removes each query's own positive(s) so that `range_max` genuine negatives remain. The number retrieved must therefore leave room for the positives.

The **non-FAISS** path does this correctly:

```python
# Keep only the range_max + max_positives highest scores. We offset by max_positives to leave room for the positive pair(s).
chunk_scores, chunk_indices = torch.topk(chunk_scores, k=range_max + max_positives, dim=1)
```

The **FAISS** path offsets by only `1`:

```python
scores, indices = index.search(query_chunk, k=range_max + 1)
```

When a query has multiple positives (`max_positives > 1`, i.e. a dataset with several positive passages per anchor), the FAISS branch retrieves up to `max_positives - 1` fewer candidates than the sequential sibling. After the positives are filtered out, it can then yield fewer hard negatives than requested — and the two `use_faiss=True`/`False` paths produce different results for the same input.

## Fix

Mirror the sibling's offset so both paths retrieve the same number of candidates:

```python
scores, indices = index.search(query_chunk, k=range_max + max_positives)
```

`max_positives` is computed earlier in the function (`max(positives_per_query)`), so it's in scope here. For single-positive datasets (`max_positives == 1`) behavior is unchanged.

### DIFF
--- a/sentence_transformers/util/hard_negatives.py
+++ b/sentence_transformers/util/hard_negatives.py
@@ -470,7 +470,8 @@ def mine_hard_negatives(
         # Iterate over query embeddings in batches so we can track the progress
         for i in trange(0, len(query_embeddings), faiss_batch_size, desc="Querying FAISS index"):
             query_chunk = query_embeddings[i : i + faiss_batch_size]
-            scores, indices = index.search(query_chunk, k=range_max + 1)
+            # Keep the range_max + max_positives highest scores. We offset by max_positives to leave room for the positive pair(s).
+            scores, indices = index.search(query_chunk, k=range_max + max_positives)
             scores_list.append(scores)
             indices_list.append(indices)
         scores = torch.from_numpy(np.concatenate(scores_list, axis=0)).to(device)
--- a/sentence_transformers/util/hard_negatives.py
+++ b/sentence_transformers/util/hard_negatives.py
@@ -553,6 +553,11 @@ def mine_hard_negatives(
         if use_multi_process:
             cross_encoder.stop_multi_process_pool(pool)
 
+    if use_faiss:
+        # FAISS pads short result sets with index -1, which would resolve to the last corpus entry.
+        # Applied after the CrossEncoder rescoring, as that overwrites the score of every candidate.
+        scores[indices == -1] = -float("inf")
+
     if not include_positives:
         # for each query, create a mask that is True for the positives and False for the negatives in the indices
         positive_mask = torch.stack(
--- a/tests/util/test_hard_negatives.py
+++ b/tests/util/test_hard_negatives.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import importlib.util
 import math
 import os
+from collections import defaultdict
 from copy import deepcopy
 from pathlib import Path
 from types import SimpleNamespace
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
Confirmed the discrepancy against current main: `max_positives` is computed at line 338 and already in scope by the FAISS branch at line 473, and the non-FAISS sibling at line 489 already uses `range_max + max_positives` via `torch.topk`. The FAISS branch requesting only `range_max + 1` from `index.search` is a real, minimal, correctly-scoped fix.

One gap worth closing before merge: there's no existing test that combines `use_faiss=True` with multiple positives per query, which is exactly the case this bug affects. `test_faiss` exercises the FAISS path but not multi-positive, and `test_multiple_positives_per_query` exercises multi-positive but doesn't pass `use_faiss=True` (default is `False`), so it goes through the already-correct `torch.topk` branch. Without a test crossing both, this exact regression could reappear silently. Worth adding a case with `use_faiss=True` and a dataset where some queries have 2+ positives, asserting the FAISS and non-FAISS paths return the same negative count for the same input - that's the actual invariant this fix restores.

Osamaali313:
Thanks @ErenAta16 — good call, that crossing was exactly the coverage gap. Added `test_faiss_multiple_positives_matches_non_faiss` (fa13696): it builds a multi-positive dataset (`max_positives == 2`), runs `mine_hard_negatives` with `use_faiss=True` and `use_faiss=False` on the same input, and asserts both paths mine the same number of negatives — the invariant this fix restores. With the old `range_max + 1` the FAISS path could under-fill and diverge; with `range_max + max_positives` the counts match. It's guarded by `skipif(faiss is None)` like `test_faiss`.

ErenAta16:
test_faiss_multiple_positives_matches_non_faiss closes the gap well, asserting the actual invariant (equal negative counts between the two paths) rather than just checking the fix doesn't crash. Good addition, no further concerns from me.

Osamaali313:
Thanks for the thorough review, @ErenAta16 — appreciate you confirming both the fix and the invariant the test locks in.

tomaarsen:
Hello!

Thanks for the fix and the accompanying test, well spotted. I pushed one commit on top fixing a related edge case that the larger `k` makes more likely to trigger: when `k = range_max + max_positives` exceeds the corpus size, FAISS pads the missing results with index `-1`. Python resolves that to `corpus[-1]`, so the last corpus entry could be mined as a "negative" for any query, even one it's a positive of. Those padded slots are now disqualified with `-inf` scores, applied after the CrossEncoder rescoring (which overwrites all candidate scores). Note that this predates your PR (any corpus smaller than `range_max + 1` could already hit it), your change just widened the window a little.

The commit also bumps `num_negatives` to 3 in your parity test so it reaches the tail of the candidate pool: with `num_negatives=1` it passed even without your fix.

Everything is green locally, so I'll merge this in a moment. Thanks again!

- Tom Aarsen

ErenAta16:
Good catch on the -1 padding, that's a real gap our fix would have made easier to hit without actually causing it. Disqualifying those slots with -inf after the CrossEncoder rescoring is the right place for it, since that's the step that would otherwise clobber any earlier guard. Thanks for tracking it down and bumping num_negatives in the parity test so it actually exercises the tail.

PR #3873, opened 2026-07-20
fix: prevent mine_hard_negatives crash when candidates exceed corpus size

Kropiunig:
## Summary

On the default (non-FAISS) path, `mine_hard_negatives` crashes with `RuntimeError: selected index k out of range` whenever the number of requested candidates exceeds the corpus size.

In `sentence_transformers/util/hard_negatives.py` the `use_faiss=False` branch runs:

```python
chunk_scores, chunk_indices = torch.topk(chunk_scores, k=range_max + max_positives, dim=1)
```

`chunk_scores` has one column per corpus entry, and `torch.topk` raises when `k > chunk_scores.size(1)`. `range_max` is either user-supplied or auto-derived (`range_min + num_negatives + max_positives`, or `range_min + num_negatives * 10 + max_positives` when a margin/score filter is set) and is **never** clamped to the corpus size on this path — the only existing cap is the FAISS-only `2048` limit.

## Reproduction

```python
from datasets import Dataset

model = SentenceTransformer("sentence-transformers/static-retrieval-mrl-en-v1")
dataset = Dataset.from_dict({"query": QUERIES, "answer": PASSAGES})  # small corpus

# Uses the docstring's own recommended setting; crashes on a modest corpus:
mine_hard_negatives(dataset, model, num_negatives=10)
# RuntimeError: selected index k out of range
```

This is reachable from the public API with realistic inputs (the docstring recommends `num_negatives=10`, and describes `range_max` as only affecting how many candidates are considered — never warning that it must not exceed the corpus size). The existing test suite works around it, using `range_max=3, # ... to avoid k out of range error` in ~14 places, which suggests the crash is a known limitation rather than intended behavior.

## Why the FAISS branch is unaffected

The `use_faiss=True` branch already supports requesting more candidates than the corpus holds: `index.search(..., k=range_max + max_positives)` returns rows padded with an index of `-1`, and those are later masked to `-inf`. The torch branch simply lacks the equivalent handling.

## Fix

- Cap `k` at the corpus size on the torch path so `torch.topk` never over-requests.
- Pad the candidate set back up to `range_max + max_positives` columns with an index of `-1`, mirroring the FAISS branch, so downstream shapes and masking are unchanged.
- Apply the existing `-1 -> -inf` masking on both branches (it is a no-op when no padding is present).

The result is that a small corpus now mines as many negatives as the corpus allows (matching the FAISS path) instead of crashing.

## Test

Added `test_range_max_larger_than_corpus_does_not_crash`, which requests far more candidates than the (deduplicated) corpus holds on the `use_faiss=False` path using the existing lightweight deterministic model. It raises `RuntimeError` before the fix and passes after.

The full `tests/util/test_hard_negatives.py` suite passes locally (93 passed, 5 skipped).

### DIFF
--- a/sentence_transformers/util/hard_negatives.py
+++ b/sentence_transformers/util/hard_negatives.py
@@ -489,17 +489,29 @@ def mine_hard_negatives(
         scores_list = []
         indices_list = []
         corpus_embeddings = torch.as_tensor(corpus_embeddings)
+        # torch.topk cannot return more candidates than there are corpus entries, so we cap k at the
+        # corpus size. This mirrors the FAISS branch above, which is allowed to request more
+        # candidates than the corpus holds and pads the shortfall with an index of -1.
+        k = min(range_max + max_positives, len(corpus))
         # Compute the similarity scores between the queries and the corpus in batches, to avoid
         # materializing the full (n_queries, n_corpus) similarity matrix
         for i in trange(0, len(query_embeddings), faiss_batch_size, desc="Computing similarity scores"):
             chunk_scores = model.similarity(query_embeddings[i : i + faiss_batch_size], corpus_embeddings).to(device)
 
             # Keep only the range_max + max_positives highest scores. We offset by max_positives to leave room for the positive pair(s).
-            chunk_scores, chunk_indices = torch.topk(chunk_scores, k=range_max + max_positives, dim=1)
+            chunk_scores, chunk_indices = torch.topk(chunk_scores, k=k, dim=1)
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Thanks for this, it's indeed a solid fix. I shrunk the comments a bit, and removed the `range_max=3,` that I added to almost all tests to avoid `k out of range` errors. That probably should've already been an indicator to fix this ages ago, haha.

- Tom Aarsen

PR #3876, opened 2026-07-21
Keep per-sample video_metadata aligned with the batch, raise on conflicting audio sampling rates

### LINKED ISSUE #3874: Per-sample `video_metadata` gets attached to the wrong video when a batch mixes dict-wrapped and raw videos
Hello!

I was trying out the per-sample metadata format that was recommended in #3864, i.e. passing videos as `{"array": ..., "video_metadata": ...}` dicts, and noticed that the metadata can silently end up on the wrong video when not every sample in the batch carries metadata.

The cause: `parse_inputs` collects metadata into a single batch-level list, but `_unwrap_video` only appends an entry for samples that actually have metadata (dict-wrapped inputs and `VideoDecoder` objects). Raw videos (paths, arrays) contribute nothing, so the list comes out shorter than the batch:

https://github.com/huggingface/sentence-transformers/blob/7c383d7fbd9d0b67599fe0fbb494d6e140376d80/sentence_transformers/base/modality.py#L132-L152

That list is then merged into the video kwargs and handed to the processor:

https://github.com/huggingface/sentence-transformers/blob/7c383d7fbd9d0b67599fe0fbb494d6e140376d80/sentence_transformers/base/modules/transformer.py#L950-L951

and on the transformers side `make_batched_metadata` pairs metadata with videos purely by index, without any length check. So the metadata of a later sample gets applied to an earlier one.

Small repro that doesn't need a model:

```python
import numpy as np
from sentence_transformers.base.modality import InputFormatter

formatter = InputFormatter(model_type="qwen3_vl", supported_modalities=["text", "image", "video", "audio"])

raw_video = np.zeros((4, 3, 8, 8), dtype=np.float32)  # no metadata
wrapped_video = {
    "array": np.ones((4, 3, 8, 8), dtype=np.float32),
    "video_metadata": {"fps": 15, "total_num_frames": 4},
}

modality, inputs, extra = formatter.parse_inputs([raw_video, wrapped_video])
print(len(inputs["video"]))
# => 2
print(extra["video"]["video_metadata"])
# => [{'fps': 15, 'total_num_frames': 4}]
```

Two videos, one metadata entry. It belongs to the second video, but the processor will apply it to the first.

Audio has the same problem in a different shape. `_unwrap_audio` stores a single `sampling_rate` for the whole batch, and every dict-wrapped sample just overwrites the previous value:

https://github.com/huggingface/sentence-transformers/blob/7c383d7fbd9d0b67599fe0fbb494d6e140376d80/sentence_transformers/base/modality.py#L115-L129

```python
audio_a = {"array": np.zeros(1600, dtype=np.float32), "sampling_rate": 16000}
audio_b = {"array": np.zeros(4410, dtype=np.float32), "sampling_rate": 44100}

modality, inputs, extra = formatter.parse_inputs([audio_a, audio_b])
print(extra["audio"])
# => {'sampling_rate': 44100}
```

The 16 kHz audio is now processed as if it were 44.1 kHz. Neither case warns or raises, the embeddings are just quietly computed with the wrong metadata/rate.

For video I think the fix is to keep the list aligned with the batch, e.g. append a default entry for samples without metadata (transformers already builds default metadata per video when none is passed, so ST could do the same for the individual samples that lack it). For audio it's less obvious because `sampling_rate` is a scalar in the feature extractor API, so raising on conflicting rates seems reasonable to me. That would at least be better than silently processing with the wrong rate.

Happy to submit a PR for both if you agree with the direction.

Versions: sentence-transformers main (7c383d7), transformers 5.14.1, Python 3.12, Linux

meutsabdahal:
Fixes #3874

`_unwrap_video` only appended a `video_metadata` entry for the samples that carried one, so a batch mixing dict-wrapped and raw videos handed the processor a metadata list shorter than the batch. Since `make_batched_metadata` in transformers pairs metadata with videos purely by index, the metadata of a later sample was silently applied to an earlier one. Audio had the same problem in a different shape: `_unwrap_audio` let every dict-wrapped sample overwrite the batch-level `sampling_rate`, so a batch mixing rates would quietly process all audio at whichever rate came last.

Changes:

- `_unwrap_video` now appends one entry per video (`None` when the sample has no metadata), so the list can never fall out of alignment with the batch.
- A new `_reconcile_video_metadata` step at the end of `parse_inputs` resolves the `None` entries. If no sample carried metadata the key is dropped entirely, which keeps the previous behavior of letting transformers infer its own defaults. Otherwise each `None` is filled with the same defaults transformers would build itself (`total_num_frames` from `len(video)`, `frames_indices`, `fps` and `duration` left as `None`). For a metadata-less path/URL video in a mixed batch the frame count cannot be known without decoding, so that case raises a clear error asking to pass metadata for every video or for none.
- `_unwrap_audio` goes through a new `_record_sampling_rate` helper that raises when two samples disagree, since the feature extractor API only accepts a single rate per batch. That seemed better than silently resampling some of the audio at the wrong rate, but happy to downgrade it to a warning if you prefer.

Added five regression tests to `tests/base/test_modality.py`, covering the mixed wrapped/raw alignment, the key being dropped when no sample has metadata, the path/URL error, and both the conflicting and matching sampling rate cases. Ran `tests/base/` locally (excluding the per-architecture matrix in `tests/base/modules/transformer/`): 469 passed, 15 skipped.

### DIFF
--- a/sentence_transformers/base/modality.py
+++ b/sentence_transformers/base/modality.py
@@ -112,6 +112,22 @@ def _is_non_text_pair(sample: Any) -> bool:
     return True
 
 
+def _record_sampling_rate(sampling_rate: int, extra_modality_kwargs: dict[str, dict[str, Any]]) -> None:
+    """Record the batch-level ``sampling_rate``, raising if samples disagree.
+
+    Feature extractors accept a single ``sampling_rate`` for the whole batch, so conflicting
+    per-sample rates cannot be honored and would silently process some audio at the wrong rate.
+    """
+    existing = extra_modality_kwargs["audio"].get("sampling_rate")
+    if existing is not None and existing != sampling_rate:
+        raise ValueError(
+            f"This batch mixes audio with different sampling rates ({existing} and {sampling_rate}), but the "
+            "processor accepts a single sampling rate per batch. Resample the audio to a common rate, or "
+            "encode each sampling rate in a separate call."
+        )
+    extra_modality_kwargs["audio"]["sampling_rate"] = sampling_rate
+
+
 def _unwrap_audio(audio_value: AudioInput, extra_modality_kwargs: dict[str, dict[str, Any]]) -> Any:
     """Unwrap dict-wrapped audio or an ``AudioDecoder`` into a raw array, collecting ``sampling_rate``.
 
@@ -119,12 +135,12 @@ def _unwrap_audio(audio_value: AudioInput, extra_modality_kwargs: dict[str, dict
     """
     if isinstance(audio_value, dict):
         if "sampling_rate" in audio_value:
-            extra_modality_kwargs["audio"]["sampling_rate"] = audio_value["sampling_rate"]
+            _record_sampling_rate(audio_value["sampling_rate"], extra_modality_kwargs)
         return audio_value["array"]
     if AudioDecoder is not None and isinstance(audio_value, AudioDecoder):
         samples = audio_value.get_all_samples()
         # AudioDecoder returns (channels, samples); mean over channels to get 1D numpy
-        extra_modality_kwargs["audio"]["sampling_rate"] = samples.sample_rate
+        _record_sampling_rate(samples.sample_rate, extra_modality_kwargs)
         return samples.data.mean(dim=0).numpy()
     return audio_value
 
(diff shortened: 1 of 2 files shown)


meutsabdahal:
Hello @tomaarsen can I get a review on this?

tomaarsen:
Hello!

Thanks for the thorough PR, and apologies for the slow review. Merging this.

I think the issue is a bit worse than you even mentioned, e.g. a short `video_metadata` list does not just misattribute metadata, but `_decode_and_sample_videos` zips `videos` with `video_metadata` without `strict`, so when `do_sample_frames` is set the loop stops at the shorter list and the trailing videos are dropped from the batch entirely. Users get back fewer embeddings than they passed in.

On your `sampling_rate` question: let's just keep the raise. Feature extractors take one rate per batch, so a warning would still leave some audio processed at the wrong rate, and the error already says exactly what to do.

I pushed some phrasing nits, `strict=True` on the `zip` in `_reconcile_video_metadata`, and a test for a video passed as a list of frames. I also softened the docstring claim about filling entries with "the same defaults transformers would build". It's close but not exact, since `make_batched_metadata` also sets `height` and `width`. I did look at calling that helper directly instead, but it isn't a public export, it has to be preceded by `make_batched_videos` (otherwise it raises `AttributeError` on a list of frames), and nothing in the processing path reads `height`/`width` anyway.

Thanks again!

- Tom Aarsen

PR #3877, opened 2026-07-21
Fix video_metadata misalignment when a batch mixes metadata-carrying and raw videos

### LINKED ISSUE #3874: Per-sample `video_metadata` gets attached to the wrong video when a batch mixes dict-wrapped and raw videos

winklemad:
Fixes #3874.

`_unwrap_video` only appended a `video_metadata` entry for metadata-carrying videos (dict-wrapped or `VideoDecoder`), so when a batch mixes those with raw videos (paths/arrays) the collected list ends up shorter than the batch. The processor pairs metadata to videos by index, so each metadata entry then lands on the wrong video.

This appends one entry per video — `None` for a raw video — so the list stays index-aligned with the batch, and drops the list entirely when no video carries metadata (leaving the all-raw case behaving exactly as before).

Added a regression test in `tests/base/test_modality.py` asserting the mixed-batch alignment yields `[meta1, None, meta2]`; it fails against the old `[meta1, meta2]` output.

### DIFF
--- a/sentence_transformers/base/modality.py
+++ b/sentence_transformers/base/modality.py
@@ -133,23 +133,28 @@ def _unwrap_video(video_value: VideoInput, extra_modality_kwargs: dict[str, dict
     """Unwrap dict-wrapped video or a ``VideoDecoder`` into a raw array, collecting ``video_metadata``.
 
     Passes through unchanged if ``video_value`` is already a raw array/tensor/URL/path.
+
+    Appends one ``video_metadata`` entry per video (``None`` when the video carries no
+    metadata) so the collected list stays index-aligned with the batch. Otherwise a raw
+    video contributes no entry, the list ends up shorter than the batch, and the processor
+    (which pairs metadata to videos by index) attaches metadata to the wrong video.
     """
+    metadata: Any = None
+    result: Any = video_value
     if isinstance(video_value, dict):
-        if "video_metadata" in video_value:
-            extra_modality_kwargs["video"].setdefault("video_metadata", []).append(video_value["video_metadata"])
-        return video_value["array"]
-    if VideoDecoder is not None and isinstance(video_value, VideoDecoder):
+        metadata = video_value.get("video_metadata")
+        result = video_value["array"]
+    elif VideoDecoder is not None and isinstance(video_value, VideoDecoder):
         frame_batch = video_value.get_frames_in_range(0, len(video_value))
-        extra_modality_kwargs["video"].setdefault("video_metadata", []).append(
-            {
-                "fps": video_value.metadata.average_fps,
-                "total_num_frames": video_value.metadata.num_frames,
-                "duration": video_value.metadata.duration_seconds,
-                "frames_indices": list(range(frame_batch.data.shape[0])),
-            }
-        )
-        return frame_batch.data
-    return video_value
+        metadata = {
+            "fps": video_value.metadata.average_fps,
+            "total_num_frames": video_value.metadata.num_frames,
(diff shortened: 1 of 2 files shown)


winklemad:
Closing this — #3875 and #3876 already cover the same video_metadata batch-alignment fix (both predate this; I missed them when I opened it). Sorry for the duplicate noise, and thanks to @Solaris-star / @meutsabdahal for the fix.

PR #3878, opened 2026-07-21
Fix misaligned results CSV in BinaryClassificationEvaluator

vineethsaivs:
## Summary

`BinaryClassificationEvaluator` writes a results CSV whose header row lists every metric for every requested similarity function, but the data row is misaligned: the `*_accuracy_threshold` and `*_f1_threshold` columns are missing, so every later value lands under the wrong header and the trailing `*_ap`/`*_mcc` columns are left empty.

## Root cause

The data row is assembled by filtering the headers with `header_name.count("_") == 1`:

```python
for header_name in self.csv_headers:
    if header_name.count("_") == 1:
        sim_fct, metric = header_name.split("_", maxsplit=1)
        if sim_fct in scores:
            file_output_data.append(scores[sim_fct][metric])
```

Two of the eight metrics, `accuracy_threshold` and `f1_threshold`, already contain an underscore, so their headers (for example `cosine_accuracy_threshold`) have two underscores and are skipped, while the header row still includes them. The data row therefore has two fewer columns per similarity function than the header, shifting each subsequent value one column left and leaving the final `*_ap`/`*_mcc` cells blank.

With `similarity_fn_names=["cosine"]` the header has 10 columns but the data row has 8, so `cosine_accuracy_threshold` shows the F1 value, `cosine_f1` shows precision, and so on, with `cosine_ap`/`cosine_mcc` empty. The returned `metrics` dict a few lines below is built from all of `scores.items()`, confirming the intent is to emit every metric.

## Fix

Accept any header with at least one underscore (`count("_") >= 1`). Similarity-function names never contain an underscore, so `split("_", maxsplit=1)` still splits correctly into `(sim_fct, metric)`; `epoch`/`steps` (no underscore) stay excluded, and all metric columns, including the `_threshold` ones, are now written.

## Test

Added a parametrized test that runs the evaluator with `output_path` set (for `["cosine"]` and for all four similarity functions), reads the CSV back, and asserts the data row has the same number of columns as the header. It fails on the current code (10 vs 8 columns for a single function) and passes with the fix. The existing evaluator tests are unchanged.

### DIFF
--- a/sentence_transformers/sentence_transformer/evaluation/binary_classification.py
+++ b/sentence_transformers/sentence_transformer/evaluation/binary_classification.py
@@ -183,7 +183,7 @@ class BinaryClassificationEvaluator(BaseEvaluator):
         file_output_data = [epoch, steps]
 
         for header_name in self.csv_headers:
-            if header_name.count("_") == 1:
+            if header_name.count("_") >= 1:
                 sim_fct, metric = header_name.split("_", maxsplit=1)
                 if sim_fct in scores:
                     file_output_data.append(scores[sim_fct][metric])
--- a/sentence_transformers/sentence_transformer/evaluation/binary_classification.py
+++ b/sentence_transformers/sentence_transformer/evaluation/binary_classification.py
@@ -183,7 +183,7 @@ class BinaryClassificationEvaluator(BaseEvaluator):
         file_output_data = [epoch, steps]
 
         for header_name in self.csv_headers:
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

This is a clean fix I think, but I simplified it a bit by sticking a bit closer to the original intention: anything not epoch or steps. I'll merge this when the tests go green.

- Tom Aarsen

Issue #3329, opened 2025-04-15
AdaptiveLayer Evaluation


bobox2997:
In the evaluation sections of the docs i see that there is an option to test the performance while truncating the dimensions (i assume it is to test models trained with MatryoshkaLoss). Is there anything similar that can be used to test models while training with AdaptiveLayerLoss? 
someone has any suggestions/advice about that?

thanks in advance!

tomaarsen:
Hello!

I considered adding this support natively when I added the AdaptiveLayer loss, but it ended up not being as simple as I would have imagined. When Sentence Transformers loads a transformers-based model, it considers it a bit of a black box. This is convenient, as a lot of different architectures supported in `transformers` work out of the box in Sentence Transformers, but it also means that I don't always know what architecture is being used.

And I need to know which architecture is being used to be able to adaptively shrink the number of layers in the underlying model. With other words, the only way I can reasonably add this, is by manually implementing the "path" to the layers of the underlying model for every single `transformers` architecture. I wanted to avoid this, so I didn't implement it in Sentence Transformers.

If you're doing your own evaluation with a model trained with AdaptiveLayerLoss, then I would subclass an existing evaluator, and override the `__call__` method with something like this:
```python

class AdaptiveLayerEmbeddingSimilarityEvaluator(EmbeddingSimilarityEvaluator):
    def __init__(self, num_layers: int, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.num_layers = num_layers

    def __call__(self, model, *args, **kwargs):
        all_model_layers = model[0].auto_model.bert.layers # or whatever path you have to the model layers
        model[0].auto_model.bert.layers = all_model_layers[:self.num_layers]
        output = super().__call__(model, *args, **kwargs)
        model[0].auto_model.bert.layers = all_model_layers
        return output
```

and then you'll lower the number of layers for your evaluation.

- Tom Aarsen

bobox2997:
hi @tomaarsen 

thank you for the reply, now I understand the reason!
I ended up subclassing an evaluator as you suggested and it work fine, thanks!



as i'm here..., I've been experimenting quite a bit with some slightly modified AdaptiveLayerLoss, and wanted to share a couple of observations from my training runs, just to see if they resonate or spark any discussion:

[...]

PR #3880, opened 2026-07-25
[fix] Detach KL teacher & fix per-layer weighting in AdaptiveLayerLoss

### LINKED ISSUE #3757: Question regarding KL Divergence gradient flow in `AdaptiveLayerLoss` (edit: and possible issue with layers weighting schema)
Hi @tomaarsen ,

I’ve been looking through the implementation of `AdaptiveLayerLoss` and had a question regarding the KL divergence computation between the intermediate layers and the final layer. I wanted to check if the current behavior is intended or if I might be misunderstanding a part of the implementation.

### Observation: 
In the `forward` method, the loss calculates `final_embeddings` from the full model to serve as a target for the intermediate layers:

```if self.kl_temperature > 0:    final_embeddings = forward_decorator.get_embeddings()    final_embeddings = F.softmax(final_embeddings / self.kl_temperature, dim=-1)```

Later, this is used as the `target` in `F.kl_div`:

```kl_div_loss = F.kl_div(   F.log_softmax(embeddings / self.kl_temperature, dim=-1),   final_embeddings,   reduction="batchmean",)```

(this seems to be common to both  `AdaptiveLayerLoss` and `Matryoshka2dLoss`)

### My Question: 
currently, `final_embeddings` is not detached from the computation graph. from my understanding, in PyTorch, this means that during the backward pass, gradients will flow not only into the intermediate `embeddings` (the "student", if we frame that as a distillation approach) but also back into the `final_embeddings` (the "teacher").

this creates a bidirectional optimization where the final layer is penalized for being different from the intermediate layers. In most distillation setups, ad far as i know, the teacher's output is typically detached, for obvious reasons. (still, one could argue if this really apply to this context of "self-distillation".

**Is it the intentional design of `AdaptiveLayerLoss` to allow the intermediate layers to influence the final layer's representations/optimization via this KL loss?**

or should the final layer be treated as a fixed reference point, like so?
```final_embeddings = F.softmax(final_embeddings / self.kl_temperature, dim=-1).detach()```

I'm curious to hear the thoughts of the maintainers on whether this bidirectional gradient flow is a feature of the 2DMSE approach or if it's something that might need a `.detach()`.

...a side (speculative) note:

interestingly, the recent 'Starbucks' paper (Zhuang et al., 2024, https://arxiv.org/pdf/2410.13230v1) notes that 2DMSE/ESE training can 'significantly hurt final-layer performance.' Given that they used the sentence-transformers implementation (*'For consistency, we implemented using the sentence transformers library which already included 2DMSE baseline implementations', section 4.2.1*), it's possible that the 'interference' they observed is actually the result of (or has some contribution from) this bidirectional gradient flow caused by the missing .detach() ?

btw, thanks for the great library!

Piero Boboli

*edit: fixed formatting/typo*
*edit: paper reference to v3: https://arxiv.org/pdf/2410.13230v3*

Kaif10:
### Fixes #3757

Two small, tightly-scoped fixes to `AdaptiveLayerLoss`, covering the points that were settled in the issue thread (thanks to @bobox2997 for the report/diagnosis and @wshuai190 and @tomaarsen for the discussion):

**1. Detach the KL self-distillation teacher.** `final_embeddings` (the final-layer output) is the target of the KL-divergence loss that aligns the prior layers with the last layer. It was not detached, so KL gradients flowed back into the final layer — pulling the teacher toward the intermediate students rather than the reverse. This adds `.detach()`, matching the standard stop-gradient teacher setup (agreed by @wshuai190 and @tomaarsen).

**2. Fix the per-layer weighting bracket typo.** The weighting divided by `(1 + layer_idx)` and then by `len(layer_indices)` separately, instead of by the intended bracketed fraction `((1 + layer_idx) / len(layer_indices))` that @tomaarsen identified as the intent in #3329. This only inserts the missing brackets so the existing intended formula is computed correctly (the two forms differ by a factor of `N²`).

**Explicitly out of scope:** I have *not* changed the default weighting *scheme* (linear vs. log vs. uniform). @wshuai190 noted the ESE/v2 paper uses log weighting and @tomaarsen indicated openness to updating the default — that's a separate design decision I've left to the maintainers.

**Tests.** Added two regression tests to `tests/sentence_transformer/losses/test_adaptive_layer.py`:
- the final transformer layer receives **zero** gradient from the KL term (teacher is detached), isolated by zeroing the non-KL weights;
- the total loss matches the intended `((1 + layer_idx) / len(layer_indices))` formula, not the un-bracketed one.

Both fail on `main` and pass with this change. `ruff check` / `ruff format` clean.

*Disclosure: prepared with the help of an AI coding assistant; I've reviewed and verified the change (repro of the gradient leak before/after and the weighting values) locally on CPU.*

### DIFF
--- a/sentence_transformers/sentence_transformer/losses/adaptive_layer.py
+++ b/sentence_transformers/sentence_transformer/losses/adaptive_layer.py
@@ -212,9 +212,7 @@ class AdaptiveLayerLoss(nn.Module):
             loss = self.loss(sentence_features, labels) * self.last_layer_weight
             if self.kl_temperature > 0:
                 final_embeddings = forward_decorator.get_embeddings()
-                # Detach the final-layer embeddings: they act as the (teacher) target for the
-                # self-distillation KL loss below, so no gradient should flow back into the final
-                # layer through them (standard stop-gradient teacher). See #3757.
+                # The final layer is the self-distillation teacher, so keep it out of the KL gradient (#3757)
                 final_embeddings = F.softmax(final_embeddings / self.kl_temperature, dim=-1).detach()
 
             num_layers = transformer_decorator.num_layers
@@ -230,7 +228,7 @@ class AdaptiveLayerLoss(nn.Module):
                 # Add regular loss for each layer by using the cached embeddings of that layer
                 transformer_decorator.set_layer_idx(layer_idx)
                 layer_loss = self.loss(sentence_features, labels)
-                loss = loss + layer_loss / ((1 + layer_idx) / len(layer_indices)) * self.prior_layers_weight
+                loss = loss + layer_loss / (1 + layer_idx) / len(layer_indices) * self.prior_layers_weight
 
                 # and KL-divergence loss between the current layer and the final layer
                 # Note: we use "batchmean" reduction as that aligns with the mathematical definition
--- a/sentence_transformers/sentence_transformer/losses/adaptive_layer.py
+++ b/sentence_transformers/sentence_transformer/losses/adaptive_layer.py
@@ -212,7 +212,10 @@ class AdaptiveLayerLoss(nn.Module):
             loss = self.loss(sentence_features, labels) * self.last_layer_weight
             if self.kl_temperature > 0:
                 final_embeddings = forward_decorator.get_embeddings()
-                final_embeddings = F.softmax(final_embeddings / self.kl_temperature, dim=-1)
+                # Detach the final-layer embeddings: they act as the (teacher) target for the
+                # self-distillation KL loss below, so no gradient should flow back into the final
+                # layer through them (standard stop-gradient teacher). See #3757.
+                final_embeddings = F.softmax(final_embeddings / self.kl_temperature, dim=-1).detach()
 
             num_layers = transformer_decorator.num_layers
             layer_indices = range(num_layers - 1)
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Thanks for the tightly scoped PR, and thanks @bobox2997 and @wshuai190 for the diagnosis in #3757.

Re. the detach: fully agreed, and the regression test is a nice way to pin it down. I'm keeping both.

The weighting brackets: I went back to the sources before merging, and I have to walk back my own comment from #3329. The un-bracketed formula turns out to be faithful to the original works rather than a typo:

* The original 2DMSE reference implementation ([`angle.py` from Feb 2024](https://github.com/SeanLee97/AnglE/blob/347bbb6a9adf8cce293a41ebe462c5ca09b3812d/angle_emb/angle.py#L863)) computes `loss += student_loss / sample_layer` with one uniformly sampled layer per step. Our `layer_loss / (1 + layer_idx) / len(layer_indices)` is that same linear discount, generalized with an average for when multiple layers are sampled per step.
* The ESE paper ([arXiv 2402.14776v2](https://arxiv.org/abs/2402.14776v2)) uses `w_i = 1 / (1 + ln(i))` in Eq. 2 and states that "the weights decrease as the layer depth increases", with the last layer left unweighted. Its reference code also divides the student losses by `n_layers - 1`.
* The latest arXiv version ([v3](https://arxiv.org/abs/2402.14776v3)) reverts to the 2DMSE formulation: one uniformly sampled layer, no depth weighting at all.

So no version of the paper or the reference code ever gives the prior layers more weight than the final layer. The bracketed form would multiply the prior layer term by `len(layer_indices) ** 2`, which is a no-op with the default `n_layers_per_step=1` but roughly 144x for a 12 layer model with `n_layers_per_step=-1`. Apologies for pointing you at those brackets in the first place, that one is on me.

Since you allowed maintainer edits, I took the liberty of pushing a commit that drops the weighting change and its test, so this PR is now exactly the detach fix. I hope you don't mind. 

I'll open a separate PR that makes the weighting configurable via a `layer_weighting` parameter ("uniform", "log", "linear", or a custom callable), as proposed by @bobox2997 in the issue. In early experiments on STSb, "uniform" (which matches the latest 2DMSE version and the Starbucks paper) outperformed "log" and "linear" at every truncated layer count, so it is the planned new default.

Thanks again!

- Tom Aarsen

Issue #3881, opened 2026-07-26
CrossEncoder silently drops the whole text pair when the backbone's chat template ignores the query/document roles


ishan-1010:
## Summary

If a CrossEncoder's backbone ships a chat template that branches on role names and drops the ones it does not know, the text pair is rendered to an empty string and the model gets zero-length input. No warning, and the error that eventually surfaces points somewhere else entirely.

## Reproduction

No configuration, no custom template, a 21MB ungated checkpoint carrying the stock Qwen3 instruct template:

```python
from sentence_transformers import CrossEncoder

model = CrossEncoder("trl-internal-testing/tiny-Qwen3ForCausalLM")
model.predict([("What is the capital of France?", "Paris is the capital of France.")])
```

```
IndexError: index -1 is out of bounds for dimension 1 with size 0
```

Inspecting what reaches the model shows the input is gone before it gets there:

```
modality_config  : ['text', 'message']
template renders query/document roles as: ['']
input_ids.shape  : (2, 0)
decoded[0]       : ''
```

On a sequence-classification reranker given the same kind of template, the symptom is the more familiar one:

```
RuntimeError: cannot reshape tensor of 0 elements into shape [2, 0, -1, 64]
```

## What happens

`infer_modalities` adds `"message"` to `modality_config` for any model whose processor has a `chat_template`. `Transformer.preprocess` then converts to the message format unconditionally ("Always convert to the message format if it's supported"), and for a text pair `InputFormatter.pair_to_messages` assigns the roles `query` and `document`.

Reranker templates select on those roles explicitly. A general instruct template only handles `system`/`user`/`assistant`, so both messages fall through every branch and the template renders `''`. `apply_chat_template` tokenizes with `add_special_tokens=False`, so the result is genuinely zero-length rather than a lone `[CLS] [SEP]` pair. That last part is why the tokenizer family does not matter.

## Scope

Affected: a chat template that branches on role names and drops unrecognised ones. The stock Qwen2/Qwen3 instruct template is the case to care about. Both the sequence-classification and text-generation paths hit it.

Verified unaffected: no template at all (BERT and ModernBERT rerankers, `tomaarsen/Qwen3-Reranker-0.6B-seq-cls`), purpose-built reranker templates (`Qwen/Qwen3-Reranker-0.6B`), and templates that emit the role verbatim rather than branching (`HuggingFaceTB/SmolLM2-135M-Instruct` renders `<|im_start|>query` and keeps the content).

[...]

Issue #3849, opened 2026-07-07
Problem using CrossEncoder.rank with jina-reranker-v3


PeteBleackley:
I've been working on a open source information retrieval system, called [Emily](https://codeberg.org/PeteBleackley/Emily). As the reranker component, I use a `CrossEncoder`. The model used is parameterisable, but as a default I'm using [jina-reranker-v3](https://huggingface.co/jinaai/jina-reranker-v3).

I keep getting 

```
RuntimeError: cannot reshape tensor of 0 elements into shape [32, 0, -1, 128] because the unspecified dimension size -1 can be any value and is ambiguous
```

The information I have discovered so far suggests that empty strings are being fed into the model, but I have found no evidence of this. I have tried various diagnostics to determine why this is happening, but I have drawn a blank. My best guess is that when the final batch is padded to 32 examples, empty strings may be added. 

Can anyone suggest a way to diagnose or fix the problem?

tomaarsen:
Hello!

Thanks for the report. The core problem is that `jina-reranker-v3` is not a sequence-classification cross-encoder, so it can't be driven by the `CrossEncoder` class.

`jina-reranker-v3` is a listwise reranker, instead of a pointwise reranker that `CrossEncoder` expects. Listwise rerankers look at a query and a list of documents together, and produce a single ranking of the documents, while pointwise rerankers look at a query and a single document at a time, and produce a score for that document. Reranking is then simply a matter of sorting the documents by their scores. The `CrossEncoder` class is designed for pointwise rerankers, so it can't handle listwise rerankers like `jina-reranker-v3`.

To use it, call its native API directly:

```python
from transformers import AutoModel

model = AutoModel.from_pretrained("jinaai/jina-reranker-v3", trust_remote_code=True)
results = model.rerank(query, documents, top_n=10)
for r in results:
    print(r["relevance_score"], r["index"], r["document"])
```

If you specifically want to keep using the `sentence-transformers` `CrossEncoder.rank(...)` API in your Emily project (so the reranker stays swappable), pick a model that really is a cross-encoder, for example:

```python
from sentence_transformers import CrossEncoder

model = CrossEncoder("jinaai/jina-reranker-v2-base-multilingual", trust_remote_code=True)
model.rank(query, documents, return_documents=True)
```

[...]

PR #3882, opened 2026-07-27
Don't render text pairs through chat templates that lack the query/document roles

ishan-1010:
## Summary

Refs #3881. A CrossEncoder whose backbone carries a chat template that drops unknown roles renders its text pairs to an empty string, so the model receives zero-length input and every token of the query and document is lost before the forward pass.

```python

model = CrossEncoder("trl-internal-testing/tiny-Qwen3ForCausalLM")   # stock Qwen3 template, 21MB, ungated
model.predict([("What is the capital of France?", "Paris is the capital of France.")])
```

```
main   : IndexError: index -1 is out of bounds for dimension 1 with size 0
         input_ids.shape (2, 0), decoded ''
branch : input_ids.shape (2, 14), decoded 'What is the capital of France?Paris is the capital of France.'
```

Same on a sequence-classification reranker given such a template: `(2, 0)` and `RuntimeError: cannot reshape tensor of 0 elements` on main, `(2, 17)` and `[CLS] what is the capital of france? [SEP] paris is the capital of france. [SEP]` here.

`pair_to_messages` assigns the `query` and `document` roles, which only purpose-built reranker templates select on. A general instruct template falls through every branch and renders `''`, and since `apply_chat_template` tokenizes with `add_special_tokens=False` the result is zero-length for any tokenizer family.

## Change

`Transformer._supports_pair_roles()` renders a sentinel pair through the template once and checks both probe strings survive; `preprocess` keeps its existing message conversion unless the batch is a text pair whose roles would be dropped. Result cached per instance, so one extra template render per model.

It can only skip the message path in the case where content would otherwise be lost, so a false positive leaves current behaviour intact. Verified unchanged for models with no template, purpose-built reranker templates (`Qwen/Qwen3-Reranker-0.6B`), and templates that emit the role verbatim rather than branching (`HuggingFaceTB/SmolLM2-135M-Instruct`).

## Design question

Silent fallback or a raised error? The fallback makes these models work; raising makes the misconfiguration obvious. I went with the fallback since the alternative today is silent data loss either way, but this is your call and I am happy to switch it.

## Tests

Two in `tests/cross_encoder/test_model.py`: one asserting content survives for a template without pair roles, one asserting models with a working reranker template still take the message path.

```
tests/cross_encoder/test_model.py, branch        : 66 passed, 8 skipped
tests/cross_encoder/test_model.py, main + tests  : 1 failed, 65 passed, 8 skipped   (the behavioural test)
tests/base/modules/test_transformer.py + tests/base/test_modality.py : 317 passed, 6 skipped
```

The base suites matter because the change is in shared base code. The one pre-existing failure in `tests/cross_encoder`, `test_cached_ce_mnrl_prompt_reaches_the_backward_pass`, fails the same way on an unmodified checkout.

## Note

This does not make jina-reranker-v3 work and does not try to; you have already covered why that model is out of scope in #3849. It is where the zero-length input was first noticed.

AI-assisted. I reviewed every line, ran the repro and the suites above myself, and corrected the mechanism once when the first explanation did not survive checking.

### DIFF
--- a/sentence_transformers/base/modules/transformer.py
+++ b/sentence_transformers/base/modules/transformer.py
@@ -635,6 +635,7 @@ class Transformer(InputModule):
         self._prompt_length_mapping = {}
         self._method_signature_cache: dict[str, set[str]] = {}
         self._chat_template_suffix_cache: dict[Any, list[int]] = {}
+        self._chat_template_supports_pair_roles: bool | None = None
 
         config, is_peft_model = self._load_config(model_name_or_path, backend, config_kwargs)
         self._warn_on_unsupported_attention_config(config)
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Thanks for opening this. The issue is indeed pretty big, but I'm wary about just ignoring the chat template quietly, so I'd like to firm up the detection and raise instead. I've implemented that in #3902 with some credits to your work here. I'll close this when that PR gets merged.

- Tom Aarsen

tomaarsen:
I've just merged https://github.com/huggingface/sentence-transformers/pull/3902, so I'll close this. Thanks again for the ideas!

- Tom Aarsen

PR #3883, opened 2026-07-27
[`fix`] RRF scoring: only rank retrievers that returned a document, and use 1-based ranks

eSVeeF:
### Problem 

The `ReciprocalRankFusionEvaluator` currently deviates from canonical Reciprocal Rank Fusion in two ways:

1. Ranks are zero-based (`enumerate(dense_docs)`), while RRF convention uses 1-based ranks
2. More importantly, a document missing from a retriever still receives a score contribution by falling back to `len(retriever_docs)` as a synthetic rank. This means documents returned by only one retriever get an artificial boost, while documents returned by both retrievers get less relative benefit than they should

### Reproduction

```python
dense_docs  = ["X", "a", "Y"]
sparse_docs = ["b", "c", "Y"]
positive    = ["Y"]
```

Before this fix, the fused order was `["X", "b", "c", "a", "Y"]`, so `Y` (the only relevant document, returned by both retrievers) fell outside the top-3 and `mrr@3 == 0.0`.

With canonical RRF, the order is `["Y", "X", "b", "a", "c"]` and `mrr@3 == 1.0`.

### Fix

- Build rank maps with `enumerate(..., start=1)` for 1-based ranks
- Only add a reciprocal-rank term when the document is actually present in that retriever's rank map

### Test

Added `test_rrf_only_scores_retrievers_that_returned_a_document` covering the reproduction above and asserting that the shared relevant document tops the fused ranking.

---

### Notes

While working on this I also noticed that tied RRF scores are resolved via `set` iteration order, which can vary across process invocations due to hash randomization. I kept this PR focused on the scoring correctness bug, but I'm happy to add a deterministic tie-break fix here as well if you prefer.


tomaarsen:
Hello!

Thanks for this PR, and for the clear report. I verified the fix: your new test indeed fails on `main` and passes with this change, and the updated scoring matches the canonical RRF formulation (1-based ranks, only summing over the retrievers that actually returned the document). Well spotted!

I took the liberty of pushing two small commits:

1. `Small nits`: a whitespace tweak and a simpler final assert in the new test (`mrr@3` is exactly `1.0` here, so `pytest.approx` isn't needed).
2. `Regenerate retrieve_rerank example scores`: the results table in `examples/sparse_encoder/applications/retrieve_rerank/README.md` was generated with the old scoring, so I reran `hybrid_search.py` on this branch. Reassuringly, the Hybrid RRF row improves (NDCG@10 from 32.62 to 32.95, MRR@10 from 49.63 to 50.19), while the reranked rows are essentially unchanged: exactly what we'd expect from this fix, as the cross-encoder re-scores the same candidates regardless of the fused order.

Once CI is green, this looks good to merge to me. Thanks again for the contribution!

- Tom Aarsen

PR #2966, opened 2024-09-27
[`feat`] Integrate NanoBeIR datasets; use `model.similarity` by default in evaluators

ArthurCamara:
As discussed in https://github.com/UKPLab/sentence-transformers/pull/2848#issuecomment-2373619762, This PR adds a new Evaluator based on the [NanoBEIR collection of datasets](https://huggingface.co/collections/zeta-alpha-ai/nanobeir-66e1a0af21dfd93e620cd9f6).

It creates one `InformationRetrievalEvaluator` for each dataset, and aggregates the results accordingly.

Example:
```python

from sentence_transformers import SentenceTransformer
from sentence_transformers.evaluation import NanoBEIREvaluator

# Load a model
model = SentenceTransformer('all-mpnet-base-v2')

datasets = ["QuoraRetrieval", "MSMARCO"]
query_prompts = {
"QuoraRetrieval": "Instruct: Given a question, retrieve questions that are semantically equivalent to the given question\nQuery: ",
"MSMARCO": "Instruct: Given a web search query, retrieve relevant passages that answer the query\nQuery: "
}

evaluator = NanoBEIREvaluator(
dataset_names=datasets,
name="NanoBEIR",
query_prompts=query_prompts,
)

results = evaluator(model)
'''
NanoBEeIR Evaluation of the model on ['QuoraRetrieval', 'MSMARCO'] dataset:
Evaluating NanoBeIRNanoQuoraRetrieval
Evaluating NanoBeIRNanoMSMARCO

Average Queries: 50.0
Average Corpus: 5044.5

Aggregated for Score Function: cosine
Accuracy@1: 39.00%
Accuracy@3: 57.00%
Accuracy@5: 66.00%
Accuracy@10: 77.00%
Precision@1: 39.00%
Recall@1: 34.03%
Precision@3: 20.67%
Recall@3: 54.07%
Precision@5: 15.00%
Recall@5: 64.27%
Precision@10: 8.90%
Recall@10: 75.97%
MRR@10: 0.5004
NDCG@10: 0.5513
Aggregated for Score Function: dot
Accuracy@1: 39.00%
Accuracy@3: 57.00%
Accuracy@5: 66.00%
Accuracy@10: 77.00%
Precision@1: 39.00%
Recall@1: 34.03%
Precision@3: 20.67%
Recall@3: 54.07%
Precision@5: 15.00%
Recall@5: 64.27%
Precision@10: 8.90%
Recall@10: 75.97%
MRR@10: 0.5004
[...]

PR #3886, opened 2026-07-28
Fix inverted Euclidean/Manhattan metrics in BinaryClassificationEvaluator

Kropiunig:
Hello!

## Pull Request overview
* Fix inverted Euclidean/Manhattan distance direction in `BinaryClassificationEvaluator`
* Add a regression test for the distance metric direction

## Details

`BinaryClassificationEvaluator.compute_metrics` passes the raw output of `pairwise_manhattan_sim` / `pairwise_euclidean_sim` into metric computations flagged with `"greater_is_better": False` (`sentence_transformers/sentence_transformer/evaluation/binary_classification.py:250-259`). Those functions return *negative* distances (i.e. similarities), while `greater_is_better=False` expects positive distances. This is the same sim/distance mismatch as #3702, which #3704 fixed for `TripletLoss`; the evaluator has the analogous problem. It was introduced in v3.3.0, when the score functions changed from `paired_manhattan_distances(x, y)` to `-paired_manhattan_distances(x, y)` without flipping `greater_is_better`.

As a result, every metric reported for the `"euclidean"` and `"manhattan"` similarity functions is computed for the opposite classifier ("larger distance means similar"):

* `find_best_acc_and_threshold` / `find_best_f1_and_threshold` sort ascending and sweep the most *distant* pairs first, so accuracy, F1, precision, recall and both thresholds are wrong,
* `average_precision_score(labels, scores * -1)` negates the scores a second time, ranking the most distant pairs as most similar,
* the MCC predictions `scores <= f1_threshold` label the most distant pairs as positives,
* the reported thresholds are negative and unusable as distance cutoffs.

Failing example on `main`, with 4 pairs whose similar pairs (distances 1 and 2) are perfectly separable from the dissimilar pairs (distances 10 and 20):

```python
import numpy as np
from sentence_transformers.sentence_transformer.evaluation import BinaryClassificationEvaluator

embeddings = {
    "a1": [0.0, 0.0],  "a2": [1.0, 0.0],   # distance 1, similar (label 1)
    "b1": [0.0, 10.0], "b2": [0.0, 12.0],  # distance 2, similar (label 1)
    "c1": [0.0, 0.0],  "c2": [10.0, 0.0],  # distance 10, dissimilar (label 0)
    "d1": [5.0, 5.0],  "d2": [5.0, 25.0],  # distance 20, dissimilar (label 0)
}

class MockEncoder:
    def encode(self, sentences, **kwargs):
        return np.array([embeddings[sentence] for sentence in sentences], dtype=np.float32)

evaluator = BinaryClassificationEvaluator(
    sentences1=["a1", "b1", "c1", "d1"],
    sentences2=["a2", "b2", "c2", "d2"],
    labels=[1, 1, 0, 0],
    similarity_fn_names=["euclidean"],
)
print(evaluator.compute_metrics(MockEncoder())["euclidean"])
```

Observed on `main` (every metric should be 1.0 for this perfectly separable data):

```
accuracy: 0.25, accuracy_threshold: -15.0, f1: 0.4, f1_threshold: -1.0,
precision: 0.333, recall: 0.5, ap: 0.5, mcc: 0.0
```

The symptom is also visible on real data in the docstring example output of `SparseBinaryClassificationEvaluator` (which inherits `compute_metrics`): `Matthews Correlation with Euclidean-Distance: -4.59`, recall 99.69 at precision 32.13, and thresholds of `-10.0041` / `-103.0263`.

## The fix

Negate the two `pairwise_*_sim` outputs back into positive distances, mirroring the #3704 fix for `TripletDistanceMetric`. This restores the pre-v3.3.0 behavior: metrics are computed on true distances and the reported `accuracy_threshold` / `f1_threshold` are positive distance cutoffs again. With the fix, the example above yields 1.0 for all six metrics and thresholds of 6.0 (between distance 2 and distance 10).

`EmbeddingSimilarityEvaluator` and `TripletEvaluator` use the `pairwise_*_sim` outputs sign-consistently (correlation with gold scores, and sim(a, p) > sim(a, n) comparison respectively), so only this evaluator was affected.

## Tests

Added `test_BinaryClassificationEvaluator_distance_metrics_direction`, parametrized over `euclidean` and `manhattan`. It fails on `main` (accuracy 0.25 instead of 1.0) and passes with this fix; the rest of the test file passes as well.

### DIFF
--- a/examples/sparse_encoder/evaluation/sparse_classification_evaluator.py
+++ b/examples/sparse_encoder/evaluation/sparse_classification_evaluator.py
@@ -33,27 +33,27 @@ Average Precision with Cosine-Similarity:    67.81
 Matthews Correlation with Cosine-Similarity: 49.56
 
 Accuracy with Dot-Product:             76.50    (Threshold: 23.4236)
-F1 with Dot-Product:                   67.00    (Threshold: 19.0095)
+F1 with Dot-Product:                   67.00    (Threshold: 19.0094)
 Precision with Dot-Product:            55.93
 Recall with Dot-Product:               83.54
 Average Precision with Dot-Product:    65.89
 Matthews Correlation with Dot-Product: 48.88
 
-Accuracy with Euclidean-Distance:             67.70     (Threshold: -10.0041)
-F1 with Euclidean-Distance:                   48.60     (Threshold: -0.1876)
-Precision with Euclidean-Distance:            32.13
-Recall with Euclidean-Distance:               99.69
-Average Precision with Euclidean-Distance:    20.52
-Matthews Correlation with Euclidean-Distance: -4.59
+Accuracy with Euclidean-Distance:             74.40     (Threshold: 2.5143)
+F1 with Euclidean-Distance:                   63.38     (Threshold: 5.2380)
+Precision with Euclidean-Distance:            48.98
+Recall with Euclidean-Distance:               89.75
+Average Precision with Euclidean-Distance:    65.34
+Matthews Correlation with Euclidean-Distance: 43.09
 
-Accuracy with Manhattan-Distance:             67.70     (Threshold: -103.0263)
-F1 with Manhattan-Distance:                   48.60     (Threshold: -0.8532)
-Precision with Manhattan-Distance:            32.13
-Recall with Manhattan-Distance:               99.69
-Average Precision with Manhattan-Distance:    21.05
-Matthews Correlation with Manhattan-Distance: -4.59
+Accuracy with Manhattan-Distance:             72.40     (Threshold: 11.8888)
+F1 with Manhattan-Distance:                   62.62     (Threshold: 31.5751)
+Precision with Manhattan-Distance:            48.09
+Recall with Manhattan-Distance:               89.75
+Average Precision with Manhattan-Distance:    60.15
+Matthews Correlation with Manhattan-Distance: 41.73
 
-Model Sparsity: Active Dimensions: 61.2, Sparsity Ratio: 0.9980
+Model Sparsity: Active Dimensions: 61.3, Sparsity Ratio: 0.9980
 """
 # Print the results
 print(f"Primary metric: {binary_acc_evaluator.primary_metric}")
--- a/sentence_transformers/sentence_transformer/evaluation/binary_classification.py
+++ b/sentence_transformers/sentence_transformer/evaluation/binary_classification.py
@@ -247,13 +247,15 @@ class BinaryClassificationEvaluator(BaseEvaluator):
                 "name": "Dot-Product",
                 "greater_is_better": True,
             },
+            # The pairwise_*_sim functions return negative distances (i.e. similarities), so negate
+            # them back into positive distances to match "greater_is_better": False (#3702).
             SimilarityFunction.MANHATTAN.value: {
-                "score_fn": lambda x, y: pairwise_manhattan_sim(x, y),
+                "score_fn": lambda x, y: -pairwise_manhattan_sim(x, y),
                 "name": "Manhattan-Distance",
                 "greater_is_better": False,
             },
             SimilarityFunction.EUCLIDEAN.value: {
-                "score_fn": lambda x, y: pairwise_euclidean_sim(x, y),
+                "score_fn": lambda x, y: -pairwise_euclidean_sim(x, y),
                 "name": "Euclidean-Distance",
                 "greater_is_better": False,
             },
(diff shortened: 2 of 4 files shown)


ErenAta16 (commented):
Reproduced this and it holds up. Ran your snippet against `main` and against the branch:

```
main       euclidean  {accuracy: 0.25, accuracy_threshold: -15.0, f1: 0.4, f1_threshold: -1.5,
                       precision: 0.3333, recall: 0.5, ap: 0.4167, mcc: -0.5774}
main       manhattan  same
branch     euclidean  {accuracy: 1.0, accuracy_threshold: 6.0, f1: 1.0, f1_threshold: 6.0,
                       precision: 1.0, recall: 1.0, ap: 1.0, mcc: 1.0}
branch     manhattan  same
main/branch cosine    unchanged  {accuracy: 0.75, accuracy_threshold: 0.916, ...}
main/branch dot       unchanged
```

Perfectly separable data producing `accuracy: 0.25` and a negative threshold on `main` is about as clear as it gets, and every value your test asserts (all metrics `1.0`, both thresholds `6.0`) comes out exactly on the branch.

Minor, and it does not change the conclusion: the "observed on `main`" block in your description has `f1_threshold: -1.0, ap: 0.5, mcc: 0.0`, where I get `-1.5`, `0.4167` and `-0.5774` on current `main` (`d407492`). Probably an older checkout. Worth refreshing so a reviewer running it doesn't stop at the mismatch.

## The other two evaluators using the same convention are fine

`pairwise_euclidean_sim` and `pairwise_manhattan_sim` are also called by `EmbeddingSimilarityEvaluator` and `TripletEvaluator`. I checked both, since "negate everywhere for symmetry" is the obvious next thought and would be wrong:

```
EmbeddingSimilarityEvaluator   (gold: closer pair scores higher)
   spearman_euclidean   1.0000     pearson_euclidean   0.9343
   spearman_manhattan   1.0000     pearson_manhattan   0.9343

TripletEvaluator               (positive is closer than negative)
   euclidean_accuracy   1.0000     manhattan_accuracy  1.0000
```

Both consume the negated distance monotonically, so the sign works out. `greater_is_better: False` in `binary_classification.py` is the only place that assumes positive distances, which is why it is the only one that inverts. Might be worth a line in the description, it makes the scope of the change easy to agree with.

## This is not only a reporting bug

`primary_metric` is set from the first entry of `similarity_fn_names`:

```python
self.primary_metric = f"{self.similarity_fn_names[0]}_ap"
```

so `similarity_fn_names=["euclidean"]` makes `euclidean_ap` the primary metric, and that feeds `metric_for_best_model` / `load_best_model_at_end` during training. On `main` that number ranks the most distant pairs as the most similar, so checkpoint selection was being driven by an inverted AP, not just the printed summary. That seems worth stating in the description, it raises this from "numbers in the log look odd" to "the wrong checkpoint may have been kept".

## Stale docstring

The `SparseBinaryClassificationEvaluator` docstring at [`sparse_binary_classification.py#L88-L100`](https://github.com/huggingface/sentence-transformers/blob/main/sentence_transformers/sparse_encoder/evaluation/sparse_binary_classification.py#L88-L100) is the sample output you cite as evidence, and it stays in the tree after this change:

```
Accuracy with Euclidean-Distance:             67.70     (Threshold: -10.0041)
F1 with Euclidean-Distance:                   48.60     (Threshold: -0.1876)
Matthews Correlation with Euclidean-Distance: -4.59
```

Those thresholds are negative for exactly the reason this PR fixes, so after it lands the published docs advertise output the code can no longer produce, and a negative MCC next to 99.69 recall reads as a property of the metric rather than a bug. Rerunning that example is a model download, so if that is more than you want to take on here, replacing the numbers with a note or opening a follow-up both seem fine. The dense `BinaryClassificationEvaluator` docstring only shows cosine, so it is unaffected.

## Test

Good shape. Handpicked distances (1, 2 similar / 10, 20 dissimilar) make the expected threshold `(2 + 10) / 2 = 6` something a reader can check by hand instead of a value copied from a run, and parametrizing over both metrics catches the case where only one lambda gets fixed. Confirmed it fails on `main` for both parametrizations and passes on the branch.

tomaarsen:
Hello!

Thanks for this PR, and well spotted! I can confirm the diagnosis: this is the evaluator counterpart of #3702, which #3704 fixed for `TripletLoss`. The evaluator variant is even older, it dates back to v3.3.0 (#2966), when the score functions were negated without flipping `greater_is_better`. Thanks @ErenAta16 for the independent verification as well, much appreciated!

I verified the fix and test on my end (the new test indeed fails on `main` with accuracy 0.25 and passes here) and pushed a small commit on top:
* Reran the `SparseBinaryClassificationEvaluator` example with the fix and refreshed the stale sample outputs in its docstring and in `examples/sparse_encoder/evaluation/sparse_classification_evaluator.py`. The Euclidean AP jumps from 20.52 to 65.34, and the thresholds are positive distances again.
* Trimmed the new code comment down to one line.

For completeness: `SparseBinaryClassificationEvaluator` inherits `compute_metrics`, so it is fixed by this PR too, and `primary_metric` needs no changes (AP is higher-is-better regardless of the score direction, only its value was inverted).

- Tom Aarsen

PR #3887, opened 2026-07-28
Drop FAISS padding placeholders from `semantic_search_faiss` results

ErenAta16:
`semantic_search_faiss` returns `corpus_id: -1` entries whenever the corpus has fewer vectors than the requested `k`.

FAISS pads its `indices` array with `-1` for the slots it couldn't fill. The function passes those straight into the output, so a search over a 3-document corpus with `top_k=5` comes back with five results instead of three:

```python
import numpy as np
from sentence_transformers.util import quantize_embeddings, semantic_search_faiss

rng = np.random.default_rng(0)
corpus = quantize_embeddings(rng.random((3, 16), dtype=np.float32), precision="ubinary")
queries = rng.random((2, 16), dtype=np.float32)

results, _ = semantic_search_faiss(queries, corpus_embeddings=corpus,
                                   corpus_precision="ubinary", top_k=5, rescore=True)
print([[hit["corpus_id"] for hit in r] for r in results])
```

```
main:  [[0, 1, 2, -1, -1], [0, 1, 2, -1, -1]]
here:  [[0, 1, 2], [0, 1, 2]]
```

Two problems with letting `-1` through:

1. It isn't a sentinel the caller can trip over. The documented pattern is `corpus[hit["corpus_id"]]`, and `corpus[-1]` silently returns the **last** document. Nothing raises, so a small corpus quietly produces duplicate results with a plausible-looking score attached (`2147483647.0` for binary indices, i.e. `INT_MAX` sitting in the padding slot).
2. On the rescore path, `corpus_index.reconstruct(-1)` reads out of bounds rather than raising. The placeholder gets a garbage vector, that vector goes through the same `einsum` as the real candidates, and it can outrank an actual hit in the `argsort`. With `corpus_precision="uint8"` the garbage also trips `RuntimeWarning: invalid value encountered in cast` on the `astype(int)` a few lines below. On an empty index `reconstruct()` segfaults the process.

Reproduces on every combination I tried, `ubinary` and `uint8`, `exact=True` and `exact=False`, `rescore=True` and `rescore=False`:

```
ubinary  exact=True   rescore=True   -> [[0, 1, 2, -1, -1], [0, 1, 2, -1, -1]]
ubinary  exact=True   rescore=False  -> [[0, 1, 2, -1, -1], [0, 1, 2, -1, -1]]
ubinary  exact=False  rescore=True   -> [[0, 1, 2, -1, -1], [0, 1, 2, -1, -1]]
ubinary  exact=False  rescore=False  -> [[0, 1, 2, -1, -1], [0, 1, 2, -1, -1]]
uint8    exact=True   rescore=True   -> [[0, 1, 2, -1, -1], [0, 1, 2, -1, -1]]
uint8    exact=True   rescore=False  -> [[0, 1, 2, -1, -1], [0, 1, 2, -1, -1]]
uint8    exact=False  rescore=True   -> [[0, 1, 2, -1, -1], [0, 1, 2, -1, -1]]
uint8    exact=False  rescore=False  -> [[0, 1, 2, -1, -1], [2, 0, 1, -1, -1]]
```

All eight are clean after the change.

## What the patch does

Record `padded = indices == -1` right after `corpus_index.search`, then:

- feed padded slots a zero placeholder instead of calling `reconstruct` on them,
- set their rescored score to `-inf` so they can't win the `argsort`,
- recompute the mask after the reorder and skip them when building the output dicts.

The zero placeholder is built from `corpus_index.d` rather than `reconstruct(0)`, so it also works when the index is empty. That case previously segfaulted with `rescore=True` and returned five phantom hits with `rescore=False`; it now returns `[]`.

Nothing changes for a corpus larger than `top_k` — `padded` is all-`False` and every added line is a no-op.

`semantic_search_usearch` is untouched. usearch returns a ragged `Matches` object rather than padding out to `k`, so it doesn't have the same problem.

## Tests

Added to `tests/util/test_quantization.py`, behind the same `find_spec("faiss")` skip guard used in `test_hard_negatives.py`:

- `test_semantic_search_faiss_drops_padded_indices` — parametrized over `["ubinary", "uint8"]` x `[rescore True, False]`, asserts the 3-document corpus returns exactly `[0, 1, 2]`.
- `test_semantic_search_faiss_empty_corpus` — asserts `[[], []]`.
- `test_semantic_search_faiss_returns_top_k_when_corpus_is_large_enough` — the no-regression guard: a 50-document corpus still returns 5 hits per query.

Reverting only the `quantization.py` hunk (`git stash push -- sentence_transformers/util/quantization.py`) fails the four `drops_padded` cases and segfaults on the empty-corpus one. The `top_k` guard stays green either way, which is what you'd want from it.

---

Unrelated so I left it alone, but flagging it while I'm in this file: the `semantic_search_faiss` docstring lists the `corpus_precision` options as `"float32", "int8", "binary"`, while the signature says `Literal["float32", "uint8", "ubinary"]`. Looks like it was copied over from `semantic_search_usearch`, which does use `int8`/`binary`. Happy to fold a one-line docstring fix into this PR or send it separately, whichever you prefer.

### DIFF
--- a/sentence_transformers/util/quantization.py
+++ b/sentence_transformers/util/quantization.py
@@ -144,10 +144,25 @@ def semantic_search_faiss(
     start_t = time.time()
     scores, indices = corpus_index.search(query_embeddings, k)
 
+    # FAISS pads `indices` with -1 when the index holds fewer vectors than requested. These are
+    # not corpus ids: `reconstruct(-1)` returns a garbage vector instead of raising, and a caller
+    # indexing their corpus with -1 silently gets the *last* document. Track them so they are
+    # neither rescored nor returned.
+    padded = indices == -1
+
     # If rescoring is enabled, we need to rescore the results using the rescore_embeddings
     if rescore_embeddings is not None:
+        # Padded slots get a zero placeholder rather than a `reconstruct` call: `reconstruct(-1)`
+        # reads out of bounds, and on an empty index it segfaults outright.
+        if corpus_precision == "ubinary":
+            zero_embedding = np.zeros(corpus_index.d // 8, dtype=np.uint8)
+        else:
+            zero_embedding = np.zeros(corpus_index.d, dtype=np.float32)
         top_k_embeddings = np.array(
-            [[corpus_index.reconstruct(idx.item()) for idx in query_indices] for query_indices in indices]
+            [
+                [corpus_index.reconstruct(idx.item()) if idx != -1 else zero_embedding for idx in query_indices]
+                for query_indices in indices
+            ]
         )
         # If the corpus precision is binary, we need to unpack the bits
         if corpus_precision == "ubinary":
@@ -161,9 +176,12 @@ def semantic_search_faiss(
         # We use einsum to calculate the dot product between the query and the top_k embeddings, equivalent to looping
         # over the queries and calculating 'rescore_embeddings[i] @ top_k_embeddings[i].T'
         rescored_scores = np.einsum("ij,ikj->ik", rescore_embeddings, top_k_embeddings)
+        # A padded slot must never outrank a real hit, whatever its garbage vector scored.
+        rescored_scores = np.where(padded, -np.inf, rescored_scores)
         rescored_indices = np.argsort(-rescored_scores)[:, :top_k]
         indices = indices[np.arange(len(query_embeddings))[:, None], rescored_indices]
         scores = rescored_scores[np.arange(len(query_embeddings))[:, None], rescored_indices]
+        padded = indices == -1
 
     delta_t = time.time() - start_t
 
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Thanks for this PR, this was a good catch, and apologies that it initially went without a review. I noticed you had closed it, but the bug was real and the fix was well made, so I reopened it, pushed one commit on top, and merged it.

The extra commit simplifies the padding handling to match the equivalent fix in `mine_hard_negatives` from #3872 (an inline `-inf` disqualification and a `!= -1` filter on the output, rather than a separate mask), and switches the new tests to `standard_normal` data: uniform data always quantizes to the all-ones pattern with binary precision, so the test documents were indistinguishable.

Thanks again for the clear report, the reproductions, and the tests. They made this easy to verify and merge.

- Tom Aarsen

PR #3888, opened 2026-07-28
Correct the `corpus_precision` values documented for `semantic_search_faiss`

ErenAta16:
`semantic_search_faiss` documents `corpus_precision` values it does not accept.

```
docstring:  "float32", "int8", or "binary"
signature:  Literal["float32", "uint8", "ubinary"]
code:       if corpus_precision in ("float32", "uint8") ... elif corpus_precision == "ubinary"
```

Following the docstring gets you this:

```python
>>> semantic_search_faiss(queries, corpus_embeddings=corpus, corpus_precision="int8", top_k=3)
AttributeError: 'NoneType' object has no attribute 'add'
```

Neither branch matches, `corpus_index` stays `None`, and the failure lands on `corpus_index.add(...)` a few lines later without naming the argument that caused it. Same for `"binary"`, and for any typo.

The wording looks copied over from `semantic_search_usearch`, which does take `"int8"` and `"binary"`, so this is a rename that did not reach the docs rather than a design decision.

## Changes

**Document the values the function actually takes.** `"float32"`, `"uint8"`, `"ubinary"`, in both the `corpus_precision` entry and the `corpus_embeddings` one that repeats them.

**Validate the argument.** `semantic_search_usearch` already does this a few lines into its body:

```python
if corpus_precision not in ["float32", "int8", "ubinary", "binary"]:
    raise ValueError('corpus_precision must be "float32", "int8", "ubinary", "binary" for usearch')
```

The faiss twin has no equivalent, which is why a bad value travels as far as it does. Added the same check with the same message shape, so the two functions fail the same way.

**Add `"ubinary"` to the usearch `Literal`.** That one is the reverse problem: the runtime check accepts `["float32", "int8", "ubinary", "binary"]`, there is a full `ubinary` branch building a `b1` Hamming index, and the docstring lists it. Only the type hint left it out, so a type-checker rejects a call that works.

## Tests

`test_semantic_search_faiss_rejects_unsupported_precision`, parametrized over `"int8"`, `"binary"` and `"float16"`, asserting a `ValueError` that names `corpus_precision`. All three fail on `main` with the `AttributeError` above.

```
$ pytest tests/util/test_quantization.py
19 passed
```

## Note on overlap

This touches `tests/util/test_quantization.py`, same as #3887. The additions sit in different parts of the file (this one after `test_binary_quantize_row_independence`, that one at the end), so whichever lands second should apply cleanly, but flagging it in case. The two are independent otherwise, no shared source hunks.

I raised the docstring mismatch at the bottom of #3887 and offered to fold it in there. Sending it separately instead so #3887 stays a single behaviour change and this stays a docs/validation change. Happy to squash them if you would rather have one.


tomaarsen:
Hello!

Thanks for this one as well. Like #3887 I reopened it: the docstrings really did document values that `semantic_search_faiss` does not accept, and the new validation error beats the `AttributeError` it replaces.

I pushed a few commits on top: a merge with `main` to pick up #3887 (the test files overlapped), a move of the new test in with the other `semantic_search_faiss` tests so it can reuse the shared fixtures, and some further polish of the same docstrings (a few more leftover `int8` mentions and similar copy-paste artifacts). I'll merge this once CI agrees.

Thanks for both PRs!

- Tom Aarsen

PR #3890, opened 2026-07-30
[`v6.1`] Resolve Hub revisions before model loading

Wauplin:
> **Note (2026-08-31, @tomaarsen):** finalized for v6.1. See the update at the end of this description for what changed since the original scope and for the loading benchmarks.

## Summary

- resolve a Hub model revision once before loading SentenceTransformer, SparseEncoder, or CrossEncoder components
- do the same for direct `Transformer` pipelines, resolving a separate processor repository independently when needed
- use the resolved commit for ONNX/OpenVINO repository listings and model-card metadata, avoiding a duplicate model-info lookup
- preserve the existing `huggingface-hub>=0.23.0` compatibility floor through a conditional import; older Hub versions keep the legacy loading behavior
- memoize repeated resolution only in the shared test cache to keep unauthenticated CI below Hub API limits, while preserving normal behavior for isolated cache tests

## Motivation

[`huggingface_hub` v1.26.0](https://github.com/huggingface/huggingface_hub/releases/tag/v1.26.0) introduced `resolve_revision` in [huggingface_hub#4604](https://github.com/huggingface/huggingface_hub/pull/4604). Passing its `ResolvedRevision` through the loading pipeline pins all component downloads to one commit, preventing mixed-revision loads while avoiding repeated revision-resolution requests.

## Validation

- Ruff lint and formatting checks
- focused compatibility tests with legacy `huggingface_hub` and with the v1.26 implementation
- affected utility, base model, Transformer, and model-card suites
- complete GitHub Actions matrix: quality, minimum dependency versions, and all 16 Python/OS/Transformers unit-test jobs

---

## Update (2026-08-31, @tomaarsen)

The description above is the original one. Since then I merged `main` and pushed three commits that narrow and harden the scope a bit:

* `Transformer` no longer resolves revisions itself. `BaseModel` already resolves once and passes the result to every module, so the only piece kept in `Transformer.__init__` is a guard that keeps a separate `tokenizer_name_or_path` repository from being pinned to the model's commit.
* Resolution failures degrade gracefully. A missing revision still raises, an inaccessible repository (private or gated without a valid token) is resolved from the local cache like `hf_hub_download` does, and anything else (for example a 429) logs a warning and loads without pinning, which is the previous behavior.
* `list_repo_files` is not `ResolvedRevision`-aware (unlike `hf_hub_download` and `snapshot_download`), so the ONNX/OpenVINO listing passes `.resolved` explicitly. This was the one red test across the CI matrix.

Loading benchmarks (fresh process per load, constructor time only, huggingface_hub 1.29.0 in both arms, the "before" arm with the resolution disabled, medians over 7 runs). Warm cache, online:

| Model | Before | After | Delta |
|---|---|---|---|
| sentence-transformers-testing/stsb-bert-tiny-safetensors | 5.78s, 34 requests | 2.26s, 8 requests | -61% |
| sentence-transformers/all-MiniLM-L6-v2 | 6.39s, 35 requests | 2.66s, 9 requests | -58% |
| sparse-encoder-testing/splade-bert-tiny-nq (SparseEncoder) | 5.77s, 34 requests | 2.25s, 8 requests | -61% |
| cross-encoder/ms-marco-MiniLM-L6-v2 (CrossEncoder) | 4.38s, 23 requests | 2.48s, 9 requests | -43% |

With `HF_HUB_OFFLINE=1` both arms are identical, and even a cold cache load of the tiny model is 21% faster (54 to 42 requests). Details in the comments below.


tomaarsen:
I saw you mention this earlier today, and I already wondered if this would allow me to make some improvements!
I'll review this soon, I'm currently wrapping up some final reviews on the v6.0 prep, which has resulted in me slacking on some recent PR reviews. I'll catch up with all PRs soon, and incorporate them in a v5.7 prior to the big v6.0.

- Tom Aarsen

sentence_transformers/backend/utils.py
```diff
@@ -92,10 +92,11 @@ def backend_should_export(
     if is_local:
         model_file_names = [path.relative_to(load_path).as_posix() for path in load_path.glob(glob_pattern)]
     else:
+        revision = model_kwargs.get("revision")
         all_files = list_repo_files(
             load_path.as_posix(),
             repo_type="model",
-            revision=model_kwargs.get("revision", None),
+            revision=getattr(revision, "resolved", revision),
```
Wauplin, line 100:
```suggestion
            revision=model_kwargs.get("revision", None),
```

sentence_transformers/backend/utils.py
```diff
@@ -92,10 +92,11 @@ def backend_should_export(
     if is_local:
         model_file_names = [path.relative_to(load_path).as_posix() for path in load_path.glob(glob_pattern)]
     else:
+        revision = model_kwargs.get("revision")
```
Wauplin:
```suggestion
```

Wauplin:
thanks for checking in @tomaarsen ! This PR is indeed related to this [internal slack thread](https://huggingface.slack.com/archives/C02V5EA0A95/p1785317821409099). I won't have time to review/test it myself before getting in vacay but I can take care of it when I'm back. Ofc don't hesitate to test and merge yourself if you have some bandwidth (especially if prepping for a major release soon :hugs:). The core idea of the PR is pretty simple, it's just a matter of resolving the revision only once to avoid redundant HTTP calls downstream.

tomaarsen:
I'm going to postpone this for v6.1 as I want to make sure I don't make a mistake with it + my bandwidth is a bit low. Renaming the PR to remind myself.

- Tom Aarsen

tomaarsen:
I picked this up again for v6.1 and pushed a few (AI-assisted, human-reviewed) commits: merged `main`, restored the `.resolved` unwrap in `backend_should_export` that the review suggestions had reverted (`list_repo_files` is not `ResolvedRevision`-aware, unlike `hf_hub_download`, which is why the whole matrix was red), reduced the `Transformer` changes to a guard that keeps a separate `tokenizer_name_or_path` repository from being pinned to the model's commit (`BaseModel` already resolves once and hands the result to every module), and made `_resolve_model_revision` fall back to the requested revision with a warning when resolution fails for anything other than a missing repository or revision, since `resolve_revision` raises on 429 where `hf_hub_download` would have served the cache (a cache fallback for 429 in hub itself might be worth considering).

The main reason to get this in is loading speed. Each load in a fresh process, constructor time only, huggingface_hub 1.29.0 in both arms, with the "before" arm disabling the resolution (`_hub_resolve_revision = None`), which reproduces the previous behavior exactly. Medians over 7 runs, arms alternated, HTTP requests counted at the httpx transport.

Warm cache, online:

| Model | Before | After | Delta |
|---|---|---|---|
| sentence-transformers-testing/stsb-bert-tiny-safetensors | 5.78s, 34 requests | 2.26s, 8 requests | -61% |
| sentence-transformers/all-MiniLM-L6-v2 | 6.39s, 35 requests | 2.66s, 9 requests | -58% |
| sparse-encoder-testing/splade-bert-tiny-nq (SparseEncoder) | 5.77s, 34 requests | 2.25s, 8 requests | -61% |
| cross-encoder/ms-marco-MiniLM-L6-v2 (CrossEncoder) | 4.38s, 23 requests | 2.48s, 9 requests | -43% |

With `HF_HUB_OFFLINE=1` both arms are identical (about 0.25s, no requests), and with a cold cache the tiny model still loads 21% faster (9.18s and 54 requests before, 7.29s and 42 requests after). Previously every cached file cost a HEAD on `resolve/main/<file>` plus its redirect, several files were probed more than once, and `model_info` was called for the model card on top. Downloads pinned to a commit hash skip all of that, what remains are the optional file probes and `tree/main` listings that transformers does.

I'll mark this ready for review once CI is green. Thanks for the PR, this turned out to be a bigger win than I expected!

- Tom Aarsen

Wauplin:
Glad to hear that!

tomaarsen:
Two more (AI-assisted, human-reviewed) commits, both about the CI flakiness that showed up once loading got faster.

The failures were all `test_safetensors` erroring at fixture setup with a `FileNotFoundError` on `whitespacetokenizer_config.json`. The cause is that `WordEmbeddings.load` calls `load_config` before `load_dir_path`, and that first call already creates `snapshots/<sha>/0_WordEmbeddings/` holding just `wordembedding_config.json`. When the later `snapshot_download` gets rate limited, hub hands back that existing partial snapshot without raising, and the old guard in `load_dir_path` only checked that the subfolder directory exists, which it now does. So a single 429 turned into a confusing missing-file error deep in tokenizer loading rather than a download error.

`load_dir_path` now rejects a cached snapshot that does not contain the requested subfolder (both on the direct return and on the cache-only retry) and lets `IncompleteSnapshotError` propagate instead of converting it into a `None` module path. That one is worth a release note, since it changes what callers can see: `load_dir_path` can now raise `OSError` or `IncompleteSnapshotError` where it previously returned a path or `None`. It is a pre-existing issue rather than something this PR introduced, it just became much easier to hit.

Worth being precise about what that does and does not fix. It covers the case where the subfolder is missing entirely. It does not cover a subfolder that is present but incomplete, which is the case above, because hub only raises `IncompleteSnapshotError` when a cached tree listing proves files are missing, and a cold CI cache has no such listing. Neither hub nor `load_dir_path` can know which files were expected there, so I left it at that.

The thing that hopefully makes CI reliable is the second commit. The `avg_word_embeddings_levy` fixture pulled `sentence-transformers/average_word_embeddings_levy_dependency`, which is 401 MB on disk (208 MB of `model.safetensors` plus 208 MB of `pytorch_model.bin` that we never read, since `load_dir_path` passes `allow_patterns="<subfolder>/**"` and takes the whole subfolder). That is a lot of Hub traffic for a test that only needs a non-transformer module for a safetensors round trip, and it is what kept running into the anonymous rate limit on fork CI. The fixture now builds a small `WordEmbeddings` plus `Pooling` locally, so it makes no requests at all. Measured on a warm cache the old fixture was 3.35s and 17 requests before the download even mattered.

No coverage is lost that I care about. Nothing asserted on those pretrained vectors, and loading an old style multi subfolder repository from the Hub is still covered by `average_word_embeddings_komninos` in `tests/sentence_transformer/test_pretrained_stsb.py`, plus the mocked `load_dir_path` tests in `tests/util/test_file_io.py`.

- Tom Aarsen

PR #3794, opened 2026-06-03
[`v6`] Add support for MultiVectorEncoder models

tomaarsen:
Hello!

## Pull Request overview
* Introduce `MultiVectorEncoder`, a first-class model family for ColBERT-style / late-interaction retrieval
* Ship the full training, evaluation, and scoring surface alongside the model
* Add `SimilarityFunction.MAXSIM` and `maxsim` / `maxsim_pairwise` utilities

## Details
This brings PyLate's feature set into `sentence-transformers` as a first-class `MultiVectorEncoder` family sitting alongside `SentenceTransformer`, `SparseEncoder`, and `CrossEncoder`. Models produce a sequence of token-level vectors per input, while scoring uses MaxSim late-interaction. The naming mirrors `SparseEncoder`'s output-shape framing ("late interaction" is one scoring strategy on top of multi-vector outputs, not the encoder itself).

The package at `sentence_transformers/multi_vector_encoder/` ships with `MultiVectorTransformer` (`query_length` / `document_length`, PAD -> MASK query expansion, `attend_to_expansion_tokens`), `MultiVectorMask` (skiplist), `HierarchicalPooling` (Ward clustering for storage compression), four losses (`MultiVectorMultipleNegativesRankingLoss`, `CachedMultiVectorMultipleNegativesRankingLoss`, `MultiVectorDistillKLDivLoss`, `MultiVectorMarginMSELoss`), and five evaluators (`MultiVector{InformationRetrieval,NanoBEIR,Triplet,Distillation,Reranking}Evaluator`). ColBERT MaxSim scoring lives in `util/similarity.py` so `model.similarity` dispatch and evaluators can request `"MaxSim"` by name, while XTR scoring lives in `multi_vector_encoder/scoring/xtr.py` and slots into any of the four losses as `score_metric=XTRScores()`. There's also `KDProcessing` for join-at-iteration-time KD data, re-exported from `sentence_transformers.util` since dense distillation flows benefit too.

[...]

PR #3892, opened 2026-07-31
[`v6`] fix: compute similarity scores in float32 to avoid low-precision ties (HPS)

### LINKED ISSUE #3894: Similarity and cross-encoder scores computed below float32 create spurious ties in low-precision inference
### Summary

When a model runs in **float16/bfloat16** (e.g. `model_kwargs={"torch_dtype": torch.float16}`), sentence-transformers computes the final relevance score in that same low precision in two paths. Reduced-mantissa formats coarsely bucket values in the narrow score range, so **distinct query–document scores collapse into identical values ("spurious ties")**. Ranking then falls back to arbitrary tie-breaking (index order), which inflates the variance and bias of retrieval metrics and can even flip model rankings.

This is analyzed in [*Reliable Evaluation Protocol for Low-Precision Retrieval*](https://aclanthology.org/2026.acl-short.33) (Yang et al., ACL 2026). The remedy, **High-Precision Scoring (HPS), upcasts only the final scoring op to float32**, leaving the forward pass in low precision. It is value-preserving (every fp16/bf16 value is exactly representable in fp32), zero-parameter, and adds negligible cost.

MTEB recently merged the same two fixes: embeddings-benchmark/mteb#4933 (similarity primitives) and embeddings-benchmark/mteb#4934 (cross-encoder logits), from issue embeddings-benchmark/mteb#4932.

### Where it happens on current `main`

1. **Similarity primitives** — `sentence_transformers/util/tensor.py::_convert_to_tensor` already upcasts **sparse** tensors to float32, but leaves dense tensors at their incoming dtype. A fp16/bf16 model with `encode(..., convert_to_tensor=True)` returns low-precision embeddings, so `model.similarity()` / `util.cos_sim` / `dot_score` (and every evaluator built on them, e.g. `InformationRetrievalEvaluator`) scores in low precision. Note that `encode(..., convert_to_numpy=True)` upcasts **bfloat16** on conversion (numpy has no bf16) but keeps **float16** as fp16 numpy arrays, which are converted back to fp16 tensors when scoring.

2. **Cross-encoder scoring** — `CrossEncoder.predict` applies `activation_fn` (sigmoid) / softmax directly to the model's logits, i.e. in model dtype. The `convert_to_numpy` tail does call `.float()`, but only *after* the activation has already bucketized the (0, 1) probability range.

### Illustration

```python
import torch
q, d = torch.randn(1, 32), torch.randn(200, 32)
qn = torch.nn.functional.normalize
naive = qn(q.bfloat16(), dim=1) @ qn(d.bfloat16(), dim=1).T  # score in bf16
hps   = qn(q.bfloat16().float(), dim=1) @ qn(d.bfloat16().float(), dim=1).T  # HPS
print(torch.unique(naive).numel(), torch.unique(hps).numel())  # e.g. 184 vs 200

logits = torch.randn(200).bfloat16()  # cross-encoder logits
p_bf16 = torch.sigmoid(logits)
p_hps  = torch.sigmoid(logits.float())
print(torch.unique(p_bf16).numel(), torch.unique(p_hps).numel())  # e.g. ~150 vs 200
```

Fewer unique scores = spurious ties, broken arbitrarily during ranking.

### Proposed fix

Apply HPS at both sites, exactly as merged in MTEB:

- Upcast sub-float32 dense floating tensors to float32 in `_convert_to_tensor` — this mirrors the float32 upcast the same function already applies to sparse inputs.
- Upcast cross-encoder logits to float32 before the activation in `CrossEncoder.predict`.

Both are no-ops for float32 inputs, so full-precision results are unchanged. Under AMP/autocast training the affected ops are re-cast per autocast policy, so training behavior is effectively unchanged as well.

I've prepared two small, separate PRs and am opening them alongside this issue. Happy to adjust scope if maintainers prefer.

Disclosure: I'm the first author of the referenced paper; the same fixes were merged into MTEB this week (embeddings-benchmark/mteb#4933, embeddings-benchmark/mteb#4934).

KisuYang:
Fixes #3894 (similarity-primitives part).

The similarity primitives in sentence_transformers.util convert non-tensor inputs to tensors and upcast sparse tensors to float32, but leave dense tensors at their incoming dtype. When embeddings are produced in float16/bfloat16 (e.g. models loaded with torch_dtype=torch.float16 and encode(..., convert_to_tensor=True)), cosine/dot scoring is therefore performed in low precision. Reduced-mantissa formats coarsely bucket values in the narrow score range, collapsing distinct similarity scores into spurious ties; rankings and downstream retrieval metrics then depend on arbitrary tie-breaking, inflating their variance and bias.

Upcast sub-float32 floating tensors to float32 in _convert_to_tensor (High-Precision Scoring), mirroring the existing float32 upcast already applied to sparse inputs on the lines above. The upcast is value-preserving (every float16/bfloat16 value is exactly representable in float32) and a no-op for float32 inputs, so full-precision results are unchanged.

Reference
- embeddings-benchmark/mteb#4933
- [*Reliable Evaluation Protocol for Low-Precision Retrieval*](https://aclanthology.org/2026.acl-short.33) (Yang et al., ACL 2026)


tomaarsen:
Hello!

Thanks for this PR and its sibling #3893, and apologies for the delay while the big v6 branch landed. The diagnosis here is spot on as far as I can tell: matmul rounds its output to the input dtype, so float16/bfloat16 scoring coarsely buckets the narrow score range and collapses distinct scores into spurious ties. It also matches a decision I ended up making on the multi-vector side in #3794, where `maxsim` always returns float32 scores for exactly this reason (although it does compute its cosine similarity in bf16/fp16).

While reviewing I measured just how bad this is for cosine specifically, since one might hope that [-1, 1] is safe territory: bfloat16 can only represent 129 distinct values in all of [0.5, 1.0), which is where retrieval cosine scores tend to live. In a quick probe with 10k realistic cosine scores (mean 0.7, std 0.05), float32 keeps 9983 distinct values, float16 keeps 593, and bfloat16 keeps just 93. So a 10k-document ranking scored in bfloat16 cosine has roughly 93 score levels, i.e. massive ties throughout. Your `test_hps_collapses_spurious_ties` pins this nicely.

I made one reshape before merging: the upcast moved out of `_convert_to_tensor` into a new `_convert_to_float_tensor`, which the single-vector scoring functions use (directly in the pairwise functions, and via `_convert_to_batch_tensor` for `cos_sim`, `dot_score`, etc.). The reason is that `_convert_to_tensor` is also the conversion helper for the multi-vector (ColBERT-style) scoring stack, which deliberately keeps its matmul in the input dtype and accumulates in float32 instead. Upcasting there would have doubled both the inputs and the einsum intermediate that `document_chunk_elements` exists to bound. With the relocation, your fix applies to all single-vector similarity functions while the multi-vector path keeps its documented behavior. Your tests needed no changes and all pass as-is.

One trade worth stating for the record: users who deliberately store a large corpus in float16 to save memory and score it with one raw `cos_sim` call now pay a transient float32 copy of the inputs. The chunked utilities such as `semantic_search` bound that per chunk, and correct float32 rankings are the right default, so this is intended.

Thanks a bunch for the careful tests, they made this easy to verify and reshape.

- Tom Aarsen

PR #3893, opened 2026-07-31
[`v6`] fix: upcast cross-encoder logits to float32 before activation (HPS)

### LINKED ISSUE #3894: Similarity and cross-encoder scores computed below float32 create spurious ties in low-precision inference

KisuYang:
Fixes #3894 (cross-encoder part).

CrossEncoder.predict applies the activation function (sigmoid/softmax) directly to the logits produced by the model. When the model runs in float16/bfloat16, the activation is computed in that same low precision; the reduced mantissa coarsely buckets the (0, 1) probability range, so distinct query-document relevance scores collapse into spurious ties. Reranking then depends on arbitrary tie-breaking, which inflates the variance and bias of downstream retrieval metrics. The existing convert_to_numpy path already calls .float(), but only after the activation has bucketized the scores.

Upcast sub-float32 logits to float32 before the activation (High-Precision Scoring). The upcast is value-preserving (every float16/bfloat16 value is exactly representable in float32) and a no-op for float32 logits, so full-precision results are unchanged.

Reference
- embeddings-benchmark/mteb#4934
- [*Reliable Evaluation Protocol for Low-Precision Retrieval*](https://aclanthology.org/2026.acl-short.33) (Yang et al., ACL 2026)


tomaarsen:
Hello!

Thanks for this PR as well! Same story as #3892: the diagnosis is right, and the fix sits in exactly the right place, before the activation. Once sigmoid has run in bfloat16, the saturation damage is unrecoverable: the later `.float()` that the numpy path already did just re-encodes the already-bucketed values in a wider container.

To quantify it, I evaluated `cross-encoder/ettin-reranker-32m-v1` in bfloat16 on NanoBEIR (msmarco, nq, fiqa2018), reranking 100 candidates per query, with and without this fix:

- With the model's default activation, which for this particular model is `Identity` (raw logits), the metrics are identical with and without the fix: bfloat16 logits keep roughly 0.4 percent relative spacing, which is enough resolution for 100 candidates.
- With `activation_fn=torch.nn.Sigmoid()`, i.e. what a default single-label reranker uses, the difference is dramatic:

| | with fix | without fix |
|---|---|---|
| NanoBEIR mean NDCG@10 | **0.6795** | 0.1849 |
| NanoBEIR mean MRR@10 | **0.6797** | 0.3986 |
| unique scores over 15,040 pairs | 710 | 270 |

The mechanism is sigmoid saturation: a confident reranker pushes many of each query's candidates toward 1.0, where bfloat16 has a spacing of about 0.0039, so whole blocks of top candidates collapse onto the same value and the positive document lands at an arbitrary position inside the tie block. NanoMSMARCO NDCG@10 falls from 0.7093 to 0.0965, essentially randomized. Your test pins exactly this failure mode.

One small addition on my side: the new test casts a model to bfloat16 and runs CPU inference, which can hard-crash part of our Windows CI fleet on torch 2.13 and newer, so I added our `skip_bfloat16_cpu_crash` marker next to your older-torch xfail, one guard per end of the torch range.

Worth noting for completeness: even with the fix, the 710 unique scores show the remaining ceiling. A bfloat16 model produces logits on the bfloat16 grid, so the fix removes the damage the library adds on top, not the rounding the model itself already did.

Thanks again, this and #3892 make low-precision inference behave the way everyone assumed it already did!

- Tom Aarsen

KisuYang:
Thank you @tomaarsen for the thorough verification. The NanoBEIR numbers demonstrate the impact better than I could have, and the reshape in #3892 makes perfect sense. Happy to see these land in v6.0.

PR #3895, opened 2026-08-02
[`trainer`] Fix eval DataLoader worker leak with dataloader_persistent_workers

mjun0812:
Hello from Japan, @tomaarsen!

## Pull Request overview

* Fix a DataLoader worker process leak when training with `dataloader_persistent_workers=True` and a dict `eval_dataset`: every evaluation leaked `num_eval_datasets × dataloader_num_workers` worker processes, eventually exhausting file descriptors (`OSError: [Errno 24] Too many open files`) or host memory.
* Mirror the `transformers` eval dataloader caching (huggingface/transformers#30627, huggingface/transformers#39717): resolve dataset names in `get_eval_dataloader`, and reuse the prepared DataLoader per eval dataset name while persistent workers are enabled.
* Add a regression test following the style of the `transformers` test suite for this behaviour.

## Details

`BaseTrainer.get_eval_dataloader()` fully overrides the `transformers` implementation, which is necessary for the custom batch samplers, multi-dataset handling and `even_batches` toggling. However, the override predates the fork bomb protection that `transformers` added in huggingface/transformers#29538 / huggingface/transformers#30627 / huggingface/transformers#39717 (see huggingface/transformers#28469 for the original report), so it builds and `accelerator.prepare()`s a fresh DataLoader on every call. Since `Accelerator.prepare()` keeps a reference to every prepared dataloader in `accelerator._dataloaders`, the dataloaders are never garbage collected, and with `persistent_workers=True` their worker processes stay alive forever.

Reproduction with a tiny model, 3 eval datasets and 2 workers per dataloader (script below), on v5.6.1 (the code is identical on current `main`) with `transformers` v5.14.1 and `accelerate` v1.14.0:

```text
after evaluate #1: children=8   accelerator._dataloaders=3
after evaluate #2: children=14  accelerator._dataloaders=6
after evaluate #3: children=20  accelerator._dataloaders=9   <- +6 workers per evaluation
```

In a real training run (60 eval datasets × 4 workers, evaluating every 10k steps) this leaked ~230 processes per evaluation and repeatedly killed multi-day training runs through fd exhaustion and OOM.

## The fix

Two coordinated changes, matching the semantics `transformers` converged on:

1. `evaluate()` no longer materializes `self.eval_dataset` when no `eval_dataset` argument is passed. The superclass then recurses over dict eval datasets with their *names*, exactly like `transformers` does (huggingface/transformers#30627). `self.eval_dataset` is already preprocessed in `__init__`, so no preprocessing is lost. A `str` coming back from that recursion is passed through.
2. `get_eval_dataloader()` resolves `str` names via `self.eval_dataset[name]`, and while `dataloader_persistent_workers=True`, caches the **prepared** dataloader per dataset name (storing the prepared dataloader rather than the raw one follows huggingface/transformers#39717, which fixed the remaining leak in the original transformers fix).

Note that simply removing the override in favour of the `transformers` implementation is not possible, and neither is keeping `evaluate()` passing dataset objects: in that case all datasets of a dict collapse onto the `"eval"` cache key, and every dataset silently reuses the first dataset's dataloader, corrupting the per-dataset metrics. This is why the cache is keyed by dataset name and the name propagation in `evaluate()` is part of the fix. The added regression test asserts that each name resolves to its own dataset to guard against this.

With `dataloader_persistent_workers=False` (the default), behaviour is unchanged: no caching, dataloaders are built per evaluation as before.

After the fix, the reproduction shows stable worker counts and per-name cached dataloaders:

```text
after evaluate #1: children=8  accelerator._dataloaders=3
after evaluate #2: children=8  accelerator._dataloaders=3
after evaluate #3: children=8  accelerator._dataloaders=3
```

and a training run with periodic evaluation (18 steps, eval every 2 steps, 3 eval datasets) ends with the expected `1 + 3` prepared dataloaders instead of one per evaluation.

<details>
<summary>Reproduction script and steps</summary>

Runs on CPU only, no GPU required (`use_cpu=True` is set in the script; tested on macOS arm64 — the original failures occurred on Linux x86_64). The only dependency besides `sentence-transformers[train]` is `psutil` for counting child processes.

Save the script below as `repro_worker_leak.py`, then reproduce with a single [uv](https://docs.astral.sh/uv/) command (or install the same requirements with `pip` into a fresh environment):

```bash
# Reproduce the leak on the latest release
uv run --no-project --with "sentence-transformers[train]==5.6.1" --with psutil repro_worker_leak.py

# ... or against the current main branch
uv run --no-project --with "sentence-transformers[train] @ git+https://github.com/huggingface/sentence-transformers.git@main" --with psutil repro_worker_leak.py

# Verify the fix by installing directly from this PR's branch
uv run --no-project --with "sentence-transformers[train] @ git+https://github.com/mjun0812/sentence-transformers.git@fix-eval-dataloader-worker-leak" --with psutil repro_worker_leak.py
```

Expected output on v5.6.1 / current `main` (leak: +6 children per evaluation):

```text
[case persistent_workers=True] datasets=3, num_workers=2
  after evaluate #1: children=8, accelerator._dataloaders=3
  after evaluate #2: children=14, accelerator._dataloaders=6
  after evaluate #3: children=20, accelerator._dataloaders=9
```

Expected output with this PR (workers are reused, count stays flat):

```text
[case persistent_workers=True] datasets=3, num_workers=2
  after evaluate #1: children=8, accelerator._dataloaders=3
  after evaluate #2: children=8, accelerator._dataloaders=3
  after evaluate #3: children=8, accelerator._dataloaders=3
```

```python
"""Reproduce the eval DataLoader worker leak with persistent_workers.

Calls trainer.evaluate() repeatedly and counts child processes after each call.
With the transformers Trainer the eval dataloaders are cached and reused, so the
count stays flat; before this fix, sentence-transformers rebuilt them every time.
"""

import os
import sys
import tempfile


def build_trainer(persistent: bool, num_workers: int, n_datasets: int):
    import datasets as hf_datasets
    from transformers import BertConfig, BertModel, BertTokenizerFast
    from sentence_transformers import SentenceTransformer, SentenceTransformerTrainer
    from sentence_transformers.models import Transformer, Pooling
    from sentence_transformers.losses import MultipleNegativesRankingLoss
    from sentence_transformers.training_args import (
        BatchSamplers,
    )

    vocab = ["[PAD]", "[UNK]", "[CLS]", "[SEP]", "[MASK]"] + [f"tok{i}" for i in range(50)]
    vocab_dir = tempfile.mkdtemp()
    with open(os.path.join(vocab_dir, "vocab.txt"), "w") as f:
        f.write("\n".join(vocab))
    tok = BertTokenizerFast(vocab_file=os.path.join(vocab_dir, "vocab.txt"))
    config = BertConfig(
        vocab_size=len(vocab), hidden_size=8, num_hidden_layers=1,
        num_attention_heads=2, intermediate_size=16, max_position_embeddings=32,
    )
    BertModel(config).save_pretrained(vocab_dir)
    tok.save_pretrained(vocab_dir)

    model = SentenceTransformer(modules=[
        Transformer(vocab_dir, max_seq_length=16),
        Pooling(8),
    ])

    def make_ds(name: str, n: int):
        return hf_datasets.Dataset.from_dict({
            "anchor": [f"{name} anchor {i}" for i in range(n)],
            "positive": [f"{name} positive {i}" for i in range(n)],
        })

    val_datasets = {f"ds_{i}": make_ds(f"ds_{i}", 8) for i in range(n_datasets)}
    train_datasets = {f"ds_{i}": make_ds(f"ds_{i}", 8) for i in range(n_datasets)}

        output_dir=tempfile.mkdtemp(),
        per_device_eval_batch_size=2,
        dataloader_persistent_workers=persistent,
        dataloader_num_workers=num_workers,
        dataloader_pin_memory=False,
        report_to=[],
        remove_unused_columns=False,
        use_cpu=True,
    )

    return SentenceTransformerTrainer(
        model=model,
        args=args,
        train_dataset=train_datasets,
        eval_dataset=val_datasets,
        loss=MultipleNegativesRankingLoss(model),
    )


def run_case(persistent: bool, n_datasets: int = 3, num_workers: int = 2) -> None:
    import psutil

    trainer = build_trainer(persistent, num_workers, n_datasets)
    proc = psutil.Process()
    print(f"[case persistent_workers={persistent}] datasets={n_datasets}, num_workers={num_workers}")
    print(f"  before any evaluate: children={len(proc.children(recursive=True))}")
    for round_idx in range(3):
        trainer.evaluate()
        n_children = len(proc.children(recursive=True))
        n_dl = len(trainer.accelerator._dataloaders)
        print(
            f"  after evaluate #{round_idx + 1}: children={n_children}, "
            f"accelerator._dataloaders={n_dl}"
        )


if __name__ == "__main__":
    import transformers
    import sentence_transformers
    import accelerate
    import torch

    print(
        f"versions: transformers={transformers.__version__}, "
        f"sentence-transformers={sentence_transformers.__version__}, "
        f"accelerate={accelerate.__version__}, torch={torch.__version__}, "
        f"python={sys.version.split()[0]}, platform={sys.platform}"
    )
    run_case(persistent=True)
    print()
    run_case(persistent=False)
```

</details>

## Tests

* Added `test_trainer_get_eval_dataloader_with_persistent_workers` (parametrized over `persistent_workers`), following the style of the equivalent `transformers` test from huggingface/transformers#30627. It fails on `main` and passes with this PR.
* `tests/base/test_trainer.py`, `tests/sentence_transformer/test_trainer.py`, `tests/cross_encoder/test_trainer.py` and `tests/sparse_encoder/test_trainer.py` all pass locally (72 passed).

### DIFF
--- a/sentence_transformers/base/trainer.py
+++ b/sentence_transformers/base/trainer.py
@@ -619,14 +619,16 @@ class BaseTrainer(Trainer, ABC):
 
     def evaluate(
         self,
-        eval_dataset: Dataset | dict[str, Dataset] | None = None,
+        eval_dataset: str | Dataset | dict[str, Dataset] | None = None,
         ignore_keys: list[str] | None = None,
         metric_key_prefix: str = "eval",
     ) -> dict[str, float]:
-        if eval_dataset is not None:
+        # If eval_dataset is None, we keep it as None: with a dict `self.eval_dataset` (already preprocessed
+        # in __init__), the superclass then recurses per dataset with its *name*, so `get_eval_dataloader`
+        # can cache eval DataLoaders per dataset name when `dataloader_persistent_workers` is used.
+        # A str eval_dataset comes from that recursion and is resolved in `get_eval_dataloader`.
+        if eval_dataset is not None and not isinstance(eval_dataset, str):
             eval_dataset = self.preprocess_dataset(eval_dataset, dataset_name="eval")
-        else:
-            eval_dataset = self.eval_dataset
         return super().evaluate(eval_dataset, ignore_keys, metric_key_prefix)
 
     def evaluation_loop(
@@ -925,15 +927,18 @@ class BaseTrainer(Trainer, ABC):
         )
         return self._train_dataloader
 
-    def get_eval_dataloader(self, eval_dataset: Dataset | DatasetDict | IterableDataset | None = None) -> DataLoader:
+    def get_eval_dataloader(
+        self, eval_dataset: str | Dataset | DatasetDict | IterableDataset | None = None
+    ) -> DataLoader:
         """
         Returns the evaluation [`~torch.utils.data.DataLoader`].
 
         Subclass and override this method if you want to inject some custom behavior.
 
         Args:
-            eval_dataset (`torch.utils.data.Dataset`, *optional*):
-                If provided, will override `self.eval_dataset`. If it is a [`~datasets.Dataset`], columns not accepted
+            eval_dataset (`str` or `torch.utils.data.Dataset`, *optional*):
+                If a `str`, will use `self.eval_dataset[eval_dataset]` as the evaluation dataset. If provided
+                otherwise, will override `self.eval_dataset`. If it is a [`~datasets.Dataset`], columns not accepted
                 by the `model.forward()` method are automatically removed. It must implement `__len__`.
         """
         if eval_dataset is None and self.eval_dataset is None:
@@ -942,15 +947,38 @@ class BaseTrainer(Trainer, ABC):
                 return DataLoader([])
             raise ValueError(f"Evaluation requires specifying an eval_dataset to the {self.__class__.__name__}.")
 
+        # If we have persistent workers, don't do a fork bomb especially as eval datasets
+        # don't change during training. Mirrors the transformers Trainer behaviour, see
+        # https://github.com/huggingface/transformers/pull/30627 and
+        # https://github.com/huggingface/transformers/pull/39717
+        dataloader_key = eval_dataset if isinstance(eval_dataset, str) else "eval"
+        if (
+            hasattr(self, "_eval_dataloaders")
+            and dataloader_key in self._eval_dataloaders
+            and self.args.dataloader_persistent_workers
+        ):
+            return self._eval_dataloaders[dataloader_key]
+
+        if isinstance(eval_dataset, str):
+            eval_dataset = self.eval_dataset[eval_dataset]
         eval_dataset = eval_dataset if eval_dataset is not None else self.eval_dataset
 
         # If 'even_batches' is True, it will use the initial few samples to pad out the last sample. This can
         # cause issues with multi-dataset training, so we want to set this to False during training.
         # For evaluation, setting 'even_batches' to False results in hanging, so we keep it as True here.
         self.accelerator.even_batches = True
-        return self.accelerator.prepare(
+        dataloader = self.accelerator.prepare(
             self._build_dataloader(eval_dataset, self.args.eval_batch_size, dataset_kind="eval")
         )
+
+        # Store the prepared dataloader for subsequent evaluations if using persistent workers.
+        if self.args.dataloader_persistent_workers:
+            if hasattr(self, "_eval_dataloaders"):
+                self._eval_dataloaders[dataloader_key] = dataloader
+            else:
+                self._eval_dataloaders = {dataloader_key: dataloader}
+
+        return dataloader
 
     def get_test_dataloader(self, test_dataset: Dataset | DatasetDict | IterableDataset) -> DataLoader:
         """
--- a/sentence_transformers/base/trainer.py
+++ b/sentence_transformers/base/trainer.py
@@ -327,6 +327,7 @@ class BaseTrainer(Trainer, ABC):
             self.train_dataset = self.preprocess_dataset(train_dataset, dataset_name="train")
         if self.eval_dataset is not None:
             self.eval_dataset = self.preprocess_dataset(eval_dataset, dataset_name="eval")
+        self._eval_dataloaders: dict[str, DataLoader] = {}
         self.add_model_card_callback(default_args_dict)
 
     def get_data_collator(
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Thanks for this excellent PR! I worked with an agent to get this reviewed. Your analysis is spot on and the reproduction script made verification easy: I can reproduce the leak on `main` and confirm that with this branch the prepared dataloader count stays flat across repeated evaluations (tested on Windows with `transformers` v5.13.1 and `accelerate` v1.14.0). Mirroring the `transformers` approach of caching the prepared dataloader per dataset name is exactly the right call, also because it keeps our behavior aligned with theirs going forward.

One edge case turned up during review: every dataset that reaches `get_eval_dataloader` as an object rather than a name lands on the shared `"eval"` cache key. That happens in two scenarios:

1. On `transformers` 4.41.x (our current minimum), `Trainer.evaluate` recurses over dict eval datasets with the dataset objects themselves: the by-name recursion only landed in v4.42.0 via huggingface/transformers#30627.
2. On every version, `trainer.evaluate(eval_dataset=...)` with an explicit dict or dataset takes the `override=True` recursion, which also passes objects. The `transformers` `Trainer` on current `main` behaves the same way today.

In both cases, with persistent workers enabled, every dataset after the first would silently reuse the first dataset's cached dataloader, so all per-dataset metrics would report the first dataset's numbers. Before this PR those paths were correct (just leaky), and I did not want to trade the leak for silent metric corruption.

I pushed a commit on top with the following changes:

* `get_eval_dataloader` now only caches when the key actually identifies the dataset: a `str` name, `None`, or `self.eval_dataset` itself. Everything else keeps the previous build-per-call behavior. In practice, the leak is fixed for single eval datasets on all supported `transformers` versions and for dict eval datasets on v4.42.0 and newer, while dicts on 4.41.x keep the old (correct but leaky) behavior.
* Initialized `_eval_dataloaders` in `__init__` rather than using the `hasattr` checks, simplified the `str`/`None` resolution, and trimmed the comments down a bit.
* Extended your test with assertions that explicitly passed datasets are never cached in place of one another, and that nothing is cached without persistent workers.
* Added an end-to-end test that `evaluate()` really routes each dict entry to `get_eval_dataloader` by name and fills the cache, with a stubbed evaluation loop so no workers are spawned in CI. This guards the superclass recursion contract that the fix relies on.

Thanks again for the PR so far. Please let me know what you think, I'd like to incorporate this (and release it) soon.

- Tom Aarsen

mjun0812:
@tomaarsen

Thank you for the thorough review and the follow-up commit!

Your point about the paths where datasets are passed as objects is fair. I had tested several `transformers` versions on Linux and macOS, but I missed the older ones. Thank you for covering that with the additional fix.

I have read through your commit, and I think the PR is now better than before. Caching only under keys that really identify the dataset is a cleaner rule than what I had, and the end-to-end test that checks the by-name recursion is a nice guard.
I also verified the final state on macOS and Linux, including `transformers` 4.41.x / 4.42.x with the help of small monkeypatch(current `main` cannot construct a trainer on those versions as it is), and confirmed that the `cacheable` guard behaves as designed on every version I tried.

I also agree with the trade-off. Corrupted metrics are not acceptable, so keeping the leak on the object paths is the right call. I first noticed the leak in a run with 19 eval datasets, and later confirmed it with the 60-dataset setup described above. With a typical setup of one or a few eval datasets, training would likely finish before anyone notices, and that is probably also why this went unreported for so long.

As for the two Copilot comments, I believe both can be safely ignored. 

Thank you again, and I am looking forward to the release!

tomaarsen (approved):
I agree that the Copilot comments should be ignoreable. Let's move forward with this! You're in luck, because if everything goes well, this should be released later today.

- Tom Aarsen

PR #3897, opened 2026-08-03
Fix torchcodec runtime import failures

### LINKED ISSUE #3896: sentence-transformers>=5.0 hard-imports torchcodec at package load, crashing text-only usage when torchcodec/FFmpeg is broken
## Description

Since the multimodal backend landed, `import sentence_transformers` (and anything that imports it transitively, e.g. `bertopic`) hard-imports `torchcodec` with no `try`/`except`. If torchcodec's compiled shared library fails to load — a broken FFmpeg install, a PyTorch/torchcodec version mismatch, or simply no FFmpeg installed at all — the import raises a fatal `RuntimeError`, even for callers who only ever use plain text embeddings and never touch image/video/audio-multimodal features.

This is a real regression in practice, not just a theoretical concern: torchcodec failing to load with no FFmpeg installed is a documented, expected condition elsewhere in the ecosystem (e.g. `pyannote.audio` hits the exact same broken torchcodec install and handles it with a caught `UserWarning`, since it never actually needs torchcodec to work). Installing `sentence-transformers>=5.0` on the same machine turns that previously-harmless condition into a hard crash on plain `import sentence_transformers`.

## Repro

```
pip install sentence-transformers==5.6.1
python -c "from sentence_transformers import SentenceTransformer"
```

On a machine with a broken/missing FFmpeg (torchcodec installed but its shared library won't load):

```
RuntimeError: Could not load libtorchcodec. Likely causes:
          1. FFmpeg is not properly installed in your environment. ...
```

Traceback goes through:

```
sentence_transformers/__init__.py
  -> sentence_transformers/base/sampler.py (via base/__init__.py)
  -> sentence_transformers/base/model.py
  -> sentence_transformers/base/modality.py
  -> sentence_transformers/base/modality_types.py
       from torchcodec.decoders import AudioDecoder, VideoDecoder
```

## Expected behavior

A text-only `SentenceTransformer` import shouldn't require a working torchcodec/FFmpeg install. Wrapping the `torchcodec` import in `base/modality_types.py` in a `try`/`except`, with multimodal audio/video decoding raising a clear error only if actually invoked (rather than at package import time), would let text-only usage keep working the way it did before `>=5.0` — matching how `pyannote.audio` already handles the same underlying torchcodec fragility.

## Environment

- `sentence-transformers` 5.6.1 (also reproduced on 5.6.0)
- Windows, Python 3.13
- `torch` 2.10.0 (no FFmpeg installed)

## Workaround

Pinning to `sentence-transformers<5.0` (before the multimodal backend existed) avoids the import path entirely for text-only usage.

mturac:
Fixes #3896.

## Summary

- Treat `RuntimeError` from `torchcodec.decoders` as an unavailable optional dependency.
- Keep text-only package imports working when an installed torchcodec or FFmpeg setup is broken.
- Cover the package-import regression with an isolated subprocess test.

## Tests

- `python3 -m pytest -q tests/base/test_modality_types.py`
- `python3 -m pytest -q tests/base/test_model.py tests/base/test_model_card.py tests/base/test_modality_types.py -k 'not test_device_property'`
- `ruff check sentence_transformers/base/modality.py sentence_transformers/base/modality_types.py sentence_transformers/base/model_card.py tests/base/test_modality_types.py`
- `ruff format --check sentence_transformers/base/modality.py sentence_transformers/base/modality_types.py sentence_transformers/base/model_card.py tests/base/test_modality_types.py`


tests/base/test_modality_types.py
tomaarsen, line 1: Let's kill this test file, I'm not a fan of the format. I'm fine with accepting this without tests. Otherwise it looks good I think!

tomaarsen (changes requested):

tests/base/test_modality_types.py
mturac: Fixed in 04aecc65.

mturac:
Addressed the review feedback in 04aecc65.

tomaarsen (approved):
I think this is a solid improvement, thank you @mturac !

- Tom Aarsen

PR #3900, opened 2026-08-04
Expand the community window when similarities equal the threshold

### LINKED ISSUE #3899: community_detection truncates a community at sort_max_size when similarities equal the threshold
`community_detection` truncates a community at `sort_max_size` when similarities land exactly on the threshold:

```python
import torch
from sentence_transformers.util import community_detection

community_detection(torch.ones(200, 16), threshold=1.0, min_community_size=10)
# one community of 50, expected 200
```

In `util/retrieval.py` the window-widening loop uses `while top_val_large[-1] > threshold` while membership uses `top_val_large >= threshold`, so values equal to the threshold count as members but never widen the window. Changing the `while` to `>=` fixes it and `tests/util/` still passes.

Happy to send a PR with a test.

LK-maker-007:
Fixes #3899.

## Summary

- Widen the community candidate window while the smallest candidate is `>=` the threshold, matching the `>=` used to select members.
- Add a regression test covering `min_community_size` of 5, 10 and 30.

The window previously grew only while the smallest candidate was strictly greater than the threshold, while membership used `>=`. Similarities equal to the threshold were therefore included but never triggered an expansion, so a community was silently capped at `sort_max_size`.

```python
community_detection(torch.ones(200, 16), threshold=1.0, min_community_size=10)
# before: one community of 50
# after:  one community of 200
```

The `sort_max_size < len(embeddings)` guard is unchanged, so the loop still terminates.

The new test fails on `main` for all three parameters and passes with the fix. `tests/util/` is green (521 passed, 32 skipped), and `ruff check` / `ruff format --check` are clean. Timings on random and clustered inputs are unchanged; only the tie case behaves differently.

I don't have a GPU, so the CUDA/NPU branch is untested here. It counts members with `(cos_scores >= threshold).sum(1)` and takes `topk(k=row_wise_count.max())`, so by inspection it has no equivalent cap and should be unaffected.

### DIFF
--- a/sentence_transformers/util/retrieval.py
+++ b/sentence_transformers/util/retrieval.py
@@ -330,7 +330,7 @@ def community_detection(
                     top_val_large, top_idx_large = cos_scores[i].topk(k=sort_max_size, largest=True)
 
                     # Check if we need to increase sort_max_size
-                    while top_val_large[-1] > threshold and sort_max_size < len(embeddings):
+                    while top_val_large[-1] >= threshold and sort_max_size < len(embeddings):
                         sort_max_size = min(2 * sort_max_size, len(embeddings))
                         top_val_large, top_idx_large = cos_scores[i].topk(k=sort_max_size, largest=True)
 
--- a/tests/util/test_retrieval.py
+++ b/tests/util/test_retrieval.py
@@ -163,6 +163,25 @@ def test_community_detection_large_batch_size():
     assert all(len(community) >= 10 for community in result)
 
 
+@pytest.mark.parametrize("min_community_size", [5, 10, 30])
+def test_community_detection_similarities_equal_to_threshold(min_community_size: int):
+    """Members whose similarity equals the threshold exactly must not be truncated.
+
+    The candidate window grew only while the smallest candidate was strictly greater than
+    the threshold, while membership used >=. Similarities landing exactly on the threshold
+    were therefore eligible but never triggered an expansion, capping the community at
+    sort_max_size.
+    """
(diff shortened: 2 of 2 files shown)


LK-maker-007:
Some timings, since widening the window on ties means a larger `topk` and that is a fair thing to ask about.

CPU, torch 2.13.0, 4 threads, `min_community_size=10`, min of 7 runs:

| input | before | after | result before -> after |
| --- | --- | --- | --- |
| random 5000x64, `threshold=0.75` | 69 ms | 72 ms | `[]` -> `[]` |
| clustered 5000x64, `threshold=0.75` | 424 ms | 417 ms | 20x250 -> 20x250 |
| 5000 one-hot rows, `threshold=1.0` | 300 ms | 895 ms | `[50]` -> `[5000]` |

The first two are unchanged within run-to-run noise. The third is ~3x slower, but it is the case the fix is about: it now returns 5000 members instead of 50, so it is doing about 100x more useful work for 3x the time. Nothing gets slower where the old result was already correct.

Worth noting the trigger is narrow. It needs a similarity to land on the threshold exactly, which dense embeddings generally do not do (two identical sentences normalize to about 0.99999994, not 1.0). One-hot and sparse vectors do hit it exactly, which is where I ran into it.

tomaarsen:
Hello!

This looks like a solid fix to a reasonable bug. I also understand the timings, but they're not of concern to me. Obviously the 5k case is slower, but that's fine. I have no further feedback, PR looks great! I'll merge this when the tests go green.

- Tom Aarsen

PR #3904, opened 2026-08-04
[fix] Reject num_negatives larger than the range_min/range_max window in mine_hard_negatives

### LINKED ISSUE #3903: mine_hard_negatives crashes on a tensor shape when num_negatives exceeds the range_min/range_max window
`mine_hard_negatives` crashes on a tensor shape when `num_negatives` is larger than the `range_min`/`range_max` window:

```python
mine_hard_negatives(dataset, model, num_negatives=10, range_min=0, range_max=3)
# top:    RuntimeError: The expanded size of the tensor (3) must match the existing size (10)
# random: IndexError: index 6 is out of bounds for dimension 1 with size 3
```

The library can also reach this state by itself. With `use_faiss=True` and a corpus over 2048, an auto-derived `range_max` is capped to 2048, so `range_min=2040, num_negatives=10` leaves a window of 8:

```
Using FAISS, we can only retrieve up to 2048 documents per query. Setting range_max to 2048.
RuntimeError: The expanded size of the tensor (8) must match the existing size (10)
```

Neither message names `num_negatives`, `range_min` or `range_max`. Since only the candidates between `range_min` and `range_max` are kept, no query can supply more than `range_max - range_min` negatives, so this could be rejected up front like the other invalid arguments.

Happy to send a PR with a test.

LK-maker-007:
## Summary

Fixes #3903.

Only the candidates ranked between `range_min` and `range_max` are kept, so no query can supply more than `range_max - range_min` negatives. A larger `num_negatives` was not rejected, and instead failed further down on a tensor shape:

```python
mine_hard_negatives(dataset, model, num_negatives=10, range_min=0, range_max=3)
# sampling_strategy="top":    RuntimeError: The expanded size of the tensor (3) must match the existing size (10)
# sampling_strategy="random": IndexError: index 6 is out of bounds for dimension 1 with size 3
```

Neither message names the parameters involved. `mine_hard_negatives` can also reach the state on its own: with `use_faiss=True` and a corpus over 2048, an auto-derived `range_max` is capped to 2048, so `range_min=2040` with `num_negatives=10` leaves a window of 8 and crashes on a config the caller never wrote.

This raises a `ValueError` once `range_max` is resolved, alongside the existing argument checks, and mentions the FAISS cap when that is what shrank the window.

## Notes

- Covers both routes into the state (explicit `range_max`, and the auto-derived value capped by FAISS) and both sampling strategies.
- A window exactly equal to `num_negatives` is still valid and is covered by a test, so the check does not tighten anything that works today.
- No currently working call can start failing: every input that now raises used to crash a few lines later.
- `tests/util/` passes (620 passed, 31 skipped) and `ruff check` / `ruff format --check` are clean.

### DIFF
--- a/sentence_transformers/util/hard_negatives.py
+++ b/sentence_transformers/util/hard_negatives.py
@@ -239,7 +239,9 @@ def mine_hard_negatives(
         faiss_batch_size (int): Batch size of queries for the top-k similarity search, both with FAISS and without
             (``use_faiss=False``), where it bounds the size of the intermediate similarity matrix to
             ``faiss_batch_size * len(corpus)``. Defaults to 16384.
-        use_faiss (bool): Whether to use FAISS for similarity search. May be recommended for large datasets. Defaults to False.
+        use_faiss (bool): Whether to use FAISS for similarity search. May be recommended for large datasets. Requires
+            the ``faiss-cpu`` or ``faiss-gpu`` package. On GPU, FAISS can retrieve at most 2048 documents per query,
+            so an automatically derived ``range_max`` is capped to fit. Defaults to False.
         use_multi_process (bool | List[str], optional): Whether to use multi-GPU/CPU processing. If True, uses all GPUs if CUDA
             is available, and 4 CPU processes if it's not available. You can also pass a list of PyTorch devices like
             ["cuda:0", "cuda:1", ...] or ["cpu", "cpu", "cpu", "cpu"].
@@ -276,6 +278,15 @@ def mine_hard_negatives(
 
     if faiss_batch_size <= 0:
         raise ValueError(f"faiss_batch_size must be a positive integer, got {faiss_batch_size}.")
+
+    if use_faiss:
+        try:
+            import faiss
+        except ImportError as exc:
+            raise ImportError(
+                "Please install `faiss` to use this function with `use_faiss=True`: "
+                "`pip install faiss-cpu` (or `faiss-gpu`)."
+            ) from exc
 
     from datasets import Dataset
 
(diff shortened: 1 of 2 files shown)


LK-maker-007:
Small note while re-checking the rejected configs against main: the three window cases (`num_negatives` over the window, `range_min == range_max`, `range_min > range_max`) all crash today, so those guards just turn a shape error into a message.

`range_min=-5` is different though - on main it currently returns results rather than crashing, since the negative index slices the window from the wrong end like your test says. So that guard is a behaviour change rather than a nicer error. Fine by me either way, just flagging it in case it deserves a changelog line.

tomaarsen:
Hello!

Thanks for this PR, and well spotted. I can confirm the diagnosis: only the candidates ranked between `range_min` and `range_max` are kept, so requesting more negatives than that window holds can only ever fail. I reproduced both shape errors on `main`, and verified that this PR raises the new `ValueError` before any embedding work is done.

Beyond the commit I pushed earlier, I just pushed another one on top of your latest two. The full set of changes from my side:
* Added two sibling checks alongside this one: `range_min >= range_max` (the FAISS cap can push an auto-derived `range_max` below `range_min` as well) and `range_min < 0`. The latter was the nastiest of the bunch: a negative `range_min` did not crash at all, but `indices[:, range_min:]` slices from the wrong end, silently mining the easiest candidates in the window as "hard" negatives. You're right that this one is a behaviour change rather than a nicer error, so I'll call it out in the release notes.
* Reworked the `capped_by_faiss` flag into the note string itself, so every one of these errors can append it.
* Strengthened the exact-boundary test to assert that no rows are dropped.
* Took the Copilot remark (and your wording commit) one step further: the 2048 limit is indeed GPU-only, so the cap is now conditional on `faiss.get_num_gpus() > 0` (FAISS on CPU retrieves any k just fine). Its arithmetic was also off: the search retrieves `range_max + max_positives` documents per query, so even a capped `range_max=2048` sent k=2049 to FAISS and crashed on an actual GPU. I reproduced that crash on v5.6.1 with an RTX 3090 and verified that this branch mines successfully there with the cap at `2048 - max_positives`. As a bonus, `use_faiss=True` without faiss installed now raises a clear ImportError up front, rather than crashing only after the whole corpus was encoded.

Your two follow-up commits slot in nicely with that: the `num_negatives < 1` check is a good catch (a 0 used to quietly return the dataset with no negative columns at all), and restoring the `sampling_strategy` parametrization is fair, especially on the exact-boundary test where the two strategies genuinely take different paths.

- Tom Aarsen

PR #3906, opened 2026-08-04
[`fix`] Support mixed sparse/dense inputs in euclidean and manhattan similarity

LuShadowX:
`euclidean_sim` and `manhattan_sim` (and their pairwise variants) raise when one operand is sparse and the other is dense, while `cos_sim` and `dot_score` handle that combination fine. `SparseEncoder` documents `similarity_fn_name` as `Literal["cosine", "dot", "euclidean", "manhattan"]` and `encode` exposes `convert_to_sparse_tensor=False`, so mixing the two representations is reachable from public API:

```python
from sentence_transformers import SparseEncoder

model = SparseEncoder("sparse-encoder-testing/splade-bert-tiny-nq")
query = model.encode_query(["What is the capital of France?"])
docs = model.encode_document(["Paris is the capital of France.", "Berlin is the capital of Germany."],
                             convert_to_sparse_tensor=False)

model.similarity_fn_name = "euclidean"
model.similarity(query, docs)   # NotImplementedError: Could not run 'aten::_indices' ...
model.similarity_fn_name = "manhattan"
model.similarity(query, docs)   # RuntimeError: coalesce expected sparse coordinate tensor layout but got Strided
```

Three separate causes:

- `manhattan_sim` enters its sparse branch when *either* input is sparse, but then calls `to_scipy_coo` on both, which requires a COO tensor.
- `euclidean_sim` only checks `a.is_sparse`. With a dense `a` and sparse `b` it falls through to `torch.cdist`, which has no sparse kernel; with a sparse `a` and dense `b` its `torch.sparse.sum(b * b, ...)` gets a strided tensor.
- `pairwise_manhattan_sim` and `pairwise_euclidean_sim` compute `a - b`, and subtraction between a sparse and a dense operand is unsupported.

The fix keeps every existing path byte-for-byte and only handles the mixed case: `manhattan_sim` sparsifies whichever operand is dense before converting to SciPy, `euclidean_sim` computes its row norms through a small layout-agnostic helper and now tests both operands, and the two pairwise functions densify the sparse side. For the pairwise ones both inputs have the same shape, so the dense operand already accounts for that memory; the matrix versions keep their sparse paths so an (N, D) x (M, D) comparison is never densified.

Verified on the model above: `euclidean` and `manhattan` now return `[-2.8787, -5.6926]` and `[-18.4137, -43.7693]` for both sparse/sparse and sparse/dense, matching what the all-sparse path already returned.

## Tests

`test_similarity_mixed_sparse_dense` is parametrized over the four metrics and both mixed orders, asserting the mixed result matches the all-dense result for the matrix and pairwise form. It fails on `main` for euclidean and manhattan in both orders (4 failed, 4 passed) and passes with this change.

```sh
pytest tests/util/test_similarity.py
# 25 passed

pytest tests/util/ tests/sparse_encoder/test_model.py tests/sparse_encoder/losses/test_sparse_cosent.py \
       tests/sentence_transformer/evaluation/test_binary_classification_evaluator.py \
       tests/sentence_transformer/evaluation/test_information_retrieval_evaluator.py
# 968 passed, 36 skipped   (960 passed, 36 skipped on main — the 8 new cases are the only difference)

pre-commit run --files sentence_transformers/util/similarity.py tests/util/test_similarity.py
# ruff check / ruff format / typos all Passed
```

I ran the tests that touch these functions rather than the whole suite, since the full run pulls a lot of models. Everything here was on CPU; I have no GPU to check the CUDA paths.

### DIFF
--- a/sentence_transformers/util/similarity.py
+++ b/sentence_transformers/util/similarity.py
@@ -16,6 +16,41 @@ from .tensor import _convert_to_batch_tensor, _convert_to_tensor, normalize_embe
 logger = logging.get_logger(__name__)
 
 
+def _row_squared_norms(a: Tensor) -> Tensor:
+    """
+    Computes the row-wise sum of squares for either a sparse or a dense tensor.
+
+    Args:
+        a (Tensor): A 2-dimensional sparse or dense tensor.
+
+    Returns:
+        Tensor: Dense vector with res[i] = sum(a[i] ** 2)
+    """
+    if a.is_sparse:
+        return torch.sparse.sum(a * a, dim=1).to_dense()
+    return (a * a).sum(dim=1)
+
+
+def _match_layouts(a: Tensor, b: Tensor) -> tuple[Tensor, Tensor]:
+    """
+    Densifies the sparse tensor if exactly one of the two inputs is sparse, as element-wise operations require both
+    operands to share a layout. Both inputs have the same shape here, so the dense one already accounts for the
+    memory that densifying the other one takes.
+
+    Args:
+        a (Tensor): The first tensor.
+        b (Tensor): The second tensor.
+
+    Returns:
+        tuple[Tensor, Tensor]: The two tensors, using the same layout.
+    """
+    if a.is_sparse and not b.is_sparse:
+        return a.to_dense(), b
+    if b.is_sparse and not a.is_sparse:
+        return a, b.to_dense()
+    return a, b
+
+
 def pytorch_cos_sim(a: Tensor, b: Tensor) -> Tensor:
     """
     Computes the cosine similarity between two tensors.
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Thanks for this PR, and well spotted. I can confirm the diagnosis: I reproduced all three failure causes on `main`, including your exact "4 failed, 4 passed" on the new test, and this branch fixes them. I was also able to run everything on a GPU, which you mentioned you couldn't: the test fixture moves the tensors to CUDA there, and the mixed paths all pass on that side as well.

I did push a commit on top that reworks the strategy while leaving your tests byte-for-byte intact:

* Rather than densifying the sparse operand in the pairwise functions, `_match_layouts` now converts the lone dense operand to sparse COO, and all four functions route through it. In every mixed case actually reachable from `SparseEncoder`, the dense tensor is just sparse data in a dense container (`convert_to_sparse_tensor=False`), so the conversion is nearly free and the sparse path keeps its memory profile: on 4096x30522 inputs with ~150 active dims per row, the pairwise functions now peak at 88 MB of extra CUDA memory instead of 1000 MB, at comparable speed (a few ms either way). The tradeoff is that a truly dense operand runs slower through the sparse ops, but a genuinely dense embedding can't shape-match a 30522-dim sparse one, and that combination used to crash outright.
* This lets `euclidean_sim` keep its original body (no `_row_squared_norms` needed), makes mixed inputs hit exactly the same code as all-sparse inputs everywhere (including the same autograd behavior), and avoids `torch.matmul(dense, sparse)`, which requires a newer torch than our declared minimum.
* I extended the same treatment to `pairwise_angle_sim`, which only checked `x.is_sparse` and crashed on mixed inputs in both argument orders, with a matching test. Densifying is the right direction for that one, as it converts to dense regardless. I also dropped the `coalesce()` calls there, since `to_dense()` already sums duplicate entries itself.

Your test file is untouched and still passes as-is, which is the best kind of evidence that this is an internal rework rather than a behavior change.

- Tom Aarsen

LuShadowX:
thanks for quick review and merge. happy to help!

PR #3907, opened 2026-08-05
[fix] Keep queries aligned in semantic_search_seismic when a query matches nothing

LK-maker-007:
## Summary

`semantic_search_seismic` raises `IndexError` when a query matches no document:

```python
semantic_search_seismic(
    [[("fruit", 1.0)], [("nomatch", 1.0)], [("veg", 1.0)]],
    corpus_embeddings_decoded=corpus,
    top_k=2,
)
# IndexError: list index out of range   (search_engines.py:412)
```

Seismic returns the queries out of order, so the results were sorted with `sorted(results, key=lambda x: int(x[0][0]))`. A query that matches nothing comes back as an empty list, and an empty list carries no query id, so there is nothing to sort it by. Each row is now placed by the query id it already contains, which drops the sort and keeps empty results in the right position instead of shifting the other queries.

I also corrected the `corpus_index` type hint and docstring. They describe a `tuple[SeismicIndex, str]`, but the function builds, uses and returns a bare `SeismicIndex`, so passing the documented tuple raises `AttributeError: 'tuple' object has no attribute 'batch_search'`. Worth noting the other three engines here really do use `(client, name)` tuples, so if you would rather Seismic match them I am happy to change the code instead of the docstring.

## Notes

- Not a behaviour change: every input that returned results before returns identical results now, only the crashing cases differ. Checked against real `pyseismic-lsr` on single/multiple queries, all-empty results, and `top_k` larger than the corpus.
- The tests stub `batch_search` rather than requiring Seismic, since it is not a test dependency. The stubbed rows use the shape a real index returns, including the out-of-order rows.
- `tests/sparse_encoder/` passes (361 passed, 4 skipped), `ruff check` and `ruff format --check` are clean.

### DIFF
--- a/sentence_transformers/sparse_encoder/search_engines.py
+++ b/sentence_transformers/sparse_encoder/search_engines.py
@@ -299,15 +299,12 @@ def semantic_search_elasticsearch(
 def semantic_search_seismic(
     query_embeddings_decoded: list[list[tuple[str, float]]],
     corpus_embeddings_decoded: list[list[tuple[str, float]]] | None = None,
-    corpus_index: tuple[SeismicIndex, str] | None = None,
+    corpus_index: SeismicIndex | None = None,
     top_k: int = 10,
     output_index: bool = False,
     index_kwargs: dict[str, Any] | None = None,
     search_kwargs: dict[str, Any] | None = None,
-) -> (
-    tuple[list[list[dict[str, int | float]]], float]
-    | tuple[list[list[dict[str, int | float]]], float, tuple[SeismicIndex, str]]
-):
+) -> tuple[list[list[dict[str, int | float]]], float] | tuple[list[list[dict[str, int | float]]], float, SeismicIndex]:
     """
     Performs semantic search using sparse embeddings with Seismic.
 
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Thanks for this PR, and well spotted. I can confirm the diagnosis: with the real `pyseismic-lsr`, a query that matches nothing comes back as an empty row that carries no query id, `main` crashes with your exact `IndexError` at the sort, and this branch returns aligned results with the empty row in the right position. I also checked your stub against the real package: the row shape matches exactly (one row per query, `(query_id, score, doc_id)` tuples, an empty list on a miss), and because the stub intercepts the lazy `from seismic import ...`, the tests run without Seismic installed. That works well for CI, where Seismic isn't available.

On your question about matching the other engines: let's keep it as you have it. The `(client, name)` tuples exist because those three are server-based engines with named collections, while Seismic is an in-process index with no collection concept, so the bare `SeismicIndex` is the natural interface. The documented tuple never worked anyway, and the shipped example already round-trips the bare index.

- Tom Aarsen

PR #3909, opened 2026-08-06
[fix] Treat a 1-dimensional query embedding as a single query in semantic_search_qdrant

LK-maker-007:
`semantic_search_qdrant` loops with `for q_idx in range(query_embeddings.size(0))`. Encoding a single query returns a 1-dimensional tensor, where `size(0)` is the vocabulary size rather than the number of queries, so the loop runs once per vocabulary entry.

With a 30522 token vocabulary that is 30522 searches against Qdrant for one query, and 30522 identical rows in the returned list instead of one. Nothing raises, so it surfaces as a very slow search and a result list of the wrong length.

```python
model = SparseEncoder("sparse-encoder-testing/splade-bert-tiny-nq")
corpus = model.encode_document(corpus_texts)

query = model.encode_query("what is the capital of france")  # 1-dimensional
results, _ = semantic_search_qdrant(query, corpus_embeddings=corpus, top_k=5)
len(results)  # 30522, expected 1
```

Passing `["what is the capital of france"]` gives a 2-dimensional tensor and works correctly, so this only affects the single query form.

The function already had a `sparse_dim() == 1` branch inside the loop to read a 1-dimensional tensor, so this input was expected, but the loop bound was never adjusted to match. This promotes the tensor to a batch of one right after the input validation, which is how `decode` and `sparsity` already handle 1-dimensional input, and drops the branch that is now unreachable.

The tests stub the Qdrant client the same way the Seismic tests stub Seismic, so they add no test dependency. They cover the single query, a batch of queries, and that the promoted tensor searches for the same vector as an explicit batch of one.

### DIFF
--- a/sentence_transformers/sparse_encoder/search_engines.py
+++ b/sentence_transformers/sparse_encoder/search_engines.py
@@ -68,8 +68,6 @@ def semantic_search_qdrant(
     if not query_embeddings.is_sparse or query_embeddings.layout != torch.sparse_coo:
         raise ValueError("Query embeddings must be a sparse COO tensor")
 
-    # Encoding a single query returns a 1-dimensional tensor, whose size(0) is the vocabulary rather
-    # than the number of queries. Promoting it to a batch of one keeps the loop below over queries.
     if query_embeddings.ndim == 1:
         query_embeddings = query_embeddings.unsqueeze(0)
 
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Well spotted, and thanks for the clean fix. I pushed a small cleanup commit to your branch: the query indices and values are now extracted once before the search loop instead of once per query (like the corpus indexing block above), and I removed the comment above the promotion as the promotion itself is pretty clear.

Looks good to me, I'll merge once CI is green. Thanks again!

- Tom Aarsen

PR #3910, opened 2026-08-06
[`fix`] Prevent CSR inference from updating dead-feature statistics

eSVeeF:
### What does this change?

`SparseAutoEncoder.top_k()` was updating `stats_last_nonzero` even during inference. That buffer is only meant to track dead features for the auxiliary loss, so running inference could end up affecting later training.

This changes it so the buffer is only updated during training (with gradients enabled). Inference leaves it alone, while training behavior stays the same.

### Why?

Without this, just running inference or using the model for encoding between training steps could change what happens during the next training step. Since these stats are only used for training, it seems better if inference doesn't touch them. That makes the CSR module's behavior a bit more predictable and avoids this kind of hidden state change.

### Tests

Added regression coverage to verify that:

- inference does not update `stats_last_nonzero`
- a training forward still updates it
---
`pytest -q test_csr.py -m 'not slow'`  `3 passed`
`pytest -q test_model.py -m 'not slow'` `339 passed, 2 warnings`

### DIFF
--- a/sentence_transformers/sparse_encoder/modules/sparse_auto_encoder.py
+++ b/sentence_transformers/sparse_encoder/modules/sparse_auto_encoder.py
@@ -123,16 +123,17 @@ class SparseAutoEncoder(Module):
         z_topk = torch.zeros_like(x)
         z_topk.scatter_(-1, topk.indices, topk.values)
         latents_k = F.relu(z_topk)
-        ## set num nonzero stat ##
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello! Sorry for the delayed response.

This is a clean fix, I think. I updated the tests a bit to cover more cases, but otherwise this is totally ready. Thanks!

- Tom Aarsen

tomaarsen (approved):

PR #3913, opened 2026-08-06
[fix] Save the external data of an ONNX model over 2GB when exporting

### LINKED ISSUE #3438: Not able to get optimized onnx model weights file
```
from sentence_transformers import SentenceTransformer, export_optimized_onnx_model

def main():

    model_name = "BAAI/bge-m3"
    model = SentenceTransformer(model_name, backend="onnx")

    export_optimized_onnx_model(
     model,
     "O1",
     model_name+"_o1",
     )

    
if __name__ == "__main__":
    main()
```

When i run the above code, in the output folder there is only `model_O1.onnx`, there is not `model_O1.onn.data`, which containes the weights.
Please let me know whether am i missing out anything

The sentence transformer version is : 4.1.0
optimum version is :1.26.1
onnxruntime-gpu version is : 1.19.2

LK-maker-007:
Fixes #3438.

`save_or_push_to_hub_model` moves only the `.onnx` file into the export directory. A model over the 2GB protobuf limit keeps its weights in a separate file beside it, so that file gets left behind and the exported model cannot be loaded. Nothing raises.

This builds on the diagnosis already in the issue - @xenova identified the external data mechanism and the `.onnx.data` / `.onnx_data` naming difference back in July, and @Danial-Alh pointed at the two lines. @rkoystart's `optimum-cli` comparison is what showed the loss is on our side rather than Optimum's.

**First commit: save the files the model refers to**

The locations are read out of the ONNX file rather than derived from its name, because the name depends entirely on what wrote it:

- the exporter writes `model.onnx_data`
- `ORTOptimizer` writes `model_O1.onnx.data` (`output_path + ".data"` in `OnnxModel.save`)
- with `all_tensors_to_one_file=False` the names come from the tensor names
- and when `onnx.save_model` picks the name itself, it is a UUID - I hit `160ea76f-918f-11f1-8428-cc15316249d1.data` while testing a 2.4GB model

Optimum also passes `convert_attribute=True`, so tensors held by node attributes get externalized as well. Those are not initializers, so I use `onnx.external_data_helper._get_all_tensors` to reach them.

Reproducing this normally needs a model over 2GB. `size_threshold=0` asks for the same layout on any model, which is what the tests use, so they need no large download and no network. I did also check a real 2.4GB model separately: it exports, loads with its weights, and computes the right answer.

**Second commit: name the file the way Optimum fetches it**

Saving the file is not enough for `push_to_hub=True`. `ORTModel._cached_file` requests `filename + "_data"` and nothing else, inside `contextlib.suppress(OSError)`, so a sidecar named `model_O1.onnx.data` 404s silently and the model loads without its weights. This commit renames the file and rewrites the location recorded in the model.

I pushed a small model to a private repo three times to check:

```
main      uploaded  onnx/model_O1.onnx
          optimum requested onnx/model_O1.onnx_data -> EntryNotFoundError
          FAILS: External data path validation failed for initializer: W

commit 1  uploaded  onnx/model_O1.onnx, onnx/model_O1.onnx.data
          optimum requested onnx/model_O1.onnx_data -> EntryNotFoundError
          FAILS: External data path validation failed for initializer: W

commit 2  uploaded  onnx/model_O1.onnx, onnx/model_O1.onnx_data
          optimum fetched both
          LOADS AND RUNS
```

The middle one is what I would not have believed without running it - the weights file is on the Hub and the model still fails.

I kept this as a separate commit so it can be dropped if you would rather see it fixed in Optimum, since there is already a TODO about the naming in `optimum/onnxruntime/optimization.py`.

**Things I am unsure about**

- `_get_all_tensors` is private. Optimum imports from the same module in `optimum/onnx/utils.py`, and it is what `onnx.save` walks itself, but it is still private.
- A location that is absolute or contains `..` raises. onnx refuses to write those, so I had to edit the protobuf by hand to test it - defensive, not something the current toolchain produces.
- A location naming a file that was never written warns and is skipped rather than raising. The model was already unloadable at that point, so I did not want to turn a broken export from another tool into a crash here. Happy to change it to raise.
- `export_dynamic_quantized_onnx_model` is not affected - `quantize.py` never passes `use_external_data_format`, which defaults to `False` - so the loop is a no-op there. Mentioning it since it is the other caller.

Tests are in `tests/backend/test_utils.py` and skip when onnx is not installed, so they run on the `transformers<5.0.0` legs.

Please tell me if any of this is wrong.

### DIFF
--- a/sentence_transformers/backend/utils.py
+++ b/sentence_transformers/backend/utils.py
@@ -135,74 +135,27 @@ def backend_should_export(
     return export, model_kwargs
 
 
-def _onnx_external_data_locations(model_path: Path) -> list[str]:
+def _onnx_prepare_external_data(model_path: Path) -> list[str]:
     """
-    Return the locations of the external data files that an ONNX model refers to.
+    Point an ONNX model over the 2GB protobuf limit at its weights under the name Optimum fetches.
 
-    A model larger than the 2GB protobuf limit keeps its weights in separate files and records their
-    locations, relative to itself, inside the ONNX file. Those locations differ between tools, so
-    they are read from the model rather than derived from the model file name.
+    Such a model keeps its weights in files beside itself and records where, so the locations are read
+    out of the model rather than derived from its name, which tools disagree on. ONNX Runtime writes
+    ``<model>.onnx.data`` for an optimized model, but Optimum only ever requests ``<model>.onnx_data``
+    from the Hugging Face Hub, and fails silently, so a lone file is renamed to that along with the
+    location recorded in the model. Rewriting the model does not read the weights, so this stays cheap.
 
     Args:
-        model_path: The ONNX file to read the references from.
+        model_path: The ONNX file to read the references from, and to rewrite if a file is renamed.
 
     Returns:
-        list[str]: The locations of the referenced files, relative to the ONNX file. Empty if the
-            weights are stored in the ONNX file itself. A location without a file behind it is
-            skipped with a warning: the model naming it cannot be loaded either way.
-
-    Raises:
-        ValueError: If a location is not relative to the directory holding the model, which ONNX
-            forbids and refuses to load.
+        list[str]: The files holding the weights, relative to the ONNX file, to travel alongside it.
+            Empty if the weights are in the ONNX file itself.
     """
     import onnx
     from onnx.external_data_helper import _get_all_tensors, uses_external_data
 
-    # Tensors held by node attributes get externalized alongside the initializers, so every tensor
-    # of the model has to be considered
-    model = onnx.load(model_path.as_posix(), load_external_data=False)
-    locations = {
-        entry.value
-        for tensor in _get_all_tensors(model)
-        if uses_external_data(tensor)
-        for entry in tensor.external_data
-        if entry.key == "location"
-    }
-
-    existing_locations = []
-    for location in sorted(locations):
-        if Path(location).is_absolute() or ".." in Path(location).parts:
-            raise ValueError(
-                f"{model_path.name!r} refers to weights at {location!r}, but ONNX only loads external "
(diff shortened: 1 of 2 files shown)


tomaarsen:
Thanks for this, the diagnosis is right. I reproduced it on `main` first (forced a bert-tiny export into the external data layout, then `export_optimized_onnx_model` saved `onnx/model_O1.onnx` and silently dropped the weights) and confirmed your branch fixes it.

I pushed two commits on top.

The first merges the two helpers into `_onnx_prepare_external_data`. Both loaded the ONNX file and built the same list of location entries, so a model with inline weights got parsed twice to learn that it has no external data. One pass takes 0.76s instead of 1.48s on a 1.46 GB model.

The second drops the guards that cannot fire. `save_or_push_to_hub_model` is only reached from `export_optimized_onnx_model` and `export_dynamic_quantized_onnx_model` (plus the OpenVINO one, which skips the ONNX branch), and Optimum writes at most one file under a plain name beside the model. That makes the checks for an absolute location, one escaping the model directory, and one with no file behind it unreachable, so the docstring records the assumption instead. Your instinct there was sound, it just has no way in from those two callers. The missing file case now raises rather than warning, which I prefer for a model that was unloadable either way.

On two of your open questions: `_get_all_tensors` being private is fine, Optimum imports `_get_initializer_tensors` and `uses_external_data` from that same module. And you are right that `export_dynamic_quantized_onnx_model` is unaffected, `use_external_data_format` defaults to `False` there.

The second commit stays. `ORTModel._cached_file` only requests `filename + "_data"` inside `contextlib.suppress(OSError)`, so without the rename `push_to_hub=True` is still broken. If Optimum fixes the naming later, the rename skips itself once the name already matches.

- Tom Aarsen

Issue #3457, opened 2025-07-22
Using sentence-transformers with TPU: RuntimeError: Cannot set version_counter for inference tensor


wirthual:
I am trying to run `sentence-transformers` on a TPU. 

Executing this simple example:
```python
from sentence_transformers import SentenceTransformer
import torch_xla.core.xla_model as xm
import torch
import time


print(xm.get_xla_supported_devices("TPU"))


device = xm.xla_device()
#device = "cpu"
model = SentenceTransformer("all-MiniLM-L6-v2",device=device)

sentences = ['This is an example sentence', 'Each sentence is converted'] 

times = []
for i in range(10):
  s = time.time()
  embeddings = model.encode(sentences,device=device,convert_to_tensor=True)
  embeddings.to("cpu")
  e = time.time()
  times.append(e-s)

print(max(times),min(times),sum(times)/len(times))
```

This results in the mentioned error.

I managed to get around the problem by removing `@torch.inference_mode` from the [encode method](https://github.com/UKPLab/sentence-transformers/blob/49a60f950fa54206f6e26d1e99c0a213a1d43c20/sentence_transformers/SentenceTransformer.py#L826C1-L826C28)

See Example [colab](https://colab.research.google.com/drive/1Ok4ayZyZw2v9uIJ9P9t8KdPjT9M57Jww?usp=sharing) to reproduce.

Tried on v6 and v2 TPUs with same outcome.

<details>

<summary>Full output</summary>

```
/usr/local/lib/python3.11/dist-packages/torch_xla/core/xla_model.py:99: UserWarning: `devkind` argument is deprecated and will be removed in a future release.
  warnings.warn("`devkind` argument is deprecated and will be removed in a "

['xla:0']

/usr/local/lib/python3.11/dist-packages/huggingface_hub/utils/_auth.py:94: UserWarning: 
The secret `HF_TOKEN` does not exist in your Colab secrets.
To authenticate with the Hugging Face Hub, create a token in your settings tab (https://huggingface.co/settings/tokens), set it as secret in your Google Colab and restart your session.
You will be able to reuse this secret in all of your notebooks.
Please note that authentication is recommended but still optional to access public models or datasets.
  warnings.warn(

---------------------------------------------------------------------------

RuntimeError                              Traceback (most recent call last)

[/tmp/ipython-input-2-3307924959.py](https://localhost:8080/#) in <cell line: 0>()
     17 for i in range(10):
     18   s = time.time()
[...]

Issue #3822, opened 2026-06-15
Explicit device assignment ignored in multi-GPU (PyTriton) environment


alejandrojcastaneira:
I am running multiple SentenceTransformer models in a single Python process (via PyTriton). My goal is to orchestrate different models onto different physical GPUs (e.g., cuda:0, cuda:1).

Despite passing an explicit device argument (e.g., device='cuda:1') and attempting to use torch.cuda.set_device() context managers, I observe significant memory allocation on cuda:0 (the default device) that scales with the model size, suggesting the model weights or their initialization stages are being staged on the wrong GPU before being moved.

Environment:

```
sentence-transformers version: 5.5.1

torch version: 2.12.0

OS: Ubuntu 22.04
```

Expected Behavior:
All model weights and internal buffers should be allocated exclusively on the target GPU provided in the device parameter.

Actual Behavior:
The models consistently allocate memory on cuda:0. I have verified this using nvidia-smi. The memory footprint on cuda:0 matches the size of the model weights, confirming this is not just the standard "Global CUDA Context" overhead, but a genuine staging issue.

What I have already tried:
To isolate the issue, I have attempted the following:

- Manual Module Injection: Initializing models.Transformer and models.Pooling separately and moving them .to(device) before passing them to the SentenceTransformer constructor.

- Hardware Lockdown: Using torch.cuda.set_device(gpu_id) context managers during the initialization phase.

- RAM-First Loading: Loading the model with device='cpu' first, then calling .to(device) before the first inference.

- model_kwargs: Injecting device_map via AutoModel.from_pretrained (when bypassing the SentenceTransformer constructor).

none of these prevent the initial weight staging/allocation on cuda:0.

Tobi-Adesoye:
@alejandrojcastaneira This device-assignment bypass under PyTriton usually points to a lower-level mismatch between Triton’s inter-process memory allocation layers and PyTorch’s virtual block caching mechanism (cudaMallocManaged pooling).
When asynchronous inference hooks dispatch ragged batch shapes or dynamic request sizes concurrently, the runtime frequently clones or instantiates hidden operational layouts under the hood.
We isolated similar behavior when scaling dynamic batching pipelines across multi-device host targets.

[...]

PR #3918, opened 2026-08-07
[fix] Skip the device move in encode when the model is already there

### LINKED ISSUE #3908: `RuntimeError: _apply(): Couldn't swap Linear.weight` when `torchao` with `torch.compile`
```py
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\torch\nn\modules\module.py", line 981, in _apply
                                     param_applied = torch.nn.Parameter(
                                         # pyrefly: ignore [bad-argument-type]
                                         param_applied,
                                         requires_grad=param.requires_grad,
                                     )
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\torch\nn\parameter.py", line 60, in __new__
                                     t = data.detach().requires_grad_(requires_grad)
                                         ~~~~~~~~~~~^^
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\torchao\utils.py", line 664, in _dispatch__torch_function__
                                     return func(*args, **kwargs)
                                 RuntimeError: Cannot set version_counter for inference tensor

                                 The above exception was the direct cause of the following exception:

                                   File "e:\Sandbox\AI classification3.py", line 1288, in gpu_worker
                                     out = await loop.run_in_executor(pool, func, args)
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\concurrent\futures\thread.py", line 59, in run
                                     result = self.fn(*self.args, **self.kwargs)
                                   File "e:\Sandbox\AI classification3.py", line 922, in encode_document
                                     embed = self.model.encode_document(
                                         args,
                                     ...<17 lines>...
                                         convert_to_tensor=True,
                                     )
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\sentence_transformers\util\decorators.py", line 41, in wrapper
                                     return func(*args, **kwargs)
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\sentence_transformers\sentence_transformer\model.py", line 323, in encode_document
                                     return self.encode(
                                            ~~~~~~~~~~~^
                                         inputs=inputs,
                                         ^^^^^^^^^^^^^^
                                     ...<14 lines>...
                                         **kwargs,
                                         ^^^^^^^^^
                                     )
                                     ^
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\torch\utils\_contextlib.py", line 124, in decorate_context
                                     return func(*args, **kwargs)
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\sentence_transformers\util\decorators.py", line 41, in wrapper
                                     return func(*args, **kwargs)
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\sentence_transformers\sentence_transformer\model.py", line 641, in encode
                                     self.to(device)
                                     ~~~~~~~^^^^^^^^
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\torch\nn\modules\module.py", line 1383, in to
                                     return self._apply(convert)
                                            ~~~~~~~~~~~^^^^^^^^^
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\torch\nn\modules\module.py", line 933, in _apply
                                     module._apply(fn)
                                     ~~~~~~~~~~~~~^^^^
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\torch\nn\modules\module.py", line 933, in _apply
                                     module._apply(fn)
                                     ~~~~~~~~~~~~~^^^^
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\torch\nn\modules\module.py", line 933, in _apply
                                     module._apply(fn)
                                     ~~~~~~~~~~~~~^^^^
                                   [Previous line repeated 4 more times]
                                   File "e:\Sandbox\pixi\.pixi\envs\default\Lib\site-packages\torch\nn\modules\module.py", line 990, in _apply
                                     raise RuntimeError(
                                         f"_apply(): Couldn't swap {self._get_name()}.{key}"
                                     ) from e
                                 RuntimeError: _apply(): Couldn't swap Linear.weight
```
```py
        from transformers import TorchAoConfig
        from torchao.quantization import Int8WeightOnlyConfig

        torch.backends.cuda.enable_math_sdp(False)
        quant_config_int8wo = TorchAoConfig(Int8WeightOnlyConfig(version=2))

        model = SentenceTransformer(
            "Qwen/Qwen3-VL-Embedding-2B",
            cache_folder=r"E:\Sandbox\##SENTENCE_TRANSFORMERS_HOME",
            local_files_only=True,
            device="cuda",
            model_kwargs={
                "torch_dtype": torch.bfloat16,
                "low_cpu_mem_usage": True,
                "attn_implementation": "sdpa",
                "device_map": "cuda",
                "quantization_config": quant_config_int8wo,
            },
            processor_kwargs={"max_pixels": 800 * 800},
        )
        model.eval()

        if compile:
            model = torch.compile(model, mode="reduce-overhead")

#        self.model = model
```
When run
```py
        embed = self.model.encode_document(
            args,
            ...
        )
        return embed.to(dtype=torch.half)
```

LK-maker-007:
Resolves #3908.

`encode` sets `device` to `self.device` when the caller passes none, then calls `self.to(device)`, so every call moves the model to the device it is already on. That is free for an ordinary tensor, but `nn.Module._apply` rebuilds each parameter, and for a torchao quantized weight the rebuild raises. `encode` runs under `@torch.inference_mode()`, which is what makes the rebuild fail:

```python
from torchao.quantization import quantize_, Int8WeightOnlyConfig

model = SentenceTransformer("sentence-transformers-testing/stsb-bert-tiny-safetensors", device="cpu")
quantize_(model[0].auto_model, Int8WeightOnlyConfig())
model.encode(["The weather is lovely today."])
# RuntimeError: _apply(): Couldn't swap Linear.weight
```

The move now happens only when a parameter or a buffer is somewhere else. `to()` moves buffers as well as parameters, so both are checked, otherwise a model whose BERT `position_ids`/`token_type_ids` buffers were left behind would not be moved. The requested device goes through `torch.empty` rather than `torch.device`, because `torch.device("cuda")` and `torch.device("cuda:0")` compare unequal while a parameter on a GPU always reports the index, so `device="cuda"` would otherwise move the model on every call.

I put the check in `BaseModel` so the three model classes share it, the same place the `device_map` handling from #3823 lives.

The tests cover both directions, for each model class: a model already in place is not moved, and one with a buffer or a module elsewhere still is. The first case runs over `None`, `"cpu"`, `"cpu:0"` and `torch.device("cpu")`, since the move is only skipped correctly when those agree.

The full `tests/sentence_transformer/`, `tests/cross_encoder/` and `tests/sparse_encoder/` suites pass, along with `tests/base/test_model.py` where these tests live (1491 passed together), and `ruff check` / `ruff format --check` are clean.

I don't have a GPU, so the `"cuda"` vs `cuda:0` path is reasoned from the `torch.device` comparison rather than run, and the reporter also had `torch.compile` in play, which I could not reproduce on CPU. Happy to move this or take a different approach if you'd prefer.


uditmahato:
I independently root-caused this yesterday and landed on the same conclusion, so consider this confirmation from a second environment (Windows, RTX 4060, torch 2.13.0+cu130, torchao 0.18.0, sentence-transformers 5.7.0): quantization alone reproduces the crash, and torch.compile turns out to be unrelated. The failure fires at the self.to(device) before any compiled forward runs. Repro matrix: quant plus compile crashes, quant only crashes, compile only passes, plain passes.

Two findings from that investigation that may be useful here:

1. This most likely also fixes #3457, which hits the identical code path through XLA tensors (their workaround was deleting the inference_mode decorator). Might be worth a "Fixes #3457" in the description so it closes together with this.

2. One nuance that explains the intermittency in the original report: .to('cuda') without an index short-circuits inside torchao and passes, while the device=None path resolves to cuda:0 and crashes. Your torch.empty(0, device=device).device normalization handles exactly that case, nice touch.

For what it's worth, I also verified on my side that a guard like this leaves plain-model embeddings bit identical, and that genuine cross-device moves of quantized models still fail deeper in torch/torchao (storage swap under inference mode), so keeping that out of scope seems right.

uditmahato:
This has gone conflicting since main moved: the device block in encode is around line 920 now rather than 641, and the cross encoder and sparse encoder ones moved with it. The underlying problem is still there on main, so the fix is still worth having.

Are you still planning to take this forward? Happy to leave it with you, just checking since it has been a couple of weeks.

If you would rather hand it off, I can open a rebased version with your commit kept as is so the authorship stays yours. Worth folding in #3457 at the same time, since the TPU report is the same code path and the same guard closes it.

LK-maker-007:
Merged main into this, and the conflict was only an import line.

While rebasing I found that `MultiVectorEncoder` needs the same change. It landed in #3794 on 11 August, four days after I opened this, and its `encode` defaults `device` to `self.device` and then calls `self.to(device)` exactly like the other three. So there are four call sites now, not three, and `tests/conftest.py` gains an `mve_bert_tiny_model` fixture following the pattern the other classes already use.

Two things about the guard that were not in the original description.

It is cheaper than the call it replaces. On bert-tiny, `self.to(device)` for a model already in place costs 204 us and the check costs 89 us, so each `encode` saves roughly 115 us rather than paying anything.

It can only ever skip a move, never add one. `_to_device` calls `self.to(device)` when some tensor is elsewhere, where before it was called unconditionally, so nothing that moved a model before stops moving it now. Models loaded with a `device_map` behave the same either way.

I also reworked the tests. They use `patch.object(model, "to", wraps=model.to)`, so the real move still happens and nothing needs suppressing, and the cases are split by what each one checks instead of crossing every model class with every spelling of the device. That took them from 22 to 17 while also covering the fourth class.

Reverting the guard fails 8 of them. Dropping `self.buffers()` from the check fails 3. Reverting the new `MultiVectorEncoder` line fails 1.

One detail worth recording: `CrossEncoder` resolves `device = str(self.device)` where the others keep a `torch.device`, so both spellings reach `_to_device`. `torch.empty` takes either, and a test covers both.

#3457 looks like the same code path, since the reporter passes an XLA device the model is already on, but I have no TPU to confirm it, so I have not added a closing keyword for it.

@uditmahato thanks for reproducing it on CUDA, the quant-only versus compile-only split was useful. I am carrying this one forward, so no need for a separate rebase.

tests/base/test_model.py
```diff
@@ -939,14 +940,83 @@ def test_device_map_controls_placement(model_class: type, model_id: str) -> None
     assert all(param.device.type == "cpu" for param in model.parameters())
 
 
-def test_device_argument_warns_when_device_map_present(caplog: pytest.LogCaptureFixture) -> None:
```
ebarkhordar, line 942: This hunk removes `test_device_argument_warns_when_device_map_present`, and I do not think that is deliberate. It is gone from the whole branch rather than moved:

```
$ git rev-parse HEAD
ec4b8ed256b5b72c741a65c9b0d52f357095199d
$ git grep -c test_device_argument_warns_when_device_map_present
(no match)
$ git show $(git merge-base main HEAD):tests/base/test_model.py | grep -c test_device_argument_warns_when_device_map_present
1
```

The code it pinned is untouched by this PR and still live, at `sentence_transformers/base/model.py:202-205`, which logs "Both `device` and `model_kwargs['device_map']` were provided ... the `device` argument is ignored". Of the three tests left in this file that mention `device_map` (`test_device_map_controls_placement`, `test_device_map_aligns_auxiliary_modules_to_backbone`, `test_device_map_ignored_for_prebuilt_modules`), none asserts that warning. So #3822 keeps its placement coverage and loses its warning coverage.

Putting the function back above the new `ENCODE_MODELS` block is the whole fix.

One thing I could not settle: I restored the test and ran it against your head, and it fails in `check_and_set_device_map` on current `transformers`. It fails the same way at the merge base, so that is my environment rather than your branch, and I am not claiming anything about whether it passes on CI.

ebarkhordar: Thanks, 6a7dfc48 is exactly it. The test is back at `tests/base/test_model.py:943` with both assertions intact, and nothing else in the file moved.

LK-maker-007: Sorry for the slow reply. You were right and I should have caught it myself: I cut the test while trimming this file from 22 tests to 17 and never checked which PR each one came from, so #3823's warning coverage went with it. On the part you could not settle, it passes. I ran the suite on a fork and test_device_argument_warns_when_device_map_present passed on all 8 jobs, 3.10 through 3.13 on ubuntu and windows, so your failure looks environmental. Two of the three remaining device_map tests are CUDA-gated and skip there, which is part of why this was easy to lose.
Thanks for catching it.

tomaarsen:
Merged via #3986 which takes this and went a step further via `accelerate`. Thanks all!

- Tom Aarsen

PR #3919, opened 2026-08-07
[fix] Return the add_transitive_closure flag from ParaphraseMiningEvaluator.get_config_dict

Kropiunig:
`ParaphraseMiningEvaluator.get_config_dict()` returns `self.add_transitive_closure`, but the `__init__` flag of that name is never stored on the instance, so the attribute lookup resolves to the `add_transitive_closure` staticmethod instead of the boolean. The model card builder json-dumps this config and falls back to `str()` on the resulting TypeError, so anyone training with this evaluator on the default path ends up with `'add_transitive_closure': <function ParaphraseMiningEvaluator.add_transitive_closure at 0x7f9c...>` rendered in their model card instead of `true`/`false`.

This stores the flag as `self._add_transitive_closure` in `__init__` and returns that from `get_config_dict()`, leaving the public staticmethod untouched. Added regression tests that `get_config_dict()` reports the actual boolean for both values and that the config is JSON-serializable, plus a test that the flag still expands the duplicates via the transitive closure. The config-dict tests fail on current `main` and pass with this change.


tomaarsen (approved):
Hello!

Apologies for the delayed response. I think this is a solid fix that cleans up a bit of a messy model card. I'll approve it.

- Tom Aarsen

PR #3920, opened 2026-08-08
[fix] Report worker crashes instead of hanging in multi-process encode

codechrl:
`_multi_process_worker` only catches `queue.Empty` around a blocking `input_queue.get()`, so any exception from `model.encode()` itself (bad input, OOM, etc.) kills the worker silently instead of being caught. `_multi_process`'s collection loop then does a fixed-count blocking `output_queue.get()` per submitted chunk, so it waits forever for the result that chunk will never produce.

Catches the encode failure in the worker, reports it on the results queue instead of dying, and keeps the worker alive for future chunks. The parent raises the first reported exception instead of treating it as a normal result.

Added two tests using a lightweight picklable fake model (no download needed): one exercises `_multi_process_worker` directly, the other drives `_multi_process` end-to-end with one poisoned chunk among several good ones and asserts the crash is raised, not hung on. Confirmed test-first: without the fix the run hangs indefinitely; with it, both pass.

---

*Disclosure: this change was prepared and assisted with AI assistance.*


sentence_transformers/sentence_transformer/model.py
```diff
@@ -915,7 +922,15 @@ def _multi_process_worker(
...
                 chunk_id, inputs, kwargs = input_queue.get()
-                embeddings = model.encode(inputs, device=target_device, **kwargs)
+                try:
+                    embeddings = model.encode(inputs, device=target_device, **kwargs)
+                except Exception as exc:
+                    # Report the failure instead of letting it kill this worker process silently:
+                    # the caller in _multi_process is waiting for exactly one result per submitted
+                    # chunk, so an unreported crash here would make it block forever.
+                    logger.exception(f"Chunk {chunk_id} failed to encode in a multi-process worker")
+                    results_queue.put([chunk_id, exc])
```
Jokasa7: [P1] An arbitrary `Exception` is not necessarily picklable, and `multiprocessing.Queue.put()` serializes in a background feeder thread after this call returns. On this exact head, a plain `RuntimeError` reached the parent, but an exception containing a lambda made the feeder fail with `AttributeError: Can't get local object ... <lambda>`; the result queue stayed empty while the worker remained alive. The parent's fixed-count `output_queue.get()` therefore still hangs forever. Please enqueue an explicitly serializable failure record (type/name, message, and formatted traceback) rather than the original exception object, and add a bounded regression with an exception carrying unpicklable state. The current tests only use `RuntimeError`, and the direct worker test also calls `get()` without a timeout.

Jokasa7 (commented):
Reporting failures through the result queue fixes serializable exceptions, but placing the original arbitrary exception object on a `multiprocessing.Queue` still leaves an indefinite-hang path. Queue serialization happens asynchronously in its feeder thread, after `put()` returns; an exception carrying any unpicklable state is dropped when that background pickle fails, while this worker remains alive and waits for more input.

I executed the exact `_multi_process_worker` method from this head (`model.py` SHA-256 `9cd56cd7845bc5dff7bc4cf1e2c8d133541475ac8a6e3f6b068fcb169db9d499`) under the same spawn/queue model. A plain `RuntimeError` was delivered. A custom `Exception` containing a lambda made the queue feeder report `AttributeError: Can't get local object ... <lambda>`; after a bounded wait, the output queue was empty and the worker was still alive. `_multi_process()` would therefore continue blocking for the missing fixed-count result. The new tests only raise `RuntimeError`, and the direct worker test itself uses an unbounded `output_queue.get()`.

Please send a primitive, explicitly serializable failure record (for example, exception module/name, message, and formatted traceback) and reconstruct an appropriate parent-side error, with a regression using an exception that has unpicklable state and a bounded assertion.

PR #3930, opened 2026-08-13
[fix] Retain max_pairs candidates in paraphrase mining

### LINKED ISSUE #3929: paraphrase_mining_embeddings keeps max_pairs - 1 candidates
### Problem

`paraphrase_mining_embeddings()` removes the smallest queued candidate as soon as the total number added is **equal to** `max_pairs`:

```python
pairs.put(candidate)
num_added += 1
if num_added >= max_pairs:
    pairs.get()
```

This keeps at most `max_pairs - 1` candidates. The boundary is visible with `max_pairs=1`, which returns no pair even when a valid pair exists.

### Minimal reproduction

```python
import torch
from sentence_transformers.util import paraphrase_mining_embeddings

embeddings = torch.tensor([[1.0, 0.0], [0.9, 0.1]])
pairs = paraphrase_mining_embeddings(
    embeddings,
    max_pairs=1,
    top_k=1,
    query_chunk_size=2,
    corpus_chunk_size=2,
)

assert len(pairs) == 1  # currently 0
```

The same off-by-one applies at the default: the priority queue retains at most 499,999 candidates when `max_pairs=500_000`.

### Expected behavior

The queue should evict only after it exceeds `max_pairs`, so the documented capacity is retained. `max_pairs=0` should still retain no candidates.

### Proposed fix

Change the eviction boundary from `>=` to `>`, and add boundary regressions for `max_pairs=0`, `1`, and `2`. The existing `min_score` pruning and duplicate-pair removal can remain unchanged.

I searched open and closed issues and pull requests for `paraphrase_mining`, `max_pairs`, and the boundary expression and found no existing report or implementation. I am preparing a focused patch.

FU-max-boop:
## Summary

Fixes #3929.

`paraphrase_mining_embeddings()` evicted the smallest candidate as soon as `num_added == max_pairs`. The priority queue therefore retained at most `max_pairs - 1` candidates; in the smallest useful case, `max_pairs=1` always returned an empty result.

This changes the eviction boundary to run only after the queue capacity is exceeded. The existing score threshold, top-k behavior, and unordered-pair deduplication are unchanged.

## Tests

- added deterministic boundary coverage for `max_pairs=0`, `1`, and `2`; the `1` and `2` cases fail before the fix
- `pytest tests/util/test_retrieval.py -q`: 14 passed, 1 skipped, 1 deselected
- repository pre-commit hooks on changed files: Ruff check, Ruff format, and typos passed
- `git diff --check`
- exact issue/PR duplicate search and upstream-sync check

AI assistance was used for boundary analysis, test design, and diff review; all changed lines and test outputs were manually verified.

### DIFF
--- a/sentence_transformers/util/retrieval.py
+++ b/sentence_transformers/util/retrieval.py
@@ -139,7 +139,7 @@ def paraphrase_mining_embeddings(
                         pairs.put((scores_top_k_values[query_itr][top_k_idx], i, j))
                         num_added += 1
 
-                        if num_added >= max_pairs:
+                        if num_added > max_pairs:
                             entry = pairs.get()
                             min_score = entry[0]
 
--- a/tests/util/test_retrieval.py
+++ b/tests/util/test_retrieval.py
@@ -5,7 +5,12 @@ import pytest
 import torch
 
 from sentence_transformers import SentenceTransformer
-from sentence_transformers.util.retrieval import community_detection, paraphrase_mining, semantic_search
+from sentence_transformers.util.retrieval import (
+    community_detection,
+    paraphrase_mining,
+    paraphrase_mining_embeddings,
+    semantic_search,
+)
 from sentence_transformers.util.similarity import pytorch_cos_sim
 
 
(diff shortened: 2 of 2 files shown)


tomaarsen (approved):
Looks good to me. Thanks for this fix! The test also looks fine. I'll merge this when the tests go green.

- Tom Aarsen

PR #3956, opened 2026-08-20
Fix duplicated/NaN results from semantic_search_usearch when corpus is smaller than top_k

mayuriphad:
## Summary

`semantic_search_usearch` returns duplicated `corpus_id` entries with `NaN` (or, after
rescoring, `-inf`) scores whenever `top_k` is larger than the corpus size.

## Root cause

Unlike FAISS, which pads a short result set with a sentinel index of `-1` (already handled
by `semantic_search_faiss`), usearch pads short result sets by repeating an arbitrary key
with a `NaN` distance to fill the requested `count`. `semantic_search_usearch` never
accounted for this padding, so it leaked straight into the returned results.

Repro (corpus of 3 documents, `top_k=10`):

```python
import numpy as np
from sentence_transformers.util.quantization import semantic_search_usearch

corpus = np.random.rand(3, 8).astype(np.float32)
query = np.random.rand(2, 8).astype(np.float32)
results, _ = semantic_search_usearch(query, corpus_embeddings=corpus, top_k=10, rescore=False)
print(results[0])
# [{'corpus_id': 2, 'score': 0.09...}, {'corpus_id': 1, 'score': 0.10...}, {'corpus_id': 0, 'score': 0.27...},
#  {'corpus_id': 18446744073709551615, 'score': nan}, ... (7 more phantom entries)
```

I couldn't use the same `-1`-sentinel trick as the FAISS fix because usearch keys are
unsigned (`uint64`): casting `-1` into a `uint64` array wraps around to
`18446744073709551615`, which then also silently fails an `!= -1` comparison against the
Python int `-1` (mixed signed/unsigned comparison quirk in NumPy). Instead, the fix reads
`matches.counts` (the number of real matches per query, as returned by usearch) to build a
parallel boolean "validity" array, and threads that array through the rescoring reorder the
same way `indices`/`scores` already are, then filters by it when building the output.

## Changes

- `sentence_transformers/util/quantization.py`: `semantic_search_usearch` now derives a
  validity mask from `matches.counts`, disqualifies padded rescoring candidates via
  `-inf`, and drops non-valid entries from the final output (with and without rescoring).
- `tests/util/test_quantization.py`: added `test_semantic_search_usearch_drops_padded_indices`
  (parametrized over `corpus_precision` and `rescore`) and
  `test_semantic_search_usearch_returns_top_k_when_corpus_is_large_enough`, mirroring the
  existing FAISS padding regression tests.

## Test plan

- `python -m pytest tests/util/test_quantization.py -k usearch -q` — 6 passed
  (fails with 4/6 on the pre-fix code, confirmed by stashing the source change)
- `python -m pytest tests/util/test_quantization.py -q` — 35 passed
- `ruff check` / `ruff format --check` on the changed files — clean


sentence_transformers/util/quantization.py
```diff
@@ -352,9 +352,21 @@ def semantic_search_usearch(
     if indices.ndim < 2:
         indices = np.atleast_2d(indices)
 
+    # Unlike FAISS, usearch does not pad short result sets with a sentinel index: when a query has fewer
+    # than `k` valid matches (e.g. the corpus itself has fewer than `k` entries), the remaining slots repeat
+    # an arbitrary key with a NaN distance. `matches.counts` holds the number of real matches per query, so
+    # anything at or beyond that count is padding. Keys are unsigned (uint64), so a -1 sentinel can't be used
+    # to mark them (it wraps around to a huge positive value instead); track validity in a parallel boolean
+    # array instead, threading it through the rescoring reorder below the same way as indices/scores.
+    counts = np.atleast_1d(matches.counts)
```
Jokasa7: [P1] `matches.counts` is only available on USEARCH's `BatchMatches`. For a normal single-query array with shape `(1, dim)`, USEARCH 2.26.1 returns `Matches`, so this line now raises `AttributeError` before any result is produced—even when `top_k` is smaller than the corpus and there is no padding to remove. I ran the exact base/head functions with a three-vector corpus and `top_k=2`: base returned two neighbors, head failed here, while the same head succeeded for two query rows. Please handle both result types and add a one-query test; the current fixture always contains two queries and misses this branch.

Jokasa7 (commented):
The padding fix works for a multi-query batch, but the new `matches.counts` access breaks the single-query shape that this function already normalizes. With USEARCH 2.26.1, `Index.search()` returns `Matches` for a `(1, dim)` query and `BatchMatches` for multiple queries; only `BatchMatches` exposes `counts`.

I executed the exact base and head implementations from their GitHub source against the same three-vector float32 corpus with `top_k=2` and `rescore=False`, so this case did not require padding. The base returned one row with two results. This head raised `AttributeError: 'Matches' object has no attribute 'counts'`; the same head succeeded when given two query rows. The added tests always use the two-row `QUERIES` fixture, so they do not exercise this API branch.

Please normalize the real-match count for both `Matches` and `BatchMatches` (or otherwise avoid reading a batch-only attribute), and add a one-row regression alongside the current padding cases.

PR #3405, opened 2025-06-30
[`fix`] Use transformers Peft integration instead of manual get_peft_model call

tomaarsen:
Resolves #3402

Hello!

## Pull Request overview
* Use transformers Peft integration instead of manual get_peft_model call

## Details
This resolves the issue from #3402. In short, I actually fixed this via https://github.com/arthurbr11/sentence-transformers/pull/47, which ended up really changing up the `_load_from_checkpoint` on the trainers. However, it wasn't until #3403 by @bg-l2norm that I discovered the discrepancy between 1) loading a model with an Adapter and 2) adding an adapter with e.g. `add_adapter`. In short, the former calls `get_peft_model` and turns the base model into a `PeftModelFor...`, whereas the latter keeps the model as a e.g. `BertModel`, but with LoRA layers.

@bg-l2norm 's PR updates the latter so that both cases use `PeftModelFor...`. However, I think my preference for now is to use the `transformers` implementation as much as possible. That's why I've made this PR - to align them so that both use e.g. `BertModel` but with e.g. LoRA layers.

I tested a training script with both #3403 and this PR, and both work, but my preference is this PR. Thank you @bg-l2norm and @mjstel for helping me get this resolved nicely!

- Tom Aarsen

PR #3957, opened 2026-08-20
[fix] Support the adapter methods when the underlying model is already a PeftModel

### LINKED ISSUE #3247: Loading a PEFT model uses `peft.peft_model.PeftModelForFeatureExtraction`; incompatible with ST PEFT Methods

dylanpulver:
Fixes #3247

`has_peft_compatible_model()` gates on `isinstance(model, transformers.integrations.peft.PeftAdapterMixin)`, which is `False` once the module is wrapped in a `peft.PeftModel`, so all nine adapter methods raise `PEFT methods are only supported for Sentence Transformer models that use the Transformer module`.

One correction to the issue: its snippet no longer reproduces on `main`, since #3405 moved Hub loading onto the transformers integration. The live route is wrapping the module yourself, `model[0].model = get_peft_model(model[0].model, ...)`, which for prompt learning methods is the only route because transformers' `add_adapter` refuses them. `Transformer.__init__` already unwraps a `PeftModel` and `forward` already extends the mask for `is_prompt_learning`, so only the mixin is broken.

## The fix

Accept a `peft.PeftModel`, then translate the calls. Delegating blindly gives wrong answers: `PeftModel.__getattr__` forwards down to the inner model, so `disable_adapters()` reaches the inner transformers mixin and raises `No adapter loaded` while every `hasattr` check passes. Translated are argument order on `add_adapter`, the required positional on `load_adapter`, `active_adapters`/`active_adapter` being properties rather than methods, `get_adapter_state_dict` to the free `get_peft_model_state_dict`, and the enable/disable pair to `enable_adapter_layers`/`disable_adapter_layers` (prompt learning has no adapter layers, so that combination now raises a `ValueError` saying so).

The check still rejects what it rejected before: `transformers_model` is `None` for a `WordEmbeddings` stack, and `onnx`/`openvino` return optimum models. Neither is a `PeftModel`.

**One question.** transformers' `add_adapter` activates what it adds, `PeftModel.add_adapter` does not. I kept each side's semantics rather than silently switching a user's active adapter; happy to unify if you prefer.

## Testing

`test_adapter_methods_on_peft_wrapped_model` and `test_adapter_methods_on_prompt_learning_model` in `tests/sentence_transformer/test_model.py`, both failing on `main`. Adapter/PEFT selection across that file and `tests/multi_vector_encoder/test_model.py`: 5 passed, 2 skipped (both pre-existing). `pre-commit` clean. macOS CPU, Python 3.12.13, `peft` 0.20.0, `transformers` 5.15.1, `torch` 2.13.0.

Not verified: the full suite (killed at eighteen minutes stalled on unauthenticated Hub fetches, so CI is the first full run); other Python/OS/`peft` versions, and `enable_adapter_layers` and `get_peft_model_state_dict` are internals-adjacent enough that older `peft` may need a gate; prefix and P-tuning, and tuners other than LoRA; training through `SentenceTransformerTrainer`; and the `onnx`/`openvino` backends, which I reasoned about but did not install.

### DIFF
--- a/sentence_transformers/base/peft_mixin.py
+++ b/sentence_transformers/base/peft_mixin.py
@@ -77,9 +77,49 @@ def _peft_model_disable_adapters(model: PeftModel) -> None:
     _peft_model_toggle_adapters(model, enable=False)
 
 
+def _single_adapter_name(adapter_name: Any, method_name: str) -> str:
+    """The transformers mixin accepts a list of adapter names, ``PeftModel`` addresses exactly one."""
+    if isinstance(adapter_name, str):
+        return adapter_name
+    if isinstance(adapter_name, (list, tuple)) and len(adapter_name) == 1 and isinstance(adapter_name[0], str):
+        return adapter_name[0]
+    raise ValueError(
+        f"`{method_name}` accepts a list of adapter names on a plain transformers model, but the underlying model "
+        f"is a peft.PeftModel, which addresses a single adapter at a time. Pass one adapter name instead, calling "
+        f"`{method_name}` once per adapter if needed. Got {adapter_name!r}."
+    )
+
+
+def _peft_model_set_adapter(model: PeftModel, adapter_name: str | list[str], **kwargs) -> None:
+    """``PeftModel`` activates exactly one adapter, so a list of several names cannot be applied."""
+    return model.set_adapter(_single_adapter_name(adapter_name, "set_adapter"), **kwargs)
+
+
+def _peft_model_delete_adapter(model: PeftModel, adapter_names: str | list[str] | None = None, **kwargs) -> None:
+    """``PeftModel.delete_adapter`` takes a single ``adapter_name``, not the transformers ``adapter_names``.
+
+    ``PeftModel.delete_adapter`` forwards to ``self.base_model.delete_adapter(adapter_name=...)``. For the tuner
+    based methods ``base_model`` is the tuner, which accepts that keyword, but for prompt learning ``base_model``
+    is the raw transformers model, whose own mixin takes ``adapter_names``, so the forwarded call raises a
+    ``TypeError``. ``peft`` implements no deletion path for prompt learning at all, so reject it with an
+    explanation rather than letting that ``TypeError`` surface.
+    """
+    if adapter_names is None:
+        adapter_names = kwargs.pop("adapter_name", None)
+    adapter_name = _single_adapter_name(adapter_names, "delete_adapter")
+    config = model.peft_config.get(adapter_name)
+    if config is not None and config.is_prompt_learning:
+        raise ValueError(
+            f"Adapters of type {type(config).__name__} cannot be deleted from the underlying peft.PeftModel. "
+            "peft only implements deletion for the tuner based methods; for prompt learning methods such as "
+            "prompt tuning, prefix tuning and P-tuning it forwards the call to the base transformers model, "
+            "which does not hold the prompt adapter."
+        )
+    return model.delete_adapter(adapter_name, **kwargs)
+
+
 # ``peft.PeftModel`` and ``transformers.integrations.peft.PeftAdapterMixin`` expose overlapping but
-# differently shaped APIs. Methods listed here are translated to their ``peft`` equivalent; the rest
-# (``set_adapter``, ``delete_adapter``) match closely enough to be called directly.
+# differently shaped APIs. Methods listed here are translated to their ``peft`` equivalent.
 PEFT_MODEL_TRANSLATIONS = {
     "load_adapter": _peft_model_load_adapter,
     "add_adapter": _peft_model_add_adapter,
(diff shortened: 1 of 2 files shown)


sentence_transformers/base/peft_mixin.py
```diff
@@ -1,8 +1,94 @@
...
+
+
+def _peft_model_disable_adapters(model: PeftModel) -> None:
+    """``PeftModel`` toggles adapter layers on the tuner rather than on the model itself."""
+    _peft_model_toggle_adapters(model, enable=False)
+
+
+# ``peft.PeftModel`` and ``transformers.integrations.peft.PeftAdapterMixin`` expose overlapping but
+# differently shaped APIs. Methods listed here are translated to their ``peft`` equivalent; the rest
+# (``set_adapter``, ``delete_adapter``) match closely enough to be called directly.
```
Jokasa7: **[P2] Translate deletion before falling through to PeftModel**

The direct-call assumption is false for the prompt-learning path added here. With the versions used in this PR, PEFT 0.20.0's `PeftModel.delete_adapter` calls `self.base_model.delete_adapter(adapter_name=adapter_name)`. A prompt-learning `PeftModel` keeps the raw Transformers model as `base_model`, while Transformers 5.15.1 exposes `delete_adapter(adapter_names=...)`; consequently, `model.delete_adapter("default")` raises `TypeError: delete_adapter() got an unexpected keyword argument 'adapter_name'`. I reproduced this at exact head `8a15bf2` by executing the exact wrapper and dependency methods; the new prompt-learning test does not exercise deletion. The same fall-through also loses the documented list-valued `set_adapter`/`delete_adapter` contract. Please translate deletion on the wrapper and add a prompt-deletion regression test, while handling or clearly rejecting list-valued calls.

Jokasa7 (commented):
I verified the new wrapped-model dispatch against PEFT 0.20.0 and Transformers 5.15.1. One remaining interface mismatch still breaks adapter deletion for the prompt-learning route; details inline.

sentence_transformers/base/peft_mixin.py
dylanpulver: Both halves hold up. Fixed in 33a9e1e2.

I reproduced the deletion failure at 8a15bf2 against the versions you named, peft 0.20.0 and transformers 5.15.1:

```
  File "sentence_transformers/base/peft_mixin.py", line 114, in wrapper
    return method(*args, **kwargs)
  File "peft/peft_model.py", line 1193, in delete_adapter
    self.base_model.delete_adapter(adapter_name=adapter_name)
TypeError: PeftAdapterMixin.delete_adapter() got an unexpected keyword argument 'adapter_name'
```

The mechanism is as you describe. `PeftModel.__init__` assigns `self.base_model = model` directly on the prompt learning branch, so `base_model` is the raw `BertModel`, whose signature is `delete_adapter(adapter_names: list[str] | str)` against peft's forwarded `adapter_name=`. For the tuner based methods `base_model` is the tuner, which does take `adapter_name`, which is why the LoRA deletion in the existing test passes and the gap went unnoticed.

On the list-valued half, the divergence is real too. A plain transformers model accepts `set_adapter(["a", "b"])` and `delete_adapter(["a", "b"])`; through the fall-through both raise `TypeError: unhashable type: 'list'`. I would not call it a documented contract in this repo, though. Nothing in the docs or tests exercises list-valued calls, and these docstrings only link out to the transformers signature rather than restating it. The behavioural difference is worth closing regardless.

What I changed. `set_adapter` and `delete_adapter` are now translated rather than called directly. Both accept a single-element list and reject longer ones with a message saying `PeftModel` addresses one adapter at a time, since `PeftModel.set_adapter` is documented as one active adapter and cannot be made to honor a list.

For prompt learning deletion I rejected rather than implemented, and I would like your read on that. peft has no prompt-learning-aware deletion anywhere: `PeftModel.delete_adapter` is the only entry point and it unconditionally delegates to `base_model`. Making it work means popping `peft_config`, `prompt_encoder` and `prompt_tokens` by hand, which is inventing semantics upstream has not defined (deleting the only or the active prompt adapter leaves the model in a state peft does not specify) and reaching into peft internals from a library that does not pin peft at all. So it now raises a ValueError explaining that peft cannot delete prompt learning adapters, mirroring the existing message for enable/disable on prompt learning. Happy to file this upstream, or to implement the deletion here if you would rather this PR carry it.

Tests: added prompt-learning deletion and the list-valued cases. Reverting the source hunk fails both new tests with the `TypeError`s above; with it, `tests/sentence_transformer/test_model.py` goes 509 passed / 9 skipped on main to 511 / 9 on the branch, no other changes. ruff 0.15.6 and typos 1.45.0 clean.

Issue #3939, opened 2026-08-16
`MultiVectorInformationRetrievalEvaluator` keeps the first model’s similarity function when reused


eSVeeF:
### Description

I noticed that `MultiVectorInformationRetrievalEvaluator` appears to keep using the similarity function from the first model it evaluates.

I am evaluating several MultiVectorEncoder checkpoints with the same IR dataset and evaluator. One checkpoint uses the default `maxsim`, while another uses `meanmaxsim`. I expected the evaluator to resolve the scoring function from the model passed to each call, since the constructor documentation says the default scoring is resolved at call time.

However, after evaluating the first model, the evaluator keeps the original scorer and metric name for subsequent models.

### Reproduction

```python
from sentence_transformers import MultiVectorEncoder
from sentence_transformers.multi_vector_encoder.evaluation import (
    MultiVectorInformationRetrievalEvaluator,
)

model = MultiVectorEncoder(
    "sentence-transformers-testing/stsb-bert-tiny-safetensors"
)

evaluator = MultiVectorInformationRetrievalEvaluator(
    queries={"q0": "What is the capital of France?"},
    corpus={
        "d0": "Paris is the capital of France.",
        "d1": "Berlin is the capital of Germany.",
    },
    relevant_docs={"q0": {"d0"}},
    name="reuse",
    write_csv=False,
)

# First evaluation uses maxsim
first_results = evaluator(model)
print(first_results.keys())
# Contains keys such as:
# "reuse_maxsim_ndcg@10"

# Change the model's configured similarity function
model.similarity_fn_name = "meanmaxsim"

# Reuse the same evaluator
second_results = evaluator(model)
print(second_results.keys())
# Still contains:
# "reuse_maxsim_ndcg@10"
#
# Expected:
# "reuse_meanmaxsim_ndcg@10"
```

After the first call, the evaluator also contains the first model-derived scorer in `evaluator.score_functions`, so changing `model.similarity_fn_name` does not affect later calls.

### Expected behavior

When `score_functions` is not explicitly provided, each call should derive the scorer from the model passed to that call. Reusing an evaluator with a model configured for `meanmaxsim` should produce `meanmaxsim` metrics and use the corresponding scoring function.

Explicitly supplied `score_functions` should continue to remain fixed.

### Why this matters

[...]

PR #3962, opened 2026-08-24
Fix similarity function latch when InformationRetrievalEvaluator is reused across models

shoemoney:
`InformationRetrievalEvaluator.__call__` derives `score_functions` from the first evaluated model and keeps it permanently (`sentence_transformers/sentence_transformer/evaluation/information_retrieval.py:239` on main). A reused evaluator therefore scores every later model with the first model's captured `model.similarity` and labels the metrics, CSV columns, and model card metadata with the first model's `similarity_fn_name`. With a cosine model followed by a dot model, the second call still returns `..._cosine_ndcg@10` computed by the first model's cosine scorer; for unnormalized embeddings dot and cosine rank differently, so NDCG/MRR/Recall are numerically wrong for the later model. Any evaluator reuse across checkpoints hits this, including `SparseInformationRetrievalEvaluator` and `NanoBEIREvaluator`'s reused per-subset evaluators.

This is the same latch reported for the multi-vector subclass in #3939, whose in-flight fix #3940 covers only `multi_vector_encoder/evaluation/information_retrieval.py`. This PR mirrors that approach in the base class:

- A model-derived scorer is temporary: it is reset at the next call's entry, so each call resolves, scores, and labels with the current model's similarity function, and `primary_metric` is re-selected accordingly.
- CSV headers are appended only when the derived score name is not already registered, so repeated calls no longer duplicate columns.
- Explicitly supplied `score_functions` remain permanent, unchanged behavior.

`NanoBEIREvaluator.__call__` carries an identical copy of the derivation for its aggregate labeling (`nano_beir.py:328`). Fixing only the base class would make its reuse path crash instead of mislabel: subset results would come back keyed by the new similarity name while the aggregate reporting loop at `nano_beir.py:397` still looks up the latched one and raises KeyError. Both sites are fixed with the same pattern.

Deliberately not touched: the multi-vector subclass's own copy of the latch, which #3940 already fixes; this PR stays off that file so the two merge independently. The subclass assigns `self.score_functions` before delegating to the base `__call__`, and the reset triggers only for scorers the base class itself derived, so that path behaves identically before and after #3940.

Tests cover reuse across models with different `similarity_fn_name` (base evaluator and NanoBEIR via a stubbed dataset loader), header dedup on repeated calls, and permanence of explicit `score_functions`. Without the source change, the two reuse tests fail; with it, `tests/sentence_transformer/evaluation/` plus `tests/multi_vector_encoder/test_evaluators.py` pass (47 passed, 1 skipped).

### DIFF
--- a/sentence_transformers/sentence_transformer/evaluation/information_retrieval.py
+++ b/sentence_transformers/sentence_transformer/evaluation/information_retrieval.py
@@ -183,6 +183,10 @@ class InformationRetrievalEvaluator(BaseEvaluator):
         self.write_csv = write_csv
         self.score_functions = score_functions
         self.score_function_names = sorted(list(self.score_functions.keys())) if score_functions else []
+        # Whether score_functions currently holds a scorer derived from an evaluated model rather
+        # than one supplied by the user. Derived scorers are temporary and reset on the next call
+        # (see __call__), so a reused evaluator follows each model's own similarity function.
+        self._model_derived_score_functions = False
         self.main_score_function = SimilarityFunction(main_score_function) if main_score_function else None
         self.truncate_dim = truncate_dim
 
@@ -236,10 +240,21 @@ class InformationRetrievalEvaluator(BaseEvaluator):
 
         logger.info(f"Information Retrieval Evaluation of the model on the {self.name} dataset{out_txt}:")
 
+        if self._model_derived_score_functions:
+            # The scorer derived from a previously evaluated model is temporary: reset it so this
+            # call resolves and labels the current model's similarity function instead of scoring
+            # every later model with the first model's captured similarity.
+            self.score_functions = None
+            self.score_function_names = []
+            self.primary_metric = None
+            self._model_derived_score_functions = False
+
         if self.score_functions is None:
             self.score_functions = {model.similarity_fn_name: model.similarity}
             self.score_function_names = [model.similarity_fn_name]
-            self._append_csv_headers(self.score_function_names)
+            self._model_derived_score_functions = True
+            if not any(header.startswith(f"{model.similarity_fn_name}-") for header in self.csv_headers):
+                self._append_csv_headers(self.score_function_names)
 
         scores = self.compute_all_metrics(model, output_path=output_path, *args, **kwargs)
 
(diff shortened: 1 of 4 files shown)


sentence_transformers/sentence_transformer/evaluation/information_retrieval.py
```diff
@@ -236,10 +240,21 @@ def __call__(
...
+            self.score_function_names = []
+            self.primary_metric = None
+            self._model_derived_score_functions = False
+
         if self.score_functions is None:
             self.score_functions = {model.similarity_fn_name: model.similarity}
             self.score_function_names = [model.similarity_fn_name]
-            self._append_csv_headers(self.score_function_names)
+            self._model_derived_score_functions = True
+            if not any(header.startswith(f"{model.similarity_fn_name}-") for header in self.csv_headers):
```
Jokasa7, line 256: [P1] Extending `self.csv_headers` here does not update a CSV that the first model call already created. On this exact head, a cosine call wrote `epoch,steps,cosine-NDCG@10` plus `1,1,0.1`; a subsequent dot call appended only `2,2,0.9`, while the persisted header remained unchanged even though the in-memory headers now also contained `dot-NDCG@10`. The dot result is therefore silently labeled as cosine in the output artifact. Please make the file schema stable across scorer changes (and mirror the fix in `NanoBEIREvaluator`) and add a two-model CSV readback regression.

Jokasa7 (commented):
The model-specific scorer reset fixes the returned metrics, but the persisted CSV schema does not follow the same transition. I executed the exact `_append_csv_headers()` and `__call__()` methods from this head twice against one output file: first with `cosine` producing `0.1`, then with `dot` producing `0.9`. After the second call, `self.csv_headers` contained both `cosine-NDCG@10` and `dot-NDCG@10`, but the file header was still only `epoch,steps,cosine-NDCG@10`; the appended dot row was `2,2,0.9`. A reader therefore silently interprets the dot score as a cosine score.

This is the same reuse scenario the PR is intended to repair, and `NanoBEIREvaluator` has the identical append-only header branch. The output needs a stable on-disk schema—such as one file per score function, a score-function field with generic metric columns, or a migrated header with prior rows backfilled—rather than only extending the in-memory list. A regression that performs two calls with different model similarities and reads the resulting CSV would catch both implementations.

PR #3986, opened 2026-09-07
Fix inference device placement for quantized and dispatched models

tomaarsen:
Supersedes #3918

Hello!

Heads up, this PR was AI-generated and human-reviewed.

## Pull Request overview

* Fix redundant device moves and inference-mode parameter conversion during encoding
* Preserve device maps and Accelerate placement across all four model classes
* Align auxiliary modules and respect explicit device moves

## Details

This builds on Singaraj B's work in #3918 to fix the torchao encoding failure reported in #3908. Encoding currently calls `model.to(device)` even when every parameter and buffer is already there. With certain quantized weights, that redundant call fails with `_apply(): Couldn't swap Linear.weight`. Encoding can also override an existing device map, and offloaded parameters can make `model.device` report `meta` instead of the execution device.

I've centralized inference placement in `_resolve_inference_device`, shared by `SentenceTransformer`, `CrossEncoder`, `SparseEncoder`, and `MultiVectorEncoder`. Ordinary models still move to the requested device, including auxiliary modules and buffers that need aligning. The helper skips the move when everything is already in place and recognizes equivalent device spellings such as `cuda` and `cuda:0`. Necessary moves run outside inference mode so encoding does not leave model parameters or buffers as inference tensors.

For models placed with a device map or Accelerate hooks, I've kept placement under that mechanism's control. Hook detection searches the live module hierarchy, including composed hooks and non-Transformers modules such as `StaticEmbedding`. The execution device comes from the placement hook when present. Auxiliary modules are aligned during construction and again before inference, so adding a CPU `Dense` layer to a model running on CUDA still works. Alignment preserves loaded backbones and hooked submodules, including those nested inside a `Router`, and handles buffers registered on the outer model.

An explicit encoding device is now ignored with a warning when a device map or live placement hooks control placement. For example, a model loaded with `device_map={"": "cpu"}` stays on CPU when calling `encode(..., device="cuda")`. Without live placement hooks, an explicit `model.to(device)`, `model.cpu()`, or `model.cuda()` overrides the saved map policy and restores ordinary inference placement. Dtype-only conversions keep that policy. I've also guarded recursive moves and casts of offloaded models, because ST's `nn.Module._apply` recursion bypasses the guards Accelerate attaches to the backbone's movement methods.

Multiprocessing rejects models with live placement hooks before starting workers. Models loaded with `device_map="auto"` that fit on one device and have no placement hooks remain supported. Starting a pool takes over placement so each worker can move its copy to its target device.

The latest validation passed all 225 selected base-model tests, with 2 slow multiprocessing tests deselected, using an RTX 3090 with PyTorch 2.11.0+cu128, Transformers 5.4.0, Accelerate 1.13.0, and torchao 0.18.0. Coverage includes CPU/GPU round trips, quantized weights, CPU/disk offloading, static embeddings, appended modules, root buffers, composed hooks, and repeated inference. Ruff and the commit hooks passed. Multiple-GPU placement remains untested.

I've also reduced repeated hierarchy scans and device lookups. In the tiny-model benchmarks, delegated device resolution took roughly 40 to 70 microseconds per inference call after these optimizations. Full encoding timings were mixed, so I would not claim a consistent end-to-end speedup. These changes primarily address correctness and compatibility.

- Tom Aarsen


NewUserHa:
What is the reason to check that many arrtibutes in `_accelerate_placement_hook` and `device` in every call of `encode()` in the PR, if `encode()` is meant to be frequently called by users?

Can move the `model.to(gpu)` out from `encode()` and into model loading process, for higher performance and less cpu usages?

tomaarsen:
Thanks for looking at this @NewUserHa. I did originally try to fully remove the device management from `encode` and instead keep it in `__init__`. However, this is problematic if users add a module (e.g. a `Dense`) after the larger model has been constructed, and the initial model placement on `__init__` might contain "meta" tensors that need to be filled in.

I benchmarked the cost when I made the PR, and the cost was quite minimal. I'll try to see if I can perhaps e.g. cache `_accelerate_placement_hook` once found though. And perhaps `_placement_is_delegated` as well.

But overall, the cost of this PR is likely lower, as it skips the `model.to(...)` when tensors are already on the right device. I think it's also faster if you use `SentenceTransformer(..., model_kwargs={"device_map": "auto"})`.

- Tom Aarsen

PR #3987, opened 2026-09-07
docs: note thread-budget parity when comparing backends on CPU

ErnisBadmae:
The Speeding up Inference page compares the torch, ONNX and OpenVINO backends and reports CPU benchmarks, but does not mention thread configuration.

This is easy to get wrong in a way that silently invalidates a comparison: torch.set_num_threads() does not affect ONNX Runtime, and ONNX Runtime with the default intra_op_num_threads = 0 picks a thread count on its own. A reader who benchmarks the backends on CPU may end up measuring two different thread budgets rather than two backends.

Measured on intfloat/multilingual-e5-large on CPU: with the thread budget left at defaults, ONNX looked ~1.14x faster on a batch of 32 long texts; after setting the same budget on both sides (12 threads) the batch speedup dropped to ~1.03x. The single-short-query speedup survived and was slightly larger (1.49x -> 1.54x), which is what you would expect - batched work is dominated by matrix multiplications both backends dispatch to the same kernels, while a short query is dominated by per-op overhead that ONNX removes.

Adds a note with a snippet that equalizes the budget on both sides, and documents session_options in the list of notable model_kwargs. No API change.


tomaarsen:
Hello!

Documenting `session_options` seems useful, but I don't agree that different thread counts necessarily invalidate a backend comparison. The goal of this page is to help users get better inference performance on their hardware, and the best thread count can differ between backends.

I've substantially refreshed the benchmarks since this PR was opened. The local CPU results now use settings selected independently per model and backend from 8 and 20 threads, with ONNX Runtime's thread counts explicitly configured. This compares the best tested throughput on the same hardware. Requiring identical thread counts would answer a different question, and could leave one backend running with a less suitable configuration. I do agree that we should document the settings used, so I've added a short explanation to the benchmark details.

I've also reduced the `session_options` addition to a brief explanation of how it can help tune CPU inference performance. I removed the longer example and the general explanation of where ONNX helps most. The measurements on one model don't establish that behavior in general, and PyTorch and ONNX Runtime don't necessarily use the same matrix multiplication kernels.

I'll let Copilot have another look, and I'll merge it afterwards.

- Tom Aarsen

Issue #3864, opened 2026-07-16
Multiple videos and metadata process in one time


NewUserHa:
```py
model = SentenceTransformer(
            "Qwen/Qwen3-VL-Embedding-2B",
            device="cuda",
            cache_folder=r"...",
            local_files_only=False,
            model_kwargs={
                "torch_dtype": torch.bfloat16,
                "low_cpu_mem_usage": True,
                "attn_implementation": "sdpa",
                "quantization_config": quant_config_int4,
                "device_map": "auto",
            },
            processor_kwargs={"max_pixels": 1200 * 1200},
        )
emb = model.encode_document(
    {"video": [r"F:\40.jpg", r"F:\40.jpg", r"F:\40.jpg"]},
    processing_kwargs={
        "video": {
            "size": {
                "longest_edge": 720 * 720,
                "shortest_edge": 64 * 64,
            },
            "do_sample_frames":False,
            "video_metadata": {"total_num_frames":3,'fps': 15,"frames_indices": [0, 5, 10],},
        },
    },
)
```

Currently when process video and the input is pictures or a 4D tensor, the video_metadata has be passed to the model. But it can only be passed in processing_kwargs for batch-wide instead for each video. Therefore `encode_document` can process only 1 video each run:
When pass the `{"video_metadata": ...}` along with the video dictory to be like `{"video": ..., {"video_metadata": ...}}`, it throws
```
ValueError: Multimodal dict input contains unrecognized modality keys: {'video_metadata', 'do_sample_frames'}. Expected keys from: ['audio', 'image', 'text', 'video']
```

Is it able to process multiple videos that is preloaded and sampled into 4d tensor ahead in senten `sentence-transformers`, or is there a simpler solution? So it can max useage if the single video is small.
Or is it a special case that it is only qwe3-vl needs the `fps` and `frames_indices` metadata?

Thanks

tomaarsen:
Hello!

Good news: you can already do this today, it's just not via a `video_metadata` key next to `video`. Per-sample video metadata is attached by passing each video as a dict with `"array"` and `"video_metadata"` keys (the same shape audio uses with `"array"` + `"sampling_rate"`). Because it's per-sample, you can hand a whole list to one `encode_document` call, each video with its own metadata:

```python
emb = model.encode_document(
    [
[...]

PR #3988, opened 2026-09-07
docs: explain per-video metadata for pre-sampled batches

Con-Benksl:
Related to #3864.

The usage page lists accepted video input shapes but does not explain how to batch pre-sampled videos with different metadata. Add an `encode_document` example with per-video wrappers, explain the original frame rate/count/indices, and distinguish per-video metadata from batch-wide processing options. Include the nested text/video form and the file-path alternative.

Validation:
- `pytest tests/base/test_modality.py -q`: 216 passed, 2 skipped (optional TorchCodec dependency unavailable).
- Checked the example's input construction with generated images, metadata alignment and timestamps, nested text/video inputs, and local Qwen3-VL video preprocessing. Model weights and embedding inference were not used.
- Sphinx HTML build succeeded; its 75 warnings are identical to an unmodified-main build in the same environment.
- Pre-commit passed for the changed RST file; independent review found no blocking issues.

AI assistance: OpenAI Codex prepared and validated this documentation change and performed the independent review.

### DIFF
--- a/docs/sentence_transformer/usage/usage.rst
+++ b/docs/sentence_transformer/usage/usage.rst
@@ -138,6 +138,54 @@ These methods accept all the same input types as :meth:`~sentence_transformers.s
    # tensor([[0.3907, 0.1490],
    #         [0.1235, 0.4872]])
 
+Encoding pre-sampled videos
+---------------------------
+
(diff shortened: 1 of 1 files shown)


tomaarsen:
Hello!

Thanks for opening this! This PR made me realise that there's still a bit of confusion about how to inference for the various newly supported modalities. I think that deserves its own separate documentation page rather than just a short description/list of "you can pass tensors with a video_metadata" etc. like I was doing previously. So, I opened #4036, which also covers per-video metadata as that's a realistic edge case. 

I'll close this PR once that one gets merged.

- Tom Aarsen

tomaarsen:
Closed as I've just merged #4036. Thanks again 🤗 

- Tom Aarsen

PR #3989, opened 2026-09-07
Make reranking evaluator MRR independent of candidate order

ahsan-h-shihab:
Both reranking evaluators take the rank of the first relevant candidate from a single argsort, so equally scored candidates make the reported MRR depend on the order they were passed in. Average the reciprocal rank over the orderings a tied group allows instead, as McSherry and Najork (2008) describe, matching the tie handling ndcg_score already applies. Scores without ties are unchanged.

### DIFF
--- a/sentence_transformers/base/evaluation/metrics.py
+++ b/sentence_transformers/base/evaluation/metrics.py
@@ -1,0 +1,58 @@
+from __future__ import annotations
+
+import numpy as np
+
+
+def tie_aware_reciprocal_rank(is_relevant, scores, at_k: int) -> float:
+    """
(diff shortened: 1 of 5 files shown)


tomaarsen:
Hello!

Thanks for opening. This is indeed an edge case, but I think a separate metric implementation is a bit overkill for this. I'd be fine with a small fix of a few lines, but correctly averaging over ties seems to require quite a bit more than that, especially with multiple positives and the `at_k` cutoff.

I don't think the benefit justifies the extra code and maintenance here. I'd rather keep the current behavior, so I'm afraid I'll be closing this one.

- Tom Aarsen

PR #3990, opened 2026-09-08
Fix implicit Optional annotations

VenishPaneliya:
Seven parameters annotate a concrete type but default to `None`, which PEP 484 prohibits. In every case the body already treats `None` as the sentinel, so only the annotation is wrong — no behaviour change.

`ruff` has a rule for exactly this and it fires on current `main`:

```
$ ruff check --isolated --select RUF013 sentence_transformers/
sentence_transformers/cross_encoder/fit_mixin.py:197:19: RUF013 PEP 484 prohibits implicit `Optional`
sentence_transformers/cross_encoder/fit_mixin.py:424:19: RUF013 PEP 484 prohibits implicit `Optional`
sentence_transformers/sentence_transformer/datasets/parallel_sentences.py:75:64: RUF013 ...
sentence_transformers/sentence_transformer/datasets/parallel_sentences.py:120:24: RUF013 ...
sentence_transformers/sentence_transformer/fit_mixin.py:176:19: RUF013 ...
sentence_transformers/sentence_transformer/fit_mixin.py:472:19: RUF013 ...
sentence_transformers/sentence_transformer/modules/cnn.py:28:23: RUF013 ...
Found 7 errors.
```

It isn't caught in CI because `[tool.ruff.lint] select` is `["E", "F", "W", "I", "UP"]` and `RUF` isn't included. After this change the same command reports `All checks passed!`.

**`fit` callbacks** (`SentenceTransformer.fit` and `CrossEncoder.fit`, two overloads each) — `callback: Callable[[float, int, int], None] = None`. The `None` in that annotation is the callable's *return* type, not an accepted value for the parameter. Each call site guards it:

```python
if callback is not None:
    callbacks.append(OriginalCallback(callback, evaluator))
```

**`ParallelSentencesDataset.load_data` / `.add_dataset`** — `max_sentences: int = None`, used as `if max_sentences is not None and max_sentences > 0`. `load_data`'s own docstring already documents it as `max_sentences (int, optional): ... Defaults to None`, so the docstring and the annotation currently disagree.

**`CNN.__init__`** — `stride_sizes: list[int] = None`, followed immediately by `if stride_sizes is None: stride_sizes = [1] * len(kernel_sizes)`.

I used `| None` rather than `Optional[...]` to match the codebase (roughly 1500 uses of `| None` against ~110 of `Optional[`), and all four files already have `from __future__ import annotations` with `requires-python = ">=3.10"`, so it's safe at runtime either way.

Since these are the only seven occurrences, adding `RUF013` to `select` would be a zero-noise way to keep them from coming back — happy to include that here or leave it out, whichever you prefer.

Checked with the repo's own config: `ruff check sentence_transformers/` → `All checks passed!`, `ruff format --check` → `216 files already formatted`.


tomaarsen (approved):
Looks fine to me! Thanks.

Issue #3450, opened 2025-07-18
Drop in Dev NDCG@10 After Training on CodeSearchNet-CCR (Python only): Possible Issue with In-Batch Negatives


gnatesan:
I’m running into a weird issue when training on the CodeSearchNet-CCR dataset. I’m focusing on just one language (Python), so every batch only has Python examples. The model I’m using is granite-embedding-125m-english, and I’m training with MultipleNegativesRankingLoss. There’s no language mixing happening within batches.

Here’s what’s happening: even though my training and validation loss both go down smoothly, the dev set NDCG@10 gets worse after training compared to the baseline (pretrained) model. It’s a pretty big drop. Sometimes training for just one step can boost the score, but once I go past a few steps, the dev score drops off and keeps going down with more training, even though the loss keeps improving.

Some extra info:

My batches are all Python, so the "negatives" in the batch are just other Python function pairs.

The dataset seems to be structured as one query and one positive document per pair, basically the two halves of a Python function. There aren’t really any hard negatives.

No languages are mixed together within a batch, just Python.

My guess right now is that since there is only a one-to-one mapping and no hard negatives, the in-batch negatives are just too easy. The model might be learning to distinguish unrelated code snippets without actually learning much about semantic matching, so dev set performance gets worse even though the loss looks good.

Has anyone else run into this, or found a good way to improve training on this kind of setup? I’m open to advice on better loss functions or negative sampling. Would love to hear any best practices for CodeSearchNet-CCR.

tomaarsen:
Hello!

Hmm. If the training and validation loss are dropping, but the NDCG@10 is getting worse, then it's possible that there's a discrepancy between the training/validation data and the dev set. Are you training *and* evaluating on the CodeSearchNet-CCR dataset? (If yes, then the above isn't the reasoning)

Another possibility is that the training is e.g. improving recall or precision at the cost of NDCG. This is technically possible, but pretty niche as they tend to grow together. 

It's good to have only Python, that allows for harder random negatives, but I agree that random negatives aren't as powerful as hard negatives and might be too easy. It's possible that it is improving on distinguishing very easy options, while decreasing performance on the very difficult ones. 

[...]

Issue #2814, opened 2024-07-07
InformationRetrievalEvaluator finding the absolute rank of first relevant_docs found or calculating mrr per group with evaluator


maayansharon10:
Hello,

I'm using all-mpnet-base-v2 model with InformationRetrievalEvaluator .
My queries and relevant docs are assosiated with groups, my objective is calcualting mrr@k and recall@k per group.

The results yields the metrics recall@k and mrr@k with cosine or dot product, but I can't find a way to get the actual rank of the sentence that this recall of mrr is calculated with, aka the rank of the best scored document in relevant_docs out of the corpus in regard to a query, per k.

The only way I found is to run the inference step again on the data and calculate it myself, but that seems like a computational waste because it is already being calculated in the process. 

Is there a way to do this more elegantly? 
If so, could you please direct me to an example?
Many thanks

tomaarsen:
Hello!

I think your best bet might be to create a subclass of the `InformationRetrievalEvaluator` which also outputs the rank of the best scored "true relevant" document. I think you'll get far by overriding the [`compute_metrics` method](https://github.com/UKPLab/sentence-transformers/blob/c263b88f25523f24cc803cb4a77cb2857af9a819/sentence_transformers/evaluation/InformationRetrievalEvaluator.py#L366-L458):
```python
 
from datasets import load_dataset, Dataset
import numpy as np
from sentence_transformers.evaluation import InformationRetrievalEvaluator
from sentence_transformers import SentenceTransformer
from typing import List
import logging

logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(message)s")
logger = logging.getLogger(__name__)

class CustomInformationRetrievalEvaluator(InformationRetrievalEvaluator):
    def compute_metrics(self, queries_result_list: List[object]):
        # Init score computation values
        num_hits_at_k = {k: 0 for k in self.accuracy_at_k}
        precisions_at_k = {k: [] for k in self.precision_recall_at_k}
        recall_at_k = {k: [] for k in self.precision_recall_at_k}
        min_rank_at_k = {k: [] for k in self.precision_recall_at_k}
        MRR = {k: 0 for k in self.mrr_at_k}
        ndcg = {k: [] for k in self.ndcg_at_k}
        AveP_at_k = {k: [] for k in self.map_at_k}

        # Compute scores on results
        for query_itr in range(len(queries_result_list)):
            query_id = self.queries_ids[query_itr]

            # Sort scores
[...]

PR #4011, opened 2026-09-13
[feat] Add per-query metrics, bootstrap confidence intervals and query groups to InformationRetrievalEvaluator

RubenHaisma:
## Motivation

NanoBEIR datasets have about 50 queries each. During training we compare checkpoints on 0.5-1 nDCG@10-point differences on these sets, with no estimate of how much of that is noise. `InformationRetrievalEvaluator.compute_metrics` already computes per-query precision, recall, nDCG and MAP, but averages them away immediately (accuracy and MRR are accumulated as scalars), so there is no way to get at the per-query values without reimplementing the evaluator.

This PR keeps the per-query values around and builds three opt-in things on top of them. Nothing changes unless you ask for it.

## What's added

- **Per-query metrics.** After a call, `evaluator.per_query_metrics[score_fn][metric]` (e.g. `["cosine"]["ndcg@10"]`) is a `np.ndarray` aligned with `evaluator.queries_ids`. `compute_metrics` is now `aggregate_per_query_metrics(compute_per_query_metrics(...))` and remains the override point for subclasses.
- **Bootstrap confidence intervals** (`bootstrap_resamples`, `bootstrap_confidence_level`, `bootstrap_seed`). Percentile bootstrap over queries; one set of resample indices is shared across all metrics and score functions so the intervals are comparable. `NanoBEIREvaluator` reports an aggregate CI via a stratified bootstrap over the per-dataset query sets (per-dataset CI bounds are excluded from the plain mean, since averaging bounds is not a CI).
- **Query groups** (`query_groups={qid: label}`). Every metric is additionally reported per group, with CIs when bootstrapping is on. This is the per-slice reporting asked for in #2814.
- **`paired_bootstrap_test(values_a, values_b)`** in `sentence_transformers.util`, together with `bootstrap_confidence_interval` and `bootstrap_indices`. It compares two models on the same queries and returns the mean difference, its CI and an approximate two-sided p-value as a `PairedBootstrapTestResult`.

`SparseInformationRetrievalEvaluator` and `MultiVectorInformationRetrievalEvaluator` accept the same four arguments; `SparseNanoBEIREvaluator` and `MultiVectorNanoBEIREvaluator` accept the three bootstrap arguments.

## Key naming

| What | Key (after the `name` prefix) |
|---|---|
| main metric (unchanged) | `test_cosine_ndcg@10` |
| confidence interval | `test_cosine_ndcg@10_ci_low`, `test_cosine_ndcg@10_ci_high` |
| group | `test_cosine_ndcg@10_<group>` (+ `_ci_low` / `_ci_high` when bootstrapping) |
| NanoBEIR aggregate CI | `NanoBEIR_mean_cosine_ndcg@10_ci_low` / `_ci_high` |
| per-query attribute | `evaluator.per_query_metrics["cosine"]["ndcg@10"]` -> `np.ndarray` of shape `(n_queries,)`, aligned with `evaluator.queries_ids` |

Order in the returned dict: main metrics first, then CI keys, then group keys. The CSV and `primary_metric` only contain the main metrics, as before.

## Backward compatibility

With the new arguments left at their defaults, the returned keys, values, CSV rows, `primary_metric` and log output are bit-identical to before. The aggregation deliberately reproduces the old arithmetic (an integer count divided by `n` for accuracy, a sequential sum for MRR, `np.mean` for the rest) instead of switching everything to `np.mean`, because the summation order changes the last bits of the floats.

This is tested by:

- `test_aggregates_bit_identical_to_legacy`: the previous `compute_metrics` body is copied verbatim into the test and compared with exact equality (no `approx`) on randomised result lists for several k-lists.
- `test_results_unchanged_without_opt_ins`: same key set and values as before, no `_ci_` keys, and the means of `per_query_metrics` reproduce the aggregates.
- `test_subclass_override_of_compute_metrics_still_used`: a subclass overriding `compute_metrics` still drives `__call__`.
- `test_write_csv_unchanged_with_bootstrap`: CSV column counts are equal with and without bootstrapping.
- Bootstrap and group behaviour: bounds contain the point estimate and stay within `[0, 1]`, seed determinism, group values on the mock fixture, NanoBEIR aggregate CIs are not means of the per-dataset bounds, plus unit tests for `sentence_transformers/util/statistics.py` including a hand-computed percentile case.

## Example

`examples/sentence_transformer/evaluation/evaluation_nanobeir_confidence_intervals.py` runs `all-MiniLM-L6-v2` and `static-retrieval-mrl-en-v1` on NanoMSMARCO and NanoNQ with `bootstrap_resamples=1000`, then runs the paired test on the per-query nDCG@10. With 50 queries the 95% intervals are roughly ±0.10 nDCG wide, so a 0.5-1 point difference between two checkpoints is well inside the noise, while the paired test still resolves a real gap:

```
sentence-transformers/all-MiniLM-L6-v2
  NanoBEIR_mean  nDCG@10: 0.5722 (95% CI: 0.4909 to 0.6494)
  NanoMSMARCO    nDCG@10: 0.5540 (95% CI: 0.4417 to 0.6563)
  NanoNQ         nDCG@10: 0.5904 (95% CI: 0.4820 to 0.7059)

sentence-transformers/static-retrieval-mrl-en-v1
  NanoBEIR_mean  nDCG@10: 0.4287 (95% CI: 0.3543 to 0.5011)
  NanoMSMARCO    nDCG@10: 0.4041 (95% CI: 0.3092 to 0.5011)
  NanoNQ         nDCG@10: 0.4534 (95% CI: 0.3626 to 0.5644)

Paired bootstrap test, sentence-transformers/static-retrieval-mrl-en-v1 minus sentence-transformers/all-MiniLM-L6-v2:
  NanoMSMARCO    nDCG@10 difference: -0.1499 (95% CI: -0.2565 to -0.0419), p = 0.004, significant over 50 queries
  NanoNQ         nDCG@10 difference: -0.1370 (95% CI: -0.2405 to -0.0446), p = 0.004, significant over 50 queries
```

## Related issues

- #2814: per-group metrics for the IR evaluator, covered by `query_groups`.
- #3450: nDCG@10 discrepancy between MTEB and Sentence Transformers. Per-query values make this kind of discrepancy diagnosable by diffing individual queries instead of one aggregate.

## Docs

- `docs/package_reference/util/statistics.rst` (new) and a toctree entry.
- Two sentences in the sentence-transformer, sparse-encoder and multi-vector training overviews.
- The training skill's evaluator and troubleshooting references.


tomaarsen:
Hello!

Thanks for putting this together. I understand the motivation, especially wanting to know whether a small improvement on NanoBEIR is meaningful. However, I'm not convinced the expected usage justifies the extra code and maintenance burden.

I'd like to keep our built-in evaluators focused on straightforward evaluation and tracking training progress. Confidence intervals, significance testing, and reporting per group feel better suited to a dedicated evaluation framework. Even exposing per-query metrics separately would introduce another public API to maintain, and I'm not sufficiently convinced of the demand for that at this point.

For more involved evaluation, I'd suggest looking at [MTEB](https://github.com/embeddings-benchmark/mteb). It supports Sentence Transformer models directly and [custom tasks built from local samples](https://docs.mteb.org/contributing/adding_a_dataset/), so users can evaluate their own data. MTEB doesn't currently provide these retrieval confidence intervals out of the box either, but they do have other valuable metrics that might give a solid comparison point.

I'm going to close this as out of scope for now. Thanks again for the contribution!

- Tom Aarsen

PR #4012, opened 2026-09-13
Stabilize ListMLE loss for large finite scores

vineethsaivs:
ListMLE and PListMLE overflow for scores near +1000 and return incorrect losses near -1000.

**Cause:** The Plackett-Luce normalizer exponentiates scores directly and adds epsilon after underflow.
**Fix:** Compute suffix normalizers with logcumsumexp. Use a finite padding sentinel to keep padded suffix gradients finite.
**Test:** Eight regression cases fail before; all 12 pass afterward against independent float64 cross-entropies, including gradients, padding and both ordering modes. The complete loss test file has 22 passes and 6 CUDA-only skips on CPU. Changed-file pre-commit checks pass.


tomaarsen:
Hello!

Thanks for opening this. I think this is a nice addition. Scores don't normally get too high to really be problematic, but now we don't have to worry about it anymore, plus I always appreciate the log-sum-exp. I simplified a tad to keep things a bit more concise, but otherwise this should be good. I'll review with Copilot and await tests.

- Tom Aarsen

PR #4013, opened 2026-09-13
Keep label-grouped batches valid across shuffled rounds

vineethsaivs:
GroupByLabelBatchSampler can emit a single-class batch despite its multi-class guarantee. With three labels, four samples per label, batch_size=4 and seed=0, the second batch is [1, 1, 1, 1].

**Cause:** A leftover pair can meet the same label at the next shuffled round, leaving triplet mining without negatives.
**Fix:** Swap the first two labels when that boundary repeats a carried label.
**Test:** Six regression cases fail before. Afterward, the sampler suite has 70 passes and 3 skips. Coverage includes ten epochs, imbalanced labels, partial batches, unique sample IDs and length agreement. Pre-commit passes.


tomaarsen:
Hello!

This is indeed a bit of an oversight and a slight bug. It's uncommon, but it could happen. It's also a pretty clean fix, although I've shortened the comment a bit. I also simplified the test a bit to keep it a bit shorter/quicker while still capturing the issue. I'll have Copilot have a look and run the tests.

- Tom Aarsen

PR #4014, opened 2026-09-14
[fix] Restore umask-respecting permissions on saved safetensors files

### LINKED ISSUE #4010: model.safetensors is missing read permissions after saving with SentenceTransformer.save()
When using `SentenceTransformer.save()`, `model.safetensors` has different read permissions from all other files in the model directory:
- other files have `-rw-rw-r--`, which is the standard value for most files
- `model.safetensors` has `-rw-------`

This is a small detail but can cause a bit of annoyance in particular use cases, for example:
- One wants to include a model in a Docker image
- He downloads and saves it when building the image. The saved files are by default owned by root.
- The folder in which the model was saved looks readable (it has  `-rw-rw-r--`)
- When one runs the container (which he is presumably not doing as root) and tries to load the model he gets a `FileNotFoundError: No such file or directory: model.safetensors`

Would it be possible to fix this?

### Environment:
- `sentence-transformers` 6.0.1
- `transformers` 4.57.6
- Python 3.12.14

AmirF194:
`model.safetensors` ends up at 0600 because the safetensors writer always creates files that way and ignores umask, while everything else `save()` writes goes through `open()`/`json.dump` and respects it. Confirmed it is not something sentence-transformers does: a plain transformers `BertModel.save_pretrained(safe_serialization=True)` and a bare `safetensors.torch.save_file()` call both produce 0600 under umask 022, with sentence-transformers out of the picture entirely.

Chmods each `*.safetensors` file right after `BaseModel.save()` writes it, to whatever mode umask would have given an ordinary file (`0o666 & ~umask`), so it ends up matching `config.json` and friends instead of a hardcoded value.

Added a test comparing the saved safetensors mode against a sibling json file's; skipped on Windows since permission bits like this do not really apply there. Checked in the same container both ways: 0600 vs 0644 without the fix, matching with it.

Fixes #4010


tomaarsen:
Hello!

Thanks for looking into this! Given that it also happens with a direct `safetensors.torch.save_file()` call, I think safetensors is the right place to fix it.

It looks like this started with the temporary-file change in https://github.com/safetensors/safetensors/pull/764 (e.g. see https://github.com/safetensors/safetensors/pull/764#discussion_r3363877239), and there's already a proposed fix in https://github.com/safetensors/safetensors/pull/789. Fixing it there would cover Transformers and direct module saves too.

In the meantime, setting the ownership or permissions during the Docker build should work around the original issue.

- Tom Aarsen

PR #4015, opened 2026-09-14
Fix compiled models with losses that require media counts in the trainer

### LINKED ISSUE #3582: Collator setting include_prompt_lengths prevents compilation
Hi, I am using sentence-transformers through setfit and want to compile the model with `torch.compile()`. The (simplified) code looks like this:

``` 
model = SetFitModel.from_pretrained(...)
model = torch.compile(model)
```

However, the following error occurs:

```
<...>  
  File "/path/to/my/workspace/.pyenv/versions/workspace-3.9/lib/python3.9/site-packages/setfit/trainer.py", line 341, in __init__
    self.st_trainer = BCSentenceTransformersTrainer(
  File "/path/to/my/workspace/.pyenv/versions/workspace-3.9/lib/python3.9/site-packages/setfit/trainer.py", line 50, in __init__
    super().__init__(
  File "/path/to/my/workspace/.pyenv/versions/workspace-3.9/lib/python3.9/site-packages/sentence_transformers/trainer.py", line 278, in __init__
    self.data_collator.include_prompt_lengths = self._include_prompt_length()
  File "/path/to/my/workspace/.pyenv/versions/workspace-3.9/lib/python3.9/site-packages/sentence_transformers/trainer.py", line 1028, in _include_prompt_length
    for module in self.model:
TypeError: 'OptimizedModule' object is not iterable
```

---

**Brief analysis:**

The problematic call is here:
https://github.com/huggingface/sentence-transformers/blob/6aaa53b3c77635b62013080937953ac7c601a894/sentence_transformers/trainer.py#L280-L281

I understand the problem, compiled modules cannot be iterated, but I don't know what the intention of this logic is.

To me it seems like we first check for the attribute on `data_collator`, only to ignore it and then try to get this information from them the modules again:
https://github.com/huggingface/sentence-transformers/blob/6aaa53b3c77635b62013080937953ac7c601a894/sentence_transformers/trainer.py#L1031-L1033

Also, `include_prompt_lengths` should likely be checked for True as well, which would prevent the error already (in my case at least).

RizgarOzan:
**What**
`SentenceTransformerTrainer` (and the other `BaseTrainer` subclasses) crash at construction when given a `torch.compile`'d model together with a loss that sets `requires_media_counts` (e.g. `CachedMultipleNegativesRankingLoss`, `CachedGISTEmbedLoss`):

```
base/trainer.py, in prepare_loss
    if isinstance(model[0], Router):
TypeError: 'OptimizedModule' object is not subscriptable
```

**Why**
Fixes #3582 (the remaining case). Since #3554 a compiled model trains fine with e.g. `MultipleNegativesRankingLoss`, but `prepare_loss` still does `model[0]` to turn on media counting, and `OptimizedModule` doesn't forward `__getitem__`. The rest of the training path already looks through the compile wrapper (`override_model_in_loss`, `AdaptiveLayerLoss`), so this was the last spot.

**How**
`prepare_loss` indexes the original model via `_orig_mod` when present (`getattr(model, "_orig_mod", model)`). Uncompiled models are unaffected; DDP wrapping happens later in `Trainer`, so the compile wrapper is the only one that can reach this point.

**How I tested**
- Added `test_trainer_compiled_model_with_cached_loss` to `tests/sentence_transformer/test_trainer.py`: fails on `main` with the `TypeError`, passes with the fix.
- A full `trainer.train()` with `torch.compile(model, backend="eager")` and `CachedMultipleNegativesRankingLoss` on CPU: fails before, trains after (`MultipleNegativesRankingLoss` works in both).
- `pytest tests/sentence_transformer/test_trainer.py tests/sentence_transformer/losses -n 4` → 253 passed, 3 skipped (torch 2.14.0+cpu, Python 3.14, Windows).
- `ruff check` / `ruff format --check` (0.15.6, as pinned in pre-commit) clean.


tomaarsen:
Hello!

Thanks for opening this. This is definitely worth fixing in my opinion, and the fix is clean enough. I did something similar in AdaptiveLayerLoss I believe, but also for DDP. Because a model here should never be wrapped in DDP yet, only the torch.compile fix should already be enough. Otherwise I would've created a utility to just unwrap any `BaseModel`, but right now it's overkill.

- Tom Aarsen

PR #4016, opened 2026-09-14
[`fix`] Prevent non-finite CSR reconstruction losses

eSVeeF:
### What happened

While using CSR training in a project, I encountered non-finite values in the auxiliary reconstruction loss after reducing the training batch size.

The normalization denominator in `normalized_mean_squared_error()` is the MSE between each input and the batch mean. That denominator becomes zero when:

- The batch contains a single example.
- All examples in the batch are identical.

This caused the loss to become `inf` or `nan`, even when the reconstruction itself was valid.

### What this changes

The normalization denominator is now clamped to the smallest representable value for its dtype before division. This keeps the loss finite while preserving the expected result for perfect reconstruction.

### Tests

Added regression coverage for:

- Singleton batches.
- Constant batches.
- Perfect reconstruction for both cases.

Validated with:

```text
python -m pytest tests/sparse_encoder/losses/test_csr.py -q
4 passed
```


tomaarsen:
Hello!

Thanks for opening. This is indeed an open edge case that we should address. I think your fix does work, but the `eps` fix can still cause some overflow that would be best to avoid. Instead, when variance normalization is undefined, I'm just using normal MSE. So I'm just setting the norm loss to 1.0 for those cases. I think that's nicest. I also updated the tests accordingly.

- Tom Aarsen

tomaarsen:
I think the current fix is sufficient, although perhaps there's some tiny remaining edge cases that overflow. I think it's primarily good now though, I don't want too much code to counteract niche edge cases.

- Tom Aarsen

PR #4017, opened 2026-09-14
[`fix`] Support disabling CSR auxiliary latents with k_aux=0

eSVeeF:
### What happened

While experimenting with smaller CSR configurations in one of my projects, I tried setting `k_aux=0` to disable the auxiliary latent path.

`SparseAutoEncoder.top_k()` already treats `k_aux=0` as disabling auxiliary latents, but the training path still tries to decode the resulting `None` value:

```python
recons_aux = self.decode(latents_auxk, info)
```

This raises an `IndexError` during training instead of allowing the auxiliary loss to be disabled.

### What this changes

- Skip auxiliary decoding when `k_aux=0`.
- Skip the auxiliary reconstruction loss when no auxiliary representation is available.
- Return a zero auxiliary loss tensor so the existing loss dictionary and training aggregation remain unchanged.
- Add regression coverage for training with `k_aux=0`.

The normal sparse and 4k reconstruction paths are unchanged.

### Tests

```text
python -m pytest tests/sparse_encoder/modules/test_csr.py -q -k without_auxiliary_latents
python -m pytest tests/sparse_encoder/modules/test_csr.py -q -k "not csr_outputs and not csr_encode"
```

Both pass locally.


tomaarsen:
Hello!

Fair enough, that's indeed a bit of a niche bug. I've simplified the fix though, just by setting `L_aux` to zeros. The rest looks good.

- Tom Aarsen

PR #4019, opened 2026-09-14
Avoid overflow in batch-hard soft-margin triplet loss

vineethsaivs:
Finite embeddings with large positive triplet margins produce an infinite loss and non-finite gradients.

**Cause:** The soft-margin formula explicitly computes exp(margin), which overflows before log1p can reduce it.

**Fix:** Use the numerically stable softplus operation for the same mathematical objective.

**Test:** `python -m pytest tests/sentence_transformer/losses/test_triplet.py -q`. The two large-distance float32/float64 cases fail before. All seven triplet tests pass after; loss and gradients match an independently enumerated double-precision reference. CPU validation with PyTorch 2.10.0; no model downloads or GPU run. Changed-file lint passes.


tomaarsen:
Hello!

This seems like a fine fix, although a bit unnecessary. I simplified the test a bit as nobody trains with fp64 anyways.

- Tom Aarsen

PR #4020, opened 2026-09-15
Keep single hard pairs in OnlineContrastiveLoss

vineethsaivs:
A batch with one positive and one negative produces zero online contrastive loss even when both pairs are hard.

**Cause:** The mining thresholds ignore a label group when it contains exactly one pair, comparing the other group against its own mean instead.

**Fix:** Use the other label group whenever it is nonempty, retaining its single hard pair.

**Test:** `python -m pytest tests/sentence_transformer/losses/test_contrastive.py -q`. Three loss-and-gradient regressions fail before. All eight contrastive tests pass after. Changed-file pre-commit passes. CPU tests use the actual loss module and explicit distance/gradient references; no model downloads or GPU run.


tomaarsen:
Hello!

This fix seems fine to me. I simplified `len(...) > 0` to just `len(...)` though, as it's equivalent except truthy. I also updated the test a bit.

- Tom Aarsen

PR #4021, opened 2026-09-15
Match contrastive-tension loader length to its batches

vineethsaivs:
The contrastive-tension loader underreports its batch count and drops a final complete positive-only batch.

**Cause:** Length assumes every pair consumes two sentences, although a positive pair consumes one. Iteration always requires two remaining sentences.

**Fix:** Count the sentences actually consumed per batch and require a second sentence only for a negative pair.

**Test:** `python -m pytest tests/sentence_transformer/losses/test_contrastive_tension.py -q`. Twenty-six cases fail before. All 46 cases pass after, covering reported versus yielded lengths, positive/negative ratios and the final positive pair. Changed-file pre-commit passes. CPU data-loader tests exercise the real iterator; no full embedding training run.


tomaarsen:
Hello!

This seems like a fine fix, although this loader is only used for legacy training now. I've simplified the iterator a bit to use the corrected `__len__` directly, so we no longer need the manual indexing and exhaustion checks. I also simplified the tests to check explicit expected batch counts.

- Tom Aarsen

PR #4022, opened 2026-09-15
[fix] Use inner product for approximate FAISS search

xblwh:
## Problem

`semantic_search_faiss(..., exact=False)` builds `IndexHNSWFlat` without a metric argument. FAISS defaults that constructor to squared L2 distance, while the exact path uses `IndexFlatIP`. Switching to approximate search therefore changes both the ranking objective and the meaning of the returned scores for `float32` and `uint8` corpora.

For query `[1, 0]` and corpus `[[1, 0], [2, 0], [0, 3]]`, with `top_k=2` and `rescore=False`:

| Path | Corpus IDs | Scores |
| --- | --- | --- |
| Exact, before and after | `[1, 0]` | `[2.0, 1.0]` |
| Approximate, before | `[0, 1]` | `[0.0, 1.0]` |
| Approximate, after | `[1, 0]` | `[2.0, 1.0]` |

## Change

Pass `faiss.METRIC_INNER_PRODUCT` when constructing the non-binary HNSW index. Add real-FAISS regression coverage for both supported scalar precisions and both values of `exact`, checking ranks and scores for two queries. The small corpus keeps the regression focused on metric selection; approximate retrieval can still differ from exact retrieval on larger datasets.

## Validation

- New regression cases before the fix: **2 failed, 2 passed** (both approximate cases failed).
- `python -m pytest tests/util/test_quantization.py -q`: **69 passed**, with FAISS and USearch installed.
- `python -m pytest tests/util/test_similarity.py tests/util/test_tensor.py tests/util/test_retrieval.py -q`: **126 passed, 6 skipped, 1 deselected**.
- `make check`: all repository pre-commit hooks passed.
- `git diff --check`: passed.

Local environment: macOS, Python 3.12.14, NumPy 2.5.3, PyTorch 2.14.0, FAISS 1.15.0, USearch 2.26.2. Test commands used `OMP_NUM_THREADS=1`: the initial unfixed regression run hit a native FAISS crash with the local default thread settings. Both the failing baseline and passing fix were then tested with the same single-thread setting; no project thread configuration was changed.

The adjacent suite initially lacked one model fixture in offline mode. After downloading the public tiny SPLADE fixture and running that test successfully, the full adjacent command passed in offline mode. The complete model suite and Linux/Windows matrix were not run locally.

Implementation and regression tests were generated with OpenAI Codex; the patch and validation evidence were provided for human review before submission.


tomaarsen:
Hello!

Yes, this is indeed a big oversight. It'd be great to fix this. I also think your fix + test are clean and focused, thanks for that.

- Tom Aarsen

Issue #2282, opened 2023-08-23
Finetuning Cross-Encoder model results in constant prediction but when validated on test set it shows 100 F1 and accuracy by evaluator function


ratikapoor:
While using cross-encoder (CrossEncoder('distilroberta-base',num_labels=1)), I'm getting F1 score of 1 on the validation set with labels 0 and 1 after fine tuning. Although fine tuning is not happening properly. While inferencing it gives a constant score of 0.4118. Then also in logs, it shows 100 accuracy and F1 score.
CEBinaryClassificationEvaluator: Evaluating the model on fk-dev dataset after epoch 1:
Accuracy: 100.00 (Threshold: 0.4118)
F1: 100.00 (Threshold: 0.4118)
Precision: 100.00
Recall: 100.00
Average Precision: 30.00

I'm not able to understand why it is not learning anything where I have 11k rows in the data with the target label equally distributed.

progsi:
I am having the same issue. Any news on this? I am using the model like a normal torch model with switching between train and test modes.

FDSRashid:
Me too :(

tomaarsen:
Hello!

I'm trying to reproduce this by running this training script: [training_quora_duplicate_questions.py](https://github.com/UKPLab/sentence-transformers/blob/master/examples/training/cross-encoder/training_quora_duplicate_questions.py), which also uses `CEBinaryClassificationEvaluator` and `distilroberta-base`. (I also limited the training samples to 10000 & increased the batch size to keep the training fast)

And in my case, the accuracies and thresholds do update every epoch:
```
2023-12-21 10:22:05 - Read train dataset
2023-12-21 10:22:05 - Read dev dataset
Some weights of RobertaForSequenceClassification were not initialized from the model checkpoint at distilroberta-base and are newly initialized: ['classifier.out_proj.bias', 'classifier.out_proj.weight', 'classifier.dense.weight', 'classifier.dense.bias']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
2023-12-21 10:22:07 - Use pytorch device: cuda
2023-12-21 10:22:07 - Warmup-steps: 32
Iteration: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████| 79/79 [00:34<00:00,  2.26it/s]
2023-12-21 10:22:42 - CEBinaryClassificationEvaluator: Evaluating the model on Quora-dev dataset after epoch 0:                                                
[...]

PR #4023, opened 2026-09-15
Fix attainable binary classification thresholds for tied scores

HaokaiDing:
Refs #2282.

The binary threshold helpers evaluate splits inside tied score groups. With scores `[0.5, 0.5]` and labels `[1, 0]`, they report accuracy and F1 of 1.0 at threshold 0.5, although that threshold predicts both samples positive.

Evaluate thresholds only after complete score groups, and include all-negative/all-positive predictions. Keep ordinary midpoint thresholds, while preventing a rounded midpoint from including the next score group. Explicit ufunc dtypes preserve float32 threshold behavior across supported NumPy scalar-promotion rules.

The minimal example now returns best accuracy 0.5 and best F1 2/3. This fixes the evaluator metrics; it does not diagnose the constant predictions from the training runs discussed in #2282.

Validation:
- Before the fix: 18 failed and 2 passed in the new 20-case parameterized regression.
- After the fix: `HF_HUB_OFFLINE=1 OMP_NUM_THREADS=1 .venv/bin/python -m pytest tests/sentence_transformer/evaluation/test_binary_classification_evaluator.py -k 'not multiple_similarity_fn_names and not csv_columns_are_aligned' -q --tb=short`: **29 passed, 5 deselected**.
- Tests check returned metrics against actual threshold predictions and the best attainable sklearn score, for ties, both score directions, endpoints, single samples, adjacent float32 scores and integer inputs.
- Existing distance midpoint threshold 6.0 remains covered and passing.
- Changed-file pre-commit checks (Ruff lint, Ruff format, typos) and `git diff --check` pass.
- The five tests requiring a downloaded model and the full suite were not run. An older NumPy environment was not run separately.

Prepared with OpenAI Codex; validation ran locally, with an independent review by another Codex agent.


tomaarsen:
Hello!

Thanks for investigating this! Splitting tied score groups can indeed produce accuracy and F1 scores that no threshold can achieve, so this is a useful fix.

I kept ordinary `/ 2` midpoint arithmetic and removed the unnecessary F1 threshold initialization. I've also simplified the tests to check explicit expected metrics alongside the predictions from the returned thresholds.

Looks like Copilot is already through its quota for the month just due to how many PRs I've gotten :S 

- Tom Aarsen

PR #4025, opened 2026-09-15
Reuse calibration ranges across multi-vector embeddings

ayoubchikri:
## Problem

Quantizing a list of multi-vector document embeddings to int8/uint8 scans the same calibration corpus once per nonempty document. With N documents and T calibration tokens, calibration alone does O(N × T × dimension) work, even though every document uses the same ranges.

Compute the shared ranges once before quantizing the individual matrices. Explicit ranges still take precedence, and empty documents and all-empty corpora retain their existing output shapes.

## Evidence

CPU microbenchmark against `94d68cf`, NumPy 1.26.4, ragged float32 documents with 16–48 tokens and 64 dimensions; median of three runs:

| Documents | Precision | Before | After |
| --- | --- | ---: | ---: |
| 256 | int8 | 185 ms | 11.8 ms |
| 1,024 | int8 | 3,688 ms | 50.1 ms |
| 1,024 | uint8 | 4,185 ms | 45.0 ms |

Quantized arrays are bitwise identical on all six measured configurations. These timings cover quantization only, excluding model encoding and retrieval.

## Validation

- 12 new parameterized cases cover NumPy/Torch inputs, int8/uint8, inferred/provided calibration, explicit ranges, constant dimensions, clipping, and empty documents. They compare against dense quantization with the same ranges and check that calibration reductions are performed once.
- Against the unmodified upstream function, the eight cases that require calibration fail because each reduction runs three times. Explicit-range cases already pass.
- **Linux: 174 passed, 5 skipped**, including real FAISS/USearch indexes, similarity and tensor utilities, on Python 3.12 / PyTorch 2.8.0+cpu / Transformers 5.0.0. [Run on the exact PR commit](https://github.com/ayoubchikri/sentence-transformers/actions/runs/35001608926).
- Official pre-commit hooks (Ruff check/format and typos) pass.

Relevant suite: `python -m pytest tests/util/test_quantization.py tests/util/test_similarity.py tests/util/test_tensor.py -q`. The full project suite and GPU benchmarks were not run.


tomaarsen:
Hello!

Thanks for this! I think computing the shared ranges once is a useful improvement for callers quantizing multi-vector embeddings directly. I verified the speedup locally, with identical outputs, and all 77 quantization tests passed.

I've shortened the comments a bit and fixed a docstring typo, but otherwise this looks good to me.

- Tom Aarsen

PR #4026, opened 2026-09-15
Use quadratic-memory semi-hard triplet mining

ayoubchikri:
## Problem

`BatchSemiHardTripletLoss` repeats its pairwise distance matrix into a `(batch_size**2, batch_size)` tile and builds an equally large candidate mask. This makes mining use cubic intermediate memory, limiting batch sizes during embedding training.

Sort each anchor's negative distances once, then use `searchsorted(..., right=True)` to find the nearest negative strictly farther away than each positive. If none exists, keep the existing farthest-negative fallback. Stable sorting preserves the first-index tie break. Mining uses quadratic intermediate memory and O(B² log B) work; selected distances remain differentiable through gather.

## Evidence

CPU PyTorch 2.2.2, one thread, 32-dimensional embeddings, median of five forward/backward runs:

| Batch size | Before | After | Additional unique storage saved for backward, before → after |
| --- | ---: | ---: | ---: |
| 128 | 58.4 ms | 3.39 ms | 2,672,648 → 722,952 bytes |
| 256 | 501.8 ms | 12.8 ms | 19,075,080 → 2,887,688 bytes |

Saved storage is measured with `saved_tensors_hooks`, excluding original embeddings and labels. It is not total/peak process or GPU memory. Losses match within tolerance and embedding gradients are bitwise identical for benchmark batches 32, 64, 128 and 256. Timings cover the loss only, excluding the encoder.

## Validation

- 16 comparisons against a direct per-triplet reference cover Euclidean/cosine distances, float32/float64, tied distances, and batches without positives or negatives, including gradients.
- A storage regression fails on upstream: the largest saved tensor has 32³ elements instead of at most 32². It passes after this change.
- Three SGD steps on a small linear encoder produce the same parameters as the direct reference.
- All 21 tests in `tests/sentence_transformer/losses/test_triplet.py` pass locally.
- **All 21 tests also pass on Linux**, Python 3.12 / PyTorch 2.8.0+cpu / Transformers 5.0.0. [Run on the exact PR commit](https://github.com/ayoubchikri/sentence-transformers/actions/runs/35001608926).
- Official pre-commit hooks (Ruff check/format and typos) pass.

The full project suite and GPU/distributed training were not run.


tomaarsen:
Hello!

Thanks for this! Avoiding the cubic distance tile is a useful improvement. I verified the speedup locally on both CPU and CUDA. With a batch size of 256 on an RTX 3090, the loss used about 4.5 MiB of additional peak tensor memory instead of 210 MiB, and forward/backward took roughly half the time.

I've shortened the comment, removed the unused `_masked_minimum` helper, and trimmed the tests a bit. The gradient comparisons already cover most of what the extra SGD test was checking.

I also found that this corrects an existing gradient issue when a selected negative ties with the largest distance in its row. The old implementation can send that gradient through a different sample, whereas this version matches the direct reference. I've added a regression test for that case, so we explicitly cover the behavior change too.

Overall, this looks good to me.

- Tom Aarsen

PR #4028, opened 2026-09-15
Group tied scores when evaluating paraphrase mining

vineethsaivs:
Identical paraphrase scores can report perfect average precision and F1 merely because the positive pair sorts first.

**Cause:** The evaluator updates metrics after each pair, including intermediate points inside a group of equal scores that no threshold can separate.

**Fix:** Update precision, recall and F1 only after each complete score group, and weight its average-precision contribution by all positives in that group.

**Test:** `PYTHONPATH=. python -m pytest tests/sentence_transformer/evaluation/test_paraphrase_mining_evaluator.py -k 'not test_ParaphraseMiningEvaluator' -q`. Two tied-order regressions fail before. Seven focused tests pass after, using a real BoW encoder and independent sklearn average-precision and F1 references. Changed-file pre-commit passes. Native CPU evaluation, no downloaded transformer model or GPU run.


vineethsaivs:
CI is waiting for approval. Could a maintainer start it? Seven focused CPU checks and pre-commit pass locally.

tomaarsen:
Hello!

Thanks for this! We just merged a similar fix for `BinaryClassificationEvaluator` in #4023. This looks like a useful fix here too, and the results stay the same when there are no tied scores as far as I can tell.

I've adjusted the mixed-score test to include two positive pairs in the same tied group. Previously, each group only had one positive, so the test still passed if the new average-precision weighting was removed. 

- Tom Aarsen

PR #4032, opened 2026-09-16
Return zero ranking loss when no label pairs are ordered

vineethsaivs:
LambdaLoss and RankNetLoss return NaN when every relevance label is tied or each query has one document.

The final reduction takes the mean of an empty set of ordered label pairs. Return a graph-connected zero when there are no pairs, preserving the existing mean for nonempty pairs.

Tests: `PYTHONPATH=. python -m pytest tests/cross_encoder/losses/test_misc.py -k ranking_loss -q`. Six regressions fail before. Eight CPU tests pass after, covering zero and positive tied labels, single-document queries, zero gradients and mixed batches. Changed-file pre-commit passes. Native PyTorch 2.10 CPU tests; no model download or GPU run.


vineethsaivs:
Could you approve the CI run? The eight focused CPU tests and changed-file pre-commit checks pass.

PR #4033, opened 2026-09-16
Keep Euclidean similarity gradients finite at zero distance

vineethsaivs:
Identical dense embeddings have zero Euclidean distance but produce NaN gradients.

Differentiating sqrt(sum(square(difference))) at zero introduces an infinite square-root derivative. Use the vector-norm primitive for dense tensors, which supplies a finite zero subgradient at coincidence.

Tests: `PYTHONPATH=. python -m pytest tests/util/test_similarity.py -k euclidean -q`. The coincident-embedding gradient regression fails before. Seven Euclidean tests pass after, including existing sparse behavior and analytical 3-4-5 distance gradients. Changed-file pre-commit passes. Native PyTorch 2.10 CPU tests; sparse backward support is outside this change.


vineethsaivs:
Could you approve the CI run? The seven focused CPU tests and changed-file pre-commit checks pass.

tomaarsen:
Hello!

Thanks for this! This looks like a useful fix. I verified that identical anchor and positive embeddings can produce NaN gradients in `TripletLoss`, even when the loss is zero. The new calculation fixes that, and the values and gradients otherwise match within floating-point tolerance.

I've kept the old calculation for integer inputs, since `vector_norm` rejects those, and added a small regression test for integer lists, NumPy arrays, and tensors. I also simplified the calculation a bit and corrected the docstring.

- Tom Aarsen

tomaarsen (approved):

vineethsaivs:
Thanks for taking it the rest of the way. Keeping the old path for integer inputs is the right call, I had not considered that `vector_norm` rejects them, and the integer regression test covers the case my version would have broken.

PR #4035, opened 2026-09-16
Reduce repeated inference placement overhead

tomaarsen:
Hello!

Heads up, this PR was AI-generated and human-reviewed.

## Pull Request overview

* Cache Accelerate placement hook discovery
* Prepare delegated module placement once on first inference
* Reuse the resolved inference device for the HPU check

## Details

Following #3986, I've reduced the placement work performed on each inference call. Previously, we searched the module hierarchy for Accelerate hooks repeatedly and realigned auxiliary modules on every call when a device map or Accelerate hooks controlled placement.

I've cached `_accelerate_placement_hook`, including when no hook is present. Delegated models still align auxiliary modules during construction, then once more on first inference to include modules added after initialization. For example, appending a `Dense` module before the first `encode()` call still places it on the correct device and matches the model's dtype. Subsequent calls skip this alignment.

For changes after that first inference, `clear_placement_cache()` explicitly refreshes hook discovery and prepares auxiliary modules again. Ordinary models retain the parameter and buffer device check on each call, so explicit device requests and newly added modules still work without a placement cache reset.

I benchmarked the original #3986 placement methods, hook caching alone, and hook caching with first-inference alignment on identical loaded weights. Median repeated placement setup times in microseconds were:

| Model and placement | Original #3986 | Hook cache only | Hook cache and first-inference alignment |
| --- | ---: | ---: | ---: |
| MiniLM-L6, CUDA | 471 | 235 | 237 |
| MPNet base, CUDA | 795 | 393 | 396 |
| MiniLM-L6, CUDA device map | 458 | 302 | 14 |
| MPNet base, CUDA device map | 762 | 499 | 14 |
| BERT tiny, CPU offload to CUDA | 93 | 84 | 3 |

These measurements use an RTX 3090 and i7-13700K on Windows, with FP32 weights and nine interleaved timing rounds. Setup includes the HPU device lookup and inference device resolution. The measurements precede the final cleanup removing the duplicate HPU device lookup. First-call setup for the device-map cases remains approximately 297 microseconds for MiniLM and 470 microseconds for MPNet, excluding loading and transfers for newly appended modules.

Full `encode()` latency dropped from 4.565 to 4.090 ms for one sentence with MiniLM's CUDA device map, and from 12.138 to 11.325 ms for 32 sentences with MPNet's CUDA device map. GPU timings varied across runs, so these are workload-specific results. Embeddings matched across the three implementations.

Validation: 80 targeted placement tests passed, including appended modules, FP16 alignment, offloading, cache refresh, and retry after failed placement. Ruff lint and formatting checks passed. Mocked HPU checks verified that device resolution precedes graph wrapping and wrapping happens once. HPU hardware was not available.

- Tom Aarsen


tomaarsen:
cc @NewUserHa I'm caching a bit more, and trying to stop moving/traversing as much on every single inference. I think this should be quite efficient now, especially when using `device_map`.

- Tom Aarsen

PR #4040, opened 2026-09-17
[docs] add the .md extension to the NLI/STS README links

simpleqt:
The matryoshka (5 links) and adaptive_layer (4 links) example READMEs link `../nli/README` / `../sts/README` without the `.md` extension — GitHub doesn't resolve that form, so all 9 links 404. Every other example README already uses the `.md` form; aligned these two.


tomaarsen:
Hello!

Yeah, these are mostly for building the documentation in www.sbert.net, which didn't require the .md. However, with the .md also updates correctly to the .html that it needs to be on the website, so I feel like it seems fine to merge this.

- Tom Aarsen

simpleqt:
Thank you! The one red matrix cell (3.11/ubuntu) looks like a flake given this PR only touches README links — I can't re-run workflow runs from the fork side (403), but happy to push a re-trigger commit or leave it to CI on merge, whichever you prefer.

PR #4057, opened 2026-09-20
Pin UTF-8 encoding when reading tokenizer/IDF JSON files

dltsum:
## What

Three JSON reads fall back to the locale encoding:

- `PhraseTokenizer.load` (sentence_transformer/modules/tokenizer/phrase.py)
- `WhitespaceTokenizer.load` (sentence_transformer/modules/tokenizer/whitespace.py)
- `SparseStaticEmbedding.from_json` (sparse_encoder/modules/sparse_static_embedding.py)

## Why

On Windows with a non-UTF-8 locale (e.g. cp936/GBK), reading a JSON file that contains literal UTF-8 non-ASCII content crashes with `UnicodeDecodeError`. The library's own `save()` escapes non-ASCII (json.dump default), but vocab/IDF JSON files are also user-authored or hand-edited — an IDF file with CJK tokens saved as UTF-8 is realistic.

## Reproduction (Windows, cp936 locale)

```python
from sentence_transformers.sparse_encoder.modules.sparse_static_embedding import SparseStaticEmbedding

open("idf.json", "w", encoding="utf-8").write('{"汉 token": 1.5}')
SparseStaticEmbedding.from_json("idf.json", tokenizer=...)
# UnicodeDecodeError: 'gbk' codec can't decode byte 0x89 in position 4: illegal multibyte sequence
```

Verified red for all three before the fix; after the fix the files load correctly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tomaarsen:
Hello!

Seems like a fine fix to me, thanks for this. I'll merge when the tests go green.

- Tom Aarsen

PR #4062, opened 2026-09-21
Stop discovering tiny models from trl-internal-testing

albertvillanova:
Following up on huggingface/trl#7202, where @tomaarsen offered this.

TRL is auditing `trl-internal-testing` and retiring the repos that have no generation script. Because `find_model_for_architecture` lists `trl-internal-testing` first among its authors, that org currently backs 16 entries in `transformers_tiny_models.json`, and a deletion on our side would 404 here.

### Changes

**Discovery**: drop `trl-internal-testing` from the author list. That also makes the `author == "trl-internal-testing"` half of the match condition dead, so the condition simplifies to requiring `tiny-random` in the model id, which is what the remaining authors all use.

**Mapping**: `update_tiny_model_mapping` skips architectures that already have a value, so dropping the author does not rewrite existing entries. Three are repointed by hand, the three TRL is deleting:

| Architecture | Was | Now |
|---|---|---|
| `deepseek_v4` | `trl-internal-testing/tiny-DeepseekV4ForCausalLM` | `tiny-random/deepseek-v4-bf16` |
| `phi3` | `trl-internal-testing/tiny-Phi3ForCausalLM` | `onnx-internal-testing/tiny-random-Phi3ForCausalLM` |
| `qwen2_audio` | `trl-internal-testing/tiny-Qwen2AudioForConditionalGeneration` | `null` |

Both replacements were resolved by re-running your own matching logic with the author removed, and both repos exist.

`qwen2_audio` has no candidate in any of the remaining authors. There is `yujiepan/qwen2-audio-tiny-random`, which is a genuine `Qwen2AudioForConditionalGeneration`, but it sits outside your curated orgs so I left the entry null rather than pin a personal account on your behalf. Happy to pin it if you would rather keep the coverage.

### What is deliberately not changed

The other 13 entries still point at `trl-internal-testing`. Those models are in active use by TRL's own test suite, so they are not going anywhere, and repointing them would cost you coverage for no benefit: 10 of the 16 have no equivalent under the remaining authors, and several of them (`internvl`, `qwen2_5_vl`, `idefics2`, `diffusion_gemma`) are specially handled in your conftest rather than skipped.

One heads-up rather than a change: `dbrx` is likely to move to `hf-internal-testing`. A Hub transfer leaves a redirect, so `trl-internal-testing/tiny-DbrxForCausalLM` will keep resolving and nothing breaks, but you may prefer to repoint it once it lands.


tomaarsen:
Hello!

I ran some tests locally for this, and noticed no issues. I'm happy to assist in the trl tiny model deprecations.

- Tom Aarsen

